PR #32, opened 2018-12-12
OPS-30457 pipelines

brtj:


gdubicki:
This is Egnyte-specific code, @bjarocki. Please change it to generic operations on GitLab API pipelines and move Egnyte-specific code to our internat repo.

brtj:
Deleted hardcoded things, 2 parameters added instead.

bjarocki:
@brtj Hi Bartek! Do you by any chance keep `bjarocki` in your `~/.gitconfig` file as the github.user? I got an email about the https://github.com/egnyte/gitlabform/pull/32#issuecomment-446557912 comment mention.

gitlabform/gitlab/pipelines.py
```diff
@@ -0,0 +1,14 @@
+from gitlabform.gitlab.core import GitLabCore
+
+
+class GitLabPipelines(GitLabCore):
+
+    def get_pipelines(self, project_and_group_name, branch):
+        pipelines = self._make_requests_to_api("projects/%s/pipelines?ref=%s", (project_and_group_name, branch))
```
gdubicki: Please check if this endpoint is not paginated. In that case you need to do pass appropriate parameter to `_make_requests_to_api` method, see https://github.com/egnyte/gitlabform/commit/ee00182f973ffbf32766c23a6000548c3e2b999f as an example.

gitlabform/gitlab/pipelines.py
```diff
@@ -0,0 +1,14 @@
...
+        return pipelines
+
+    def retry_pipeline(self, project_and_group_name, pipeline_id):
+        pipeline = self._make_requests_to_api("projects/%s/pipelines/%s/retry",
```
gdubicki: Please return `pipeline` from this method. This will allow you to check if this pipeline succeeded by only querying this single pipeline by id.

Of course to be able to do that you should also add `get_pipeline` (single one, by id) method to this class.

gitlabform/gitlab/pipelines.py
brtj: Ok will do. But is it necessary if we are taking last pipeline?

gitlabform/gitlab/pipelines.py
brtj: Added as requested.
BTW in our case we are taking last pipeline, so for get_pipelines default sorting is descending.

PR #34, opened 2018-12-30
setup.py: Fix naming of tests_require keyword

mrueg:
tests_require instead of tests_requires See https://setuptools.readthedocs.io/en/latest/setuptools.html


dom-nie:
Thanks for your contribution to gitlabform.

PR #38, opened 2019-04-08
Support project push rules

icy:
TODO: add some testing support.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -70,6 +70,17 @@ group_settings:
       only_allow_merge_if_pipeline_succeeds: true
       only_allow_merge_if_all_discussions_are_resolved: true
 
+    # See https://docs.gitlab.com/ee/push_rules/push_rules.html#enabling-push-rules
+    project_push_rules:
+      commit_message_regex: "Fixes \d +"
+      branch_name_regex: ""
+      deny_delete_tag: false
+      member_check: false
+      prevent_secrets: false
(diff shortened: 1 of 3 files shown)


gdubicki:
Tests would be generally nice, but I am not going to block this MR because of them being missing if this project doesn't have them for other code yet. LGTM, thanks for your contribution! :)

icy:
You're welcome @gdubicki . 

Though the PR is merged I still wonder if we should change config name from `project_push_rules` to `push_rules`? There are also `project_settings`.

gdubicki:
I was thinking about it too. But as they are called "project push rules" in the API docs (https://docs.gitlab.com/ee/api/projects.html#push-rules-starter) and you can also set them globally for all new projects (https://docs.gitlab.com/ee/push_rules/push_rules.html#enabling-push-rules) I think that it's not a bad thing to keep the current config name. Also it would be a bit problematic to change it now. ;)

icy:
@gdubicki  I agree. I think the problem comes from Gitlab per-se. project, repository, that's quite confusing. For example, they have "push rules" for projects (`Settings -> Projects -> push rules`), but at the same time they have branch protection at repository level (`Settings -> Repository -> Branch protection`), however the api seems to link to project: https://docs.gitlab.com/ee/user/project/protected_branches.html . I'm totally confused when working with Gitlab web UI =))

PR #39, opened 2019-04-14
Add protected tags functionality

therealgambo:


gitlabform/gitlabform/core.py
```diff
@@ -307,6 +308,28 @@ def process_branches(self, project_and_group, configuration):
                 if self.args.strict:
                     exit(3)
 
+    @if_in_config_and_not_skipped
+    def process_tags(self, project_and_group, configuration):
+        for tag in sorted(configuration['tags']):
+            try:
+                if configuration['tags'][tag]['protected']:
+                    create_access_level = configuration['tags'][tag]['create_access_level'] if \
+                        'create_access_level' in configuration['tags'][tag] else '30'
```
gdubicki: I would not pass any value to GitLab's API if user won't set the `create_access_level` in gitlabform's config. Then GitLab's API default (`40`, according to https://docs.gitlab.com/ee/api/protected_tags.html#protect-repository-tags), will be used, which I think is what the user expects.

gdubicki:
Thank you for your contributions, @therealgambo ! 🥇 

I did a review, please see my comments above and in https://github.com/egnyte/gitlabform/pull/40 .

therealgambo:
@gdubicki updated with changes to only send that parameter if it is defined in the configuration.

gitlabform/gitlab/tags.py
```diff
@@ -14,3 +14,16 @@ def create_tag(self, project_and_group_name, tag_name, ref, message=None):
         }
         return self._make_requests_to_api("projects/%s/repository/tags", project_and_group_name, method='POST', data=data,
                                           expected_codes=201)
+
+    def protect_tag(self, project_and_group_name, tag_name, create_access_level):
+        data = {
+            'name': tag_name
+        }
+        if create_access_level:
```
glensc, line 22: I think this is the reason for this bug:
- https://github.com/egnyte/gitlabform/issues/172

gitlabform/gitlabform/core.py
glensc: the "exists"-check is done incorrectly:
- https://github.com/egnyte/gitlabform/pull/39#discussion_r566725404

PR #41, opened 2019-04-26
Add option to ignore SSL errors

akaRem:
* Remove global requests' session
* Split __init__ method because it's overcomplicated

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -6,6 +6,8 @@ gitlab:
   token: "<private token of an admin user>"
   # GitLab API version. As of now only 4 is supported.
   api_version: 4
+  # SSL will be verified by default, you can turn checks off by passing `ssl_verify: false`
+  ssl_verify: true
 
 #
 # You can define settings on 3 levels here:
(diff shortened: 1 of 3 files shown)


gdubicki:
I am for adding "ignore SSL" option (disabled by default, of course), but I have doubts about the refactoring because it changes the API, which would require changes in a few internal projects that are using gitlabform as a library.. Perhaps you could change the code to make the public methods, including constructor unchanged, @akaRem ?

akaRem:
@gdubicki 

Sorry that it conflicts with your internal libraries — I didn't know about this.
Refactoring is not required for this change, but I still decided to put SSL stuff on top because I don't like to patch global variables which are not excluded from export. This may lead to unexpected behaviour.

In the same time, initialiser is very complicated and includes a lot of logic which normally should be placed either in factory (preferable) methods or private/protected instance methods..

SSL and refactoring doesn't block me for using this tool (and also not very important for now).
So I don't want to spent time on reworking this PR for now, may be later.

But you see the idea for SSL stuff and you can reuse it in improving this tool in way you prefer.

gdubicki:
I have reused your code for adding option ignore SSL errors that will be soon released in v. >= 1.7.0, @akaRem . Thanks! I am closing this PR for now then.

PR #43, opened 2019-04-26
Fix error on empty groups

akaRem:
When running with ALL target, application is trying to list all
projects in all groups. But it crushes when there are groups
without any project.

This change fixes such errors.


gdubicki:
Thanks a lot for your contributions, @akaRem ! I will try to review them and get them merged within the next few days.

This particular one looks like an obvious improvement so I will just merge it now.

PR #44, opened 2019-04-26
Add simple diffing

akaRem:
Currently noop mode just notifies that something will be changed and
gives no clue regarding actual changes which will be performed.

This change adds simple diffing implemented using currently supported
api calls and functionality.

### DIFF
--- a/gitlabform/gitlabform/core.py
+++ b/gitlabform/gitlabform/core.py
@@ -6,6 +6,9 @@ import sys
 from pathlib import Path
 import os
 from functools import wraps
+from itertools import starmap
+import json
+import hashlib
 
 from gitlabform.configuration import Configuration
 from gitlabform.configuration.core import ConfigFileNotFoundException
@@ -17,7 +20,7 @@ from gitlabform.gitlab.core import NotFoundException
 def if_in_config_and_not_skipped(method):
 
     @wraps(method)
-    def method_wrapper(self, project_and_group, configuration):
+    def method_wrapper(self, project_and_group, configuration, **kwargs):
 
         wrapped_method_name = method.__name__
         # for method name like "process_hooks" this returns "hooks"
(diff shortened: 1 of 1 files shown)


gitlabform/gitlabform/core.py
```diff
@@ -61,15 +64,46 @@ def get(self, path, default=None):
...
 
-        return method(self, project_and_group, SafeDict(configuration))
+        return method(self, project_and_group, SafeDict(configuration), **kwargs)
 
     return method_wrapper
 
 
 class GitLabFormCore(object):
 
+    def log_diff(self, subject, given, current, only_changed=False, hide_entries=None):
```
gdubicki: I see that you never call this method with non-default values of `only_changed` or `hide_entries`. Do you plan to add code that will use it soon? I am asking because if not then I suggest not adding code that will not be used.

gdubicki: I find names `given` and `current` not very clear. I suggest changing them to `from_config` and `from_server`, like in a comment a few lines below.

gdubicki:
Thanks a lot! This is a cool feature that I have been thinking of from the start (as it is a key part of terraform, which this project was inspired with) but had no time to implement it. I would love to get this merged after you a kindly consider applying code review comments.

gitlabform/gitlabform/core.py
akaRem: a. Yes, there are some sensitive variables which should be hidden. I played around with it, but didn't use for now. Same for only changed. It should be set to true by default, but for development/debugging it's better to have full picture.
b. `from_config` and `from_server` sounds better, indeed. I will fix it

gdubicki: Hey @akaRem , can you please find the time to make those small fixes? I would really like to merge your PR. :)

gdubicki: Ping @akaRem . :) PS It's a coincidence that exactly one more month has passed. It's NOT like I am doing this on a schedule. Really.

gdubicki: Hey, @akaRem ! It would be cool if you could find the time to fix the minor things we discussed above. Especially as we have another person interested in this PR: https://github.com/egnyte/gitlabform/issues/61

glensc: super friendly ping!

gdubicki: Perhaps you could take this PR over, @glensc ? :)

gdubicki:
Thank you for your contribution, even though I haven't finished working on this, @akaRem . The code from this PR has been reused in #129 and you will be credited in the release notes. :) Closing this PR, please see #129 instead.

PR #49, opened 2019-05-11
recursively find groups / subgroups

therealgambo:
This PR adds recursive lookup functionality to find a group, or sub group id. We needed this functionality so that we could correctly assign merge request approvers that are in a subgroup.

Covers the following group structures:
- `group`
- `group/[subgroup/...]` (recursively)

This maintains existing functionality, while allowing to find any sub group, no matter how deep it is in the path. Probably not the most performant, as it is still relying on the `groups?search=...` api initially.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -97,7 +97,9 @@ group_settings:
         - user1
         - user2
       approver_groups:
-        - my_group1
+        - my_group
+        - my_group1/subgroup
+        - my_group2/subgroup/subsubgroup
 
     # keys names are branches names and values are as parameters described at
     # https://docs.gitlab.com/ee/api/branches.html#protect-repository-branch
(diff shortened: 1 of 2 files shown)


gdubicki:
Thanks for your contribution @therealgambo and sorry for a late response!

Your MR has inspired me to investigate if GitLab really does not have "get subgroup id" API ..and it does have it.

But it's probably my fault that you didn't notice that because of my hacky `_get_group_id` implementation and comments suggesting that this is the only way to do it, so I am sorry.

Actually it's clearly documented in https://docs.gitlab.com/ee/api/groups.html#details-of-a-group that you can provide group OR subgroup full path as `id` request parameter and it actually works.

To make it up to you I have just pushed a fix to gitlabform and will release v.1.5.0 with the feature you need in a moment.

therealgambo:
Ahh looks like I overlooked that functionality as well 😂 - Looking forward to the fix!

gdubicki:
Please check out v1.5.0 which has been released yesterday then. :)

PR #51, opened 2019-05-28
Use full_path to find groups

kumy:
I think this will help for #6. @gdubicki what do you think?

I don't know when this `attribute` was added, sorry.


gdubicki:
Thank you for this contribution, but please see https://github.com/egnyte/gitlabform/issues/6#issuecomment-499632832 .

PR #52, opened 2019-06-06
Add full subgroups support

gdubicki:
This PR adds:
* support for processing recursively all projects within group if it contains subgroups with other projects,
* support for configuring group-level and subgroup-level configs - they will be merged.

### DIFF
--- a/gitlabform/configuration/projects_and_groups.py
+++ b/gitlabform/configuration/projects_and_groups.py
@@ -18,61 +18,86 @@ class ConfigurationProjectsAndGroups(ConfigurationCore):
 
     def get_effective_config_for_project(self, group_and_project) -> dict:
         """
-        :param group_and_project: "project_name/project_group"
-        :return: merged configuration for this project, from common, group and project level. Merging is additive.
+        :param group_and_project: "project_group/project_name"
+        :return: merged configuration for this project, from common, group/subgroup and project level.
+                 If project belongs to a subgroup, like "x/y/z", then it gets config from both group "x" as well
+                 as subgroup "y".
+                 Merging is additive.
         """
-        try:
-            project_config = self.get("project_settings|%s" % group_and_project)
-        except KeyNotFoundException:
-            project_config = {}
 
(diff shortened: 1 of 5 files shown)


gdubicki:
Can review this MR please, @mkjmdski ?

gitlabform/configuration/projects_and_groups.py
```diff
@@ -18,61 +18,89 @@ def get_projects(self) -> list:
...
+        #                    \------>b = effective config to return
+        #
+        # ..where a = merged_config("x", "x/y") and b = merged_config(a, "x/y/z")
+        #
+
+        effective_config = {}
+        elements = subgroup.split('/')
+        last_element = None
+        for element in elements:
+            if last_element:
```
mkjmdski: I think that negating this if condition to `if not last_element:` would improve code readability because it will cover first iteration

gitlabform/configuration/projects_and_groups.py
```diff
@@ -18,61 +18,89 @@ def get_projects(self) -> list:
...
             else:
-                # overwrite common settings with group settings
-                group_config[key] = {**common_config[key], **group_config[key]}
+                effective_config = self.get_group_config(element)
+                logging.debug("First level config for '%s': %s" % (element, effective_config))
+                last_element = element
+
+        return effective_config
+
+    def merge_configs(self, more_general_config, more_specific_config) -> dict:
```
mkjmdski: maybe call them just `general_config` and `specific_config`?

gdubicki: Just "general" and "specific" seem to be too absolute terms here.

For example if we have configs for group X, its subgroup Y and in it a project Z, then we first merge configs X and Y and then the result with config for Z. So X has "more general" config and Y has "more specific", but then the result of their merge becomes "more general" one.

gitlabform/configuration/projects_and_groups.py
gdubicki: Thanks, corrected.

Issue #31, opened 2018-10-11
Package for Linux, Mac and maybe Windows


gdubicki:
See #30 for rationale.

gdubicki:
I think that we can close this now that we have Docker images build along with each new Python package. Thanks @mkjmdski !

Issue #30, opened 2018-09-13
installation problem on MacOS due to fixed PyYaml version


weakcamel:
When installing gitlabform on MacOS (High Sierra), I'm facing problems.

As it happens, the specified version (3.12) is affected by problems described in https://github.com/yaml/pyyaml/issues/15

therefore *gitlabform* fails to install with pip as well.

The solution seems to be simple - update the dependency to pyyaml 3.13 which solves the problem, or don't requre a specific version in the first place.

weakcamel:
Hello,
I'm happy to provide a PR if need be; just checking though - I'm not sure how active this project is; is there a chance that it would be accepted / released?

gdubicki:
`gitlabform` does install on my MacOS High Sierra (10.13.6 (17G65)) without issues:

```
(test) gdubicki@mac ~ $ pip install gitlabform
Collecting gitlabform
  Using cached https://files.pythonhosted.org/packages/cf/bb/602df59fdd89213901196ad9220eb2fc28830ff5abed520780bf9da7cf06/gitlabform-1.0.0-py3-none-any.whl
Collecting pyyaml==3.12 (from gitlabform)
Collecting requests==2.18.4 (from gitlabform)
  Using cached https://files.pythonhosted.org/packages/49/df/50aa1999ab9bde74656c2919d9c0c085fd2b3775fd3eca826012bef76d8c/requests-2.18.4-py2.py3-none-any.whl
Collecting certifi>=2017.4.17 (from requests==2.18.4->gitlabform)
  Downloading https://files.pythonhosted.org/packages/df/f7/04fee6ac349e915b82171f8e23cee63644d83663b34c539f7a09aed18f9e/certifi-2018.8.24-py2.py3-none-any.whl (147kB)
    100% |████████████████████████████████| 153kB 2.8MB/s
Collecting urllib3<1.23,>=1.21.1 (from requests==2.18.4->gitlabform)
  Using cached https://files.pythonhosted.org/packages/63/cb/6965947c13a94236f6d4b8223e21beb4d576dc72e8130bd7880f600839b8/urllib3-1.22-py2.py3-none-any.whl
Collecting chardet<3.1.0,>=3.0.2 (from requests==2.18.4->gitlabform)
  Using cached https://files.pythonhosted.org/packages/bc/a9/01ffebfb562e4274b6487b4bb1ddec7ca55ec7510b22e4c51f14098443b8/chardet-3.0.4-py2.py3-none-any.whl
[...]

PR #54, opened 2019-06-09
Docker support

mkjmdski:
This code is ready but still we need to provide way for automated docker builds.


mkjmdski:
This solves problems mentioned in: #30 #31 allowing MacOS users to run this is a binary packaged in docker

PR #55, opened 2019-06-09
Add jinja2 filesystem loader

mkjmdski:
This pull request introduces support for jinja2 filesystems what allows gitlabform to read files within repository and use them as a templates. Beside this I have also added more verbosity to error message when user forgets to provide his gitlab access token what could be treated as a bugfix.

### DIFF
--- a/gitlabform/gitlab/core.py
+++ b/gitlabform/gitlab/core.py
@@ -20,12 +20,29 @@ s.mount('https://', HTTPAdapter(max_retries=retries))
 class GitLabCore:
 
     url = None
-    __token = None
+    token = None
+
+    # This functions sets variable from configuration,
+    # but if key is not set, it uses same variable from environment
+    # in convention "GITLAB_KEYUPPERCASE"
+    # if none value was found exception is raised
+    def __init_attr_with_env_default(self, configuration, attr_name, root_key='gitlab'):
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -7,7 +7,7 @@ def convert_markdown_to_rst(file):
 
 
 setup(name='gitlabform',
-      version='1.5.0',
+      version='1.5.1',
```
gdubicki: This PR introduces 2 new features, so definitely not patch but minor version should be update. I suggest `1.6.0` then.

gitlabform/gitlab/core.py
```diff
@@ -20,12 +20,29 @@
 class GitLabCore:
 
     url = None
-    __token = None
+    token = None
+
+    # This functions sets variable from configuration,
+    # but if key is not set, it uses same variable from environment
+    # in convention "GITLAB_KEYUPPERCASE"
```
gdubicki: Can you please add info about this feature to `config.yml` too? It is the main source of information about features and how to configure gitlabform for the users, they should not have to read the code.

setup.py
gdubicki: ...but it would be even better to not update version in the PR. We will update it directly in master during release of new version.

PR #60, opened 2019-08-29
Added support for group variables

SilverXXX:
This PR add support for group secret variables, with minimal modifications.


testconfig.yml
```diff
@@ -0,0 +1,51 @@
+# GitLab API access config
```
gdubicki: This file seems to be added to this MR by mistake so I suggest removing it.

gitlabform/gitlabform/core.py
```diff
@@ -320,6 +323,25 @@ def process_secret_variables(self, project_and_group, configuration):
 
         logging.debug("Secret variables AFTER: %s", self.gl.get_secret_variables(project_and_group))
 
+    @if_in_config_and_not_skipped
+    def process_group_secret_variables(self, group, configuration):
```
gdubicki: Please make the logging messages in this method different than the ones used for project level secret variables.

gdubicki: For example `logging.debug("Group secret variables BEFORE: %s", self.gl.get_group_secret_variables(group))`.

PR #65, opened 2019-10-31
Adding python37 and python38 to tests and package information

dom-nie:
Added python37/python38 to tests and python38 to supported versions.
Removing debug from pytests and setting ignore on deprecation warnings which usually come from third party packages.


gdubicki:
Thanks for the MR, @dom-nie !

Issue #68, opened 2019-11-12
MR approvers user/group has no effect (REST API change?)


weakcamel:
Applying a configuration using `approvers` and `approver_groups` has no effect while running against Gitlab 12.3.5-ee.

The config I'm using:

```
group_settings:
  'very-silly-test-group-for-ldap-and-fun':
    merge_requests:
      approvals:
        approvals_before_merge: 1
        reset_approvals_on_push: false
        disable_overriding_approvers_per_merge_request: false
        merge_requests_author_approval: true
      approver_groups:
        - gitlab
      approvers:
        - waldekm
        - foouser
```

With the log
```
$ gitlabform -vvv -c config.yml very-silly-test-group-for-ldap-and-fun
Reading config from: /Users/waldekm/git/config/config.yml
Connected to GitLab version: 12.3.5-ee (9dbaa740018)
Config file is declared to be compatible with GitLab API v4
Reading config from: /Users/waldekm/git/config/config.yml
*** # of groups to process: 1
*** # of projects to process: 1
* [1/1] Processing: very-silly-test-group-for-ldap-and-fun/testy-project
Setting project_settings
Setting project settings: {'only_allow_merge_if_pipeline_succeeds': True, 'only_allow_merge_if_all_discussions_are_resolved': True, 'shared_runners_enabled': True, 'request_access_enabled': True, 'wiki_enabled': False, 'snippets_enabled': False}
Setting merge_requests
Setting approvals settings: {'approvals_before_merge': 1, 'reset_approvals_on_push': False, 'disable_overriding_approvers_per_merge_request': False, 'merge_requests_author_approval': True}
Setting approvers to users ['waldekm', 'foouser'] and groups ['gitlab']
Setting hooks
```

Gitlabform implementation uses endpoint `projects/:id/approvers` which has been deprecated:
https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-allowed-approvers

I could just be using it wrong (although I've tried moving the key `approvers` **under** `approvals` and this didn't work either). It might be also that the deprecated endpoint doesn't work anymore.  In any case, adding/removing users to the `approvers` list and running gitlabform doesn't show in `gitlab-rails/api_json.log`

smcgregor83:
I am seeing the same behaviour running against 12.4.2-ee.

weakcamel:
The new API - if powerful - is sadly much more complex and involves creating/updating *approval rules* (which have their own ids). 

[...]

PR #71, opened 2019-12-11
[wip] Add option to enforce membership of a project

kassybas:
Previously when removing a group from the `group:` part of the configuration, it was not removed from GitLab.

With `enforce_group_shares` option set to true, you can ensure that only the defined groups can have access to the projects.


gdubicki:
Hey @kassybas ! Thanks for working on this contribution. Please ping me if I can help you move this forward in any way!

gdubicki:
Hey @kassybas ! Please see https://github.com/egnyte/gitlabform/pull/83 as it relates to what you are working on here. Note that you will most probably need to rebase because of that, sorry!

version
```diff
@@ -1 +1 @@
-1.9.1
\ No newline at end of file
+1.9.2
```
gdubicki: Please do not uptick the version in PRs. See https://github.com/egnyte/gitlabform/commit/87819c313f36748d988fe3a08b862da4112a4661 for a rationale.

gdubicki:
Hi @kassybas ! Because of #68 we badly needed better management of group membership, because it's used as a workaround for the GitLab bug (see https://github.com/egnyte/gitlabform/issues/68#issuecomment-581003703). Therefore I have just implemented something similar to what you did here and am releasing it in v1.13.0.

However you implementation looks nicer so it would be cool if you would complete this MR and get this merged after rebasing!

Please ping me if you need any help with that.

kassybas:
Hi @gdubicki,
I started working on rebasing to your changes, but since then, GitLab made it possible to share groups with other groups, so it is not required for us to use Gitlabform. 

Using the Gitlab Terraform provider can handle the necessary configurations, and Terraform fits our current toolchain better.

Issue #72, opened 2020-01-01
Add ~integration~ acceptance tests using GitLab instance in Docker


gdubicki:
Lack of tests in this app is something I am ashamed of for a long time now. I used to excuse myself by stating that running Gitlabform at Egnyte for >800 repositories every day in our CI is some way of testing it.

And partially it is but it does not detect all issues.

Some issues are related to things that used to work stop working and we do not notice that because we did not reconfigure repos to a different config to notice lack of change. This is is the case of not noticing issues like #68.

Other issues are problems related to apparent GitLab API regressions. This is the case of #69.

Therefore I would like to add ~integration~ acceptance tests that will:
* spin up GitLab instance in a Docker container and provide a basic set up,
* for each feature of Gitlabform test if it works to: 1) enable it, 2) reconfigure it, 3) disable it.

Apart from running them after each commit I want to set up a schedul to run it every day or every week to detect problems caused by changes in the new GitLab versions.

*UPDATE*: I was pointed out that these are not integrations tests but rather acceptance tests. Thanks for that!

gdubicki:
I have started working on it here https://github.com/egnyte/gitlabform/pull/75/files but it seems that I cannot set up GitLab EE instance automatically without some manual operation to start a trial:

<img width="1607" alt="Screenshot 2020-01-01 at 23 18 48" src="https://user-images.githubusercontent.com/566632/71646812-0cb0b500-2ced-11ea-9732-b6cb277d0263.png">

If I won't be able to work around it then it will be sad because it will mean that I actually can't work on #68 with the help of automated integration tests... :(

gdubicki:
I am talking to GitLab and am trying to find a solution for the above problem.

gdubicki:
I created a WIP (Work In Progress) PR here: https://github.com/egnyte/gitlabform/pull/75 . Interestingly, the API tests for MR approvers on EE without a license seem to work on API level - it is just that the effects are not visible in the web UI.

gdubicki:
I propose to review and merge above PR first and then create separate issues to work on adding missing test coverage.

gdubicki:
On second thought: we have added the tests, so we can close this ticket.

Issue #70, opened 2019-12-11
Some services always apply certain events as true


Dridge:
When adjusting the Jira service the commit_events option is always true, and has to be set to false manually through the UI, the same is true for the mattermost service and the confidential_issues_events option.

```
project_settings:
    "group/subgroup/project":
      services:
          mattermost:
              active: true
              webhook: https://mattermost.com/hooks/xxx
              username: gitlab
              merge_requests_events: true
              merge_request_channel: "merge-requests"
              push_events: false
              issues_events: false
              confidential_issues_events: false # This doesnt appear to work, is always true, in all cases
              tag_push_events: false
              note_events: false
              confidential_note_events: false
              pipeline_events: false
              wiki_page_events: false
              branches_to_be_notified: "all"
          jira:
              active: true
              url: http://jiralo.exfo.com
              username: "user1"
              password: "password"
              merge_requests_events: true
              commit_events: false # This doesnt appear to work, is always true, in all cases
```

gdubicki:
Sounds like another GitLab API bug... See also what I wrote in #66.

gdubicki:
My new integration tests (see link above) confirm that there is an issue with JIRA service but not with Mattermost.

Can you verify with GitLab 12.6.2, @Dridge ?

gdubicki:
While checking this I found out that there is a problem with setting `push_events` field using GitLab API too. You cannot set it to `false` when you create the service (its being created with it set to `true`) and when you have a service with it set to `true` you cannot change it to `false`. I reported a bug for this in GitLab here: https://gitlab.com/gitlab-org/gitlab/-/issues/209343

gdubicki:
For the `commit_events` problem I have found an existing GitLab issue here: https://gitlab.com/gitlab-org/gitlab/-/issues/202216

gdubicki:
Thanks to [Vincent Bisserié's help](https://gitlab.com/gitlab-org/gitlab/-/issues/202216#note_301934216) (if you are reading this - thank you!) I fixed the root cause and released v1.13.5 where this issue should be completely resolved without any workarounds.

Please reopen this issue if it does not help you, @Dridge .

Issue #69, opened 2019-12-11
Re-running gitlabform with changes to jira services doesn't apply any change


Dridge:
I have noticed that the password variable set in the repo CI/CD variables can't be used in the config.yml, but even when I make changes to my JIRA service settings in the config.yml the jira service remains unchanged. Do you need to manually delete the settings? Or can they be deleted before being re-applied with new settings?

Generally struggling with getting these services set up correctly. I have to replace the password below in order to avoid leaving the password in the config.yml, and replace it at the gitlab-ci.yml level with: `sed -i -e 's/SED_REPLACE_JIRA_PASSWORD/'"$JIRA_PASSWORD"'/g' config.yml`.

My services configuration, changing the password doesn't seem to get applied, nor does changes to the commit events from true to false.

>     "group/subgroup/projectname":
>       services:
>           mattermost:
>               active: true
>               webhook: https://mattermost.com/hooks/xx9bnr7jijndtgangda64w
>               username: gitlab
>               merge_requests_events: true
>               merge_request_channel: "merge-requests-channel"
>               push_events: false
>               issues_events: false
>               confidential_issues_events: false
>               tag_push_events: false
>               note_events: false
>               confidential_note_events: false
>               pipeline_events: false
>               wiki_page_events: false
>               branches_to_be_notified: "all"
>           jira:
>               active: true
>               url: http://jiralo.xxx.com
>               username: "botaccess"
>               password: "SED_REPLACE_JIRA_PASSWORD"
>               merge_requests_events: true
>               commit_events: true

gdubicki:
Hi @Dridge !

gitlabform just goes through each service defined in this config sections and makes PUT requests to GitLab API. So both password change and commit events change flag flip should work.

But I have checked that and it seems that there is a bug in GitLab API that makes those PUT requests not work, at least for JIRA service.

We could try a workaround of deleting and recreating services with API on each gitlabform run but I am afaid of the side effects (and a bit about performance). I have quickly tried making a diff between current and target service config (or, to be precise, a check for target config being subset of the current config) to check if a PUT request is needed but for example for Jira service the GitLab API does not return password in response (which make sense, of course) what makes this diffing ineffective. Maybe we should make this "always recreate services" a default off feature (or rather a workaround)?

Dridge:
> Maybe we should make this "always recreate services" a default off feature

This sounds reasonable to me, and the only way to get around the gitlab api bug.

gdubicki:
[...]

PR #75, opened 2020-01-01
Add integrations tests

gdubicki:
Although not all of the tests are passing yet because of issues #68, #69 and #70, I would like to add them to the master branch first and then work on fixing those issues in separate branches and PRs.

There will also be more PRs from original issue #72 to add integration tests to all other features than MR approvals and services.


gitlabform/configuration/core.py
```diff
@@ -1,37 +1,55 @@
...
-            # we need config path for accessing files for relative paths
-            self.config_dir = os.path.dirname(config_path)
+        try:
+            if config_path:
+                # using this env var should be considered unofficial, we need this temporarily
+                # for backwards compatibility. support for it may be removed without notice, do not use it!
+                if 'APP_HOME' in os.environ:
+                    config_path = os.path.join(os.environ['APP_HOME'], 'config.yml')
+                # this case is only meant for using gitlabform as a library
+                elif not config_path:
```
dorotawm: will the code enter here? It is under `if config_path` anyway

gitlabform/gitlabform/core.py
```diff
@@ -69,9 +74,26 @@ def method_wrapper(self, project_and_group, configuration):
...
+    def __init__(self, project_or_group=None, config_string=None, debug=False):
+
+        if project_or_group and config_string:
+            self.project_or_group = project_or_group
+            self.config_string = config_string
+            self.verbose = False
+            if debug:
+                self.debug = True
+            else:
+                self.debug = False
```
dorotawm: can it be just one line `self.debug = bool(debug)` ? `debug` is a default argument set to `False` anyway so this covers only case when something else than boolean is passed to the init

gitlabform/gitlabform/core.py
```diff
@@ -102,26 +124,40 @@ def parse_args(self):
...
+            handler = logging.getLogger().handlers[0]
+            logging.getLogger().removeHandler(handler)
 
     def initialize_configuration_and_gitlab(self):
 
         try:
-            gl = GitLab(self.args.config.strip())
-            c = Configuration(self.args.config.strip())
+            if hasattr(self, 'config_string'):
+                gl = GitLab(config_string=self.config_string)
```
dorotawm: why don't you put both `gl = GitLab(config_string=self.config_string)` and `c = Configuration(config_string=self.config_string)` in this `if`?

gitlabform/gitlabform/test/test_merge_request_approvers.py
```diff
@@ -0,0 +1,234 @@
...
+        gf.main()
+
+        approvers = gitlab.get_approvers(GROUP_AND_PROJECT_NAME)
+        assert len(approvers['approvers']) == 2
+
+    def test__if_change_works__only_users(self, gitlab):
+        gf = GitLabForm(config_string=config__approvers_only_users, project_or_group=GROUP_AND_PROJECT_NAME)
+        gf.main()
+
+        # approvers = 2
```
dorotawm: maybe you can add assertion here instead of comment - `assert len(approvers['approvers']) == 2` - it can be a self-describing code and sanity check that it indeed changed.
The same for `test__if_change_works__only_groups`

gitlabform/gitlabform/core.py
gdubicki: Of course it can, thanks!

gitlabform/gitlabform/core.py
gdubicki: Done, thanks!

PR #78, opened 2020-01-15
Remove unused projec_key property from example config yaml

soczkers:


gdubicki:
On second look we were missing 3 other parameters requires for this integration, I have added them to master in my last commit a few minutes ago, @soczkers .

PR #79, opened 2020-01-17
Add separate contribution and implementation design guides

### LINKED ISSUE #74: Write a doc about contributing (with architecture, design and a code example)
I feel that if I document how this app is designed and provide a real-case example of how to add features to it, it will get much more PRs and with community help will improve much faster than I can do it on my own.

gdubicki:
Fixes #74.


gdubicki:
As no one has any objections, I am merging this.

PR #83, opened 2020-02-04
Add/change access level of the group members

gdubicki:
And also make it more precise in the docs that for project members we also only provide adding/changing access level but NO removal yet.

(https://github.com/egnyte/gitlabform/pull/71 is going to change that)


gdubicki:
I do not have active reviewers in this project so I am creating PRs such as this one only to show that the tests are passing.

Merging as they are.

PR #88, opened 2020-03-07
WIP: Example of a GitLab issue with JIRA and commit_events: false

gdubicki:


gdubicki:
Invalid, see https://github.com/egnyte/gitlabform/issues/70#issuecomment-596757507 .

Issue #92, opened 2020-04-08
401 Unauthorized during version test


mandoonandy:
Using a gitlab token (validated using curl) I get the following error:

```$ GITLAB_TOKEN="******************" gitlabform test
Aborting - GitLab test request failed, details: 'Request url='https://dev.xxxxx.com/api/v4/version', method=GET, data='None' failed - expected code(s) [200], got code 401 & body: 'b'{"message":"401 Unauthorized"}'''
```
From reading the code It appears that gitlabform sends the token using this a header with: 'Authorization': 'Bearer ' + self.token

Reading the issue:  https://gitlab.com/gitlab-com/support-forum/issues/2722

It would appear that the gitlabform code is creating the wrong header. It should be using a header containing "PRIVATE-TOKEN: **********************"

The line: https://github.com/egnyte/gitlabform/blob/84074ba2101a98849d61f2163361c0bc63daf531/gitlabform/gitlab/core.py#L149

should be changed to:
       `headers = {"PRIVATE-TOKEN": (self.token)}`

gdubicki:
Hi @mandoonandy !

I will check if auth still works on our instance with our auth with your PR merged and if so I will gladly merge it.

gdubicki:
Thanks for reporting the issue, @mandoonandy !

PR #93, opened 2020-04-08
Fix egnyte#92 auth header

mandoonandy:
As per https://gitlab.com/gitlab-com/support-forum/issues/2722
the auth header should use PRIVATE-TOKEN


gdubicki:
I have fixed #92 differently, so I am closing this. Thanks for the PR anyway!

Issue #27, opened 2018-06-26
Implement the "protected branches" API endpoint


born4new:
Hey there!

That endpoint would be pretty useful for us, as we want to have a branch on a project that should not be mergeable by anybody...

Sadly, I couldn't find a reference to it in the code. Are you planning on implementing it?

Cheers!

Gitlab Documentation: https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches

gdubicki:
I think that you misunderstood me: it's already implemented in gitlabform and I did it was when this API was not yet deprecated. :)

gdubicki:
Note to self: when implementing this we should take #10 into account.

gsemet:
"Protected tag" as well?

gdubicki:
Sadly I don’t have time to develop this myself currently, but we are open to accepting PRs with this.

therealgambo:
@gsemet - Ive implemented protected tag functionality in #39

gdubicki:
Note to self: we can implement this feature as:
- replacement for the current way of managing branch protection, including different syntax - this would require major version uptick,
- adding new way next to the old way with making the old one use the new API under the hood.

Of course the 2nd is more user-friendly but because we have too little maintainers of this app I would prefer the 1st as it requires less maintenance.

We can prevent the app from running ONLY IF user is using branch protection config and does not declare the config to be compatible with GitLabForm v. 2.*.

After we implement new approvers API in #68 we will probably have to do the same.

Ideally we should do both of these changes together to have only a single major version uptick here.

While creating v.2 we can think about more config syntax improvements. Let's discuss them in a separate issue.

PR #94, opened 2020-04-10
fix: #27 - Implement the "protected branches" API endpoint

rbartuzel:
Could you please take a look at my PR?

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -143,6 +143,12 @@ group_settings:
         developers_can_push: true
         developers_can_merge: false
       branch_protected_from_direct_pushes_from_no_one_but_with_merges_allowed:
+        protected: true
+        push_access_level: 0
+        merge_access_level: 30
+        unprotect_access_level: 40
+      '*-branch_protected_from_direct_pushes_from_no_one_but_with_merges_allowed_configured_using_wildcard':
+        protected: true
         push_access_level: 0
         merge_access_level: 30
         unprotect_access_level: 40
--- a/config.yml
+++ b/config.yml
@@ -142,6 +142,10 @@ group_settings:
         protected: true
         developers_can_push: true
         developers_can_merge: false
+      branch_protected_from_direct_pushes_from_no_one_but_with_merges_allowed:
+        push_access_level: 0
+        merge_access_level: 30
+        unprotect_access_level: 40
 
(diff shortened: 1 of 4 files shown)


gdubicki:
Hi @rbartuzel ! Thank you for the PR!

We are striving to add for all new/fixed features tests for initially configuring, changing and deleting given feature. Can you please add tests for changing branch protection level and for removing protection of a branch?

rbartuzel:
Hi! I have added tests for changing branch protection level and for removing protection of a branch.

wadimklincov:
For me using branch names with wildcards doesn't work, because it still uses the old endpoint for unprotecting branches.

gdubicki:
Thanks @wadimklincov ! 

@rbartuzel : how about we:

1. Make unprotect method call **both** old API (like it does now) and the new one (like @wadimklincov suggested) to ensure that given branch is really unprotected.

2. Make unprotect method calls accept 404 as the return code too (we already do that in a few other places, see https://github.com/egnyte/gitlabform/blob/master/gitlabform/gitlab/members.py#L18 as an example).

3. Add some more tests for unprotecting and protecting:

a) configure protection using old API and switch it to new,
b) the other way around,

?

I think we would have all cases covered then.

rbartuzel:
@gdubicki Sounds like a good idea, I'll try to fix it.

rbartuzel:
Hi!

I noticed that it's not required to call unprotect using both api method. Unprotect method from new api can also unprotect branches protected old way.

I modified one of existing test to cover use case with wildcards, and added new ones to cover use case with set protection using old api and updating with new api and vice versa.

If you have any comments or suggestions, just let me know.

wadimklincov:
Awesome, it works great!

I just have two things I noticed while testing:

* if `protected: false` is set, but the rest of the parameters are still specified, it's still protected
* if the branch is removed from the configuration, it still stays protected

I guess the latter one is difficult to do, though.

rbartuzel:
@wadimklincov thanks for testing!
I fixed first issue but in case of the second, I agree it could be difficult to do.

gdubicki:
LGTM. Thanks for implementation @rbartuzel and for the tests @wadimklincov !

PR #97, opened 2020-04-24
Fix interaction between new protected branch API and files

KrystianOcado:
This should fix: https://github.com/egnyte/gitlabform/issues/96

I've also added some tests.


gdubicki:
LGTM, thank you for your contribution @KrystianOcado !

Issue #110, opened 2020-06-28
Add delete variable action


ss7548:
Provide a method to delete secret variables from groups and projects
https://docs.gitlab.com/ee/api/project_level_variables.html#remove-variable
https://docs.gitlab.com/ee/api/group_level_variables.html#remove-variable

gdubicki:
Thanks @ss7548 ! Please see my review in the PR above.

gdubicki:
Added in v1.15.4.1 that has just been released (v.1.15.4 was not released as its build has failed). Thanks @ss7548 !

PR #111, opened 2020-06-28
Add feature (issue #110) delete secret variables

ss7548:
Add feature (issue #110) delete secret variables.  Provides ability to delete variables from groups or projects.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -43,7 +43,19 @@ common_settings:
       key: A_VERY_COMMONLY_USED_PASSWORD
       value: "ThisIsAVerySecretPassword"
       variable_type: env_var #or file
-      protected: false #
+      protected: "false" #
+
+  group_delete_secret_variables:
+    # If variables exist at the group level, they will be deleted
+    deleteThese:
+      - var1
+      - var2
+
+  delete_secret_variables:
+    # If variables exist at the project level they will be deleted
+    deleteTheseAlso:
+      - var3
+
 
 group_settings:
   # settings for ALL projects in 'my-group' group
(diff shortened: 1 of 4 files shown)


config.yml
```diff
@@ -43,7 +43,19 @@ common_settings:
       key: A_VERY_COMMONLY_USED_PASSWORD
       value: "ThisIsAVerySecretPassword"
       variable_type: env_var #or file
-      protected: false #
+      protected: "false" #
+
+  group_delete_secret_variables:
```
gdubicki, line 48: I suggest following the convention for deleting entities that we have for services (https://github.com/egnyte/gitlabform/blob/master/config.yml#L180) and hooks (https://github.com/egnyte/gitlabform/blob/master/config.yml#L194) - using the same config sections, but adding an extra dict element: `delete: true`

config.yml
```diff
@@ -43,7 +43,19 @@ common_settings:
       key: A_VERY_COMMONLY_USED_PASSWORD
       value: "ThisIsAVerySecretPassword"
       variable_type: env_var #or file
-      protected: false #
+      protected: "false" #
```
gdubicki, line 46: I think that this is a GitLab bug. :( But I would rather work around it in the code instead of asking users to do this counterintuitive "bool as string" thing. Can you code it please?

gdubicki: Also: I was not able to reproduce this issue, please see https://github.com/egnyte/gitlabform/issues/107#issuecomment-653876243 .

ss7548: I tested in the latest release and confirmed it works.

ss7548:
Replaced with pull request #119

PR #119, opened 2020-07-05
feature #111,  Issue #110 Add ability to delete variables

ss7548:


gitlabform/gitlabform/core.py
```diff
@@ -424,15 +434,27 @@ def process_secret_variables(self, project_and_group, configuration):
...
+                        logging.warn(f"Could not delete variable {key} in group {group}")
+                    return
+
             try:
                 self.gl.put_group_secret_variable(group, configuration['group_secret_variables'][secret_variable])
             except NotFoundException:
                 self.gl.post_group_secret_variable(group, configuration['group_secret_variables'][secret_variable])
 
         logging.debug("Groups secret variables AFTER: %s", self.gl.get_group_secret_variables(group))
+        logging.info("gRoups secret variables AFTER: %s", self.gl.get_group_secret_variables(group))
```
gdubicki, line 457: This line almost duplicates the one above so I would delete it. If someone needs to see group variables before and after run the they should use `-d` to enable the debug output.

gitlabform/gitlab/groups.py
```diff
@@ -58,6 +58,9 @@ def put_group_settings(self, project_and_group_name, group_settings):
         # ..as documented at: https://docs.gitlab.com/ee/api/groups.html#update-group
         self._make_requests_to_api("groups/%s", project_and_group_name, 'PUT', group_settings)
 
+    def delete_group_secret_variable(self, group, secret_variable_key):
+        self._make_requests_to_api("groups/%s/variables/%s", (group, secret_variable_key), method='DELETE', expected_codes=[204])
```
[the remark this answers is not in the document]
gdubicki: The same applies to the method for project-level secret variables deletion, of course.

gitlabform/gitlabform/core.py
ss7548: yeah, sorry left that in when debugging

gitlabform/gitlabform/core.py
```diff
@@ -411,9 +410,20 @@ def process_secret_variables(self, project_and_group, configuration):
...
 
+            if "delete" in configuration['secret_variables'][secret_variable]:
+                key = configuration['secret_variables'][secret_variable]["key"]
+                if configuration['secret_variables'][secret_variable]["delete"]:
+                    logging.debug(f"Deleting {secret_variable}: {key} in project {project_and_group}")
+                    try:
+                        self.gl.delete_secret_variable(project_and_group, key)
+                    except:
+                        logging.warn(f"Could not delete variable {key} in group {project_and_group}")
+                    return
```
gdubicki, line 425: I think you meant `continue`, not `return`, as we don't want to completely exit the function as we may have more secret variables to process in this `for` loop.

gitlabform/gitlabform/core.py
```diff
@@ -411,9 +410,20 @@ def process_secret_variables(self, project_and_group, configuration):
             return
 
         logging.debug("Secret variables BEFORE: %s", self.gl.get_secret_variables(project_and_group))
+
         for secret_variable in sorted(configuration['secret_variables']):
             logging.info("Setting secret variable: %s", secret_variable)
```
gdubicki, line 415: If we leave this line here then for a secret variable that is going to be deleted:
* in verbose mode you will only get a log output saying that it is going to be set,
* in debug mode you will get a log output saying that it is going to be set and then a line saying that is going to be deleted...

This will look misleading.

Therefore I suggest moving this line just above the code that actually does set/create the secret variable and changing the level in line 420 to `info`.

gdubicki: This applies to the code in group variables function too.

gitlabform/gitlabform/core.py
gdubicki: This applies to the code in group variables function too.

gitlabform/gitlabform/test/test_group_secret_variables.py
```diff
@@ -8,7 +8,7 @@
 GROUP_AND_PROJECT_NAME = GROUP_NAME + '/' + PROJECT_NAME
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="module")
```
gdubicki, line 11: This change breaks the tests in this class. After reverting it the tests pass on my machine.

gdubicki:
Thanks! I will release a new version with this feature within the next few days. :)

PR #129, opened 2020-09-29
Add simple diffing - modularisation precondition

kowpatryk:
Thought of taking over this PR: https://github.com/egnyte/gitlabform/pull/44/

I am aware that initially the idea was to have all the processing steps put into one file (what's even described [here](https://github.com/egnyte/gitlabform/blob/master/IMPLEMENTATION_DESIGN.md)... but I'd like to propose the change :) When it's accepted, I'm open to contribute to the linked design description and adjust it as well.

To justify my proposal - my arguments are:
* Adding new processor is pretty independent - the most intrusive part is to add it to the appropriate list of existing processors.
* Adding new functionality (such as logging of changes, started in this PR) is way easier as there is already a placeholder to do so.
* Navigation within the code should also be (IMO) clearer now.
* Common logic is way more centralised now (e.g. if I had to add change logging to all the processing even if dry-run is not enabled - there is only one place to change).

Please share your thoughts! I'd like to continue with adding logging here but would be happy to get intermediate feedback for this.


gitlabform/gitlabform/processors/util/difference_logger.py
```diff
@@ -0,0 +1,39 @@
+import hashlib
+import json
+import logging
+from itertools import starmap
+
+
+class DifferenceLogger(object):
+
+    @staticmethod
+    def log_diff(subject, from_config, from_server, only_changed=False, hide_entries=None):
```
kowpatryk, line 10: This part has been borrowed from https://github.com/egnyte/gitlabform/pull/44/

gdubicki:
Hey @kowpatryk , that looks like a fantastic contribution, thank you! ❤️ Sorry for a late reply, notifications from Github are failing on me again. 😓 

I am reviewing this PR now.

gitlabform/gitlabform/processors/group/group_members_processor.py
```diff
@@ -0,0 +1,121 @@
...
+        else:
+
+            logging.fatal(
+                "You cannot configure a group to have no members. GitLab requires a group "
+                " to contain at least 1 member who is an Owner (access_level = 50)."
+            )
+            sys.exit(4)
+
+    def _log_changes(self, group: str, group_members):
+        logging.info("Diffing for group_members section is not supported yet")
```
gdubicki, line 121: Nit:

Can't we somehow put "Diffing for \<this\> section is not supported yet" in the `AbstractProcessor` class?

What we currently have feels a bit copy-pasty to me.

gitlabform/gitlabform/processors/util/difference_logger.py
```diff
@@ -0,0 +1,53 @@
+import hashlib
+import json
+import logging
+from itertools import starmap
+
+
+class DifferenceLogger(object):
+    @staticmethod
+    def log_diff(
+        subject, from_config, from_server, only_changed=False, hide_entries=None
```
gdubicki, line 10: Nit:

I know that [me-in-may-2019 suggested `from_config` and `from_server` parameter names](https://github.com/egnyte/gitlabform/pull/44/files#r280068447) but me-in-october-2020 had no idea what they are about until I read what I wrote there... 😅 

Do you think that `before` and `after`, in this order, would be better as these parameter names, @kowpatryk ?

Or maybe you have a completely different proposal?

gdubicki (approved):
Yep, it is a great contribution indeed!

You have done what I have tried to do (big refactoring of the code, including splitting it) at least 2 times and never finished ...and thank goodness I didn't, because what you did is much better! 🥇 

I have just 2 [nitpicking](https://google.github.io/eng-practices/review/reviewer/looking-for.html#style) comments.

Big 👍 !

gitlabform/gitlabform/processors/util/difference_logger.py
kowpatryk: Hmm... how about `from_server` -> `current_config` and `from_config` -> `config_to_apply`? This would be quite descriptive what stands behind those parameters.

gitlabform/gitlabform/processors/group/group_members_processor.py
kowpatryk: You are right, though ultimately we do not want this to be printed anywhere. Moved to the `AbstractProcessor` and removed from classes which do not implement diffing (yet)

IMPLEMENTATION_DESIGN.md
```diff
@@ -51,21 +51,28 @@ and config should be taken into account, applies the effective config.
 The effective config contains what is called the "config sections", which are the YAML keys that can exist under projects
 in the config YAML, for example: `deploy_keys`, `secret_variables`, `group_secret_variables` and so on.
 
-Those config section are processed by the code in the methods with the `process_*` naming pattern.
+Those config sections are processed by the code in the classes inheriting from the `AbstractProcessor` class. This class
```
kowpatryk, line 54: Since we have an agreement, I've updated the implementation design to reflect the new structure of the code.

gitlabform/gitlabform/processors/util/difference_logger.py
gdubicki: Sure, that sounds good!

IMPLEMENTATION_DESIGN.md
gdubicki: 👍

kowpatryk:
@gdubicki I think this PR is ready for merge. Since I don't have access to merge actions, could you take the necessary steps in order to release the changes here? I would like to proceed with the other PR (and possibly start some work on proper diffing in dry-run mode).

PR #130, opened 2020-10-01
Added support for pipeline schedules

kowpatryk:
This implements https://github.com/egnyte/gitlabform/issues/66


gdubicki:
Thank you for this contribution too!

But I assume that now you'll want to merge #129 first and then refactor this PR to the new style, right?

kowpatryk:
Yup, now I know that modularisation is ok for you, I can adjust this PR to base on #129.

kowpatryk:
This one should be ready for merging once #129 is merged. I'm marking this PR as WIP to avoid accidental merging.

kowpatryk:
@gdubicki, this PR is now ready for review and possible merge.

gitlabform/gitlab/schedules.py
```diff
@@ -0,0 +1,117 @@
...
+    def get_pipeline_schedule(self, project_and_group_name, pipeline_schedule_id):
+        return self._make_requests_to_api(
+            "projects/%s/pipeline_schedules/%s",
+            (project_and_group_name, pipeline_schedule_id),
+        )
+
+    def create_pipeline_schedule(
+        self, project_and_group_name, description, ref, cron, optional_data=None
+    ):
+        if optional_data is None:
```
gdubicki, line 19: Nit: isn't it enough to set the `optional_data` default value to `{}`? I think that then we can get rid of this `if`.

The same would apply to similar cases below.

gitlabform/gitlabform/processors/project/schedules_processor.py
```diff
@@ -0,0 +1,127 @@
...
+                variable_data,
+            )
+
+    @staticmethod
+    def __group_schedule_ids_by_description(schedules) -> Dict[str, List[str]]:
+        schedule_ids_by_description: Dict[str, List[str]] = {}
+
+        for schedule in schedules:
+            description = schedule["description"]
+            if description in schedule_ids_by_description:
```
gdubicki, line 122: Nit: you can set dict's default value for non-existent key to an empty list `[]` to avoid this `if`. See https://github.com/egnyte/gitlabform/blob/master/gitlabform/gitlabform/processors/group/group_members_processor.py#L28-28 .

gdubicki (changes requested):
I have only 2 nit-picking comments, feel free to do something about them or not.

I just also a bit surprised that setting the schedule variables works in case they are not defined in the config, as `configuration.get("schedules|" + schedule_description + "|variables")` returns `None` in this case and you cannot do `.items()` on it... But the test passes so it does work. :)

Otherwise the PR looks very good. :)

Can you just please resolve the conflicts?

gitlabform/gitlab/schedules.py
kowpatryk: This is here because IDE suggested to me that having mutable arguments is actually quite error-prone pattern.

Leaving explanation example for your reference: https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

gitlabform/gitlabform/processors/project/schedules_processor.py
kowpatryk: Ah, that's a method I missed in Python quite a few times :) I've adjusted the code.

kowpatryk:
I've responded to both comments of yours. I cannot see any conflicts though so I think the branch is now ready to be merged as soon as all CI jobs get green :)

PR #138, opened 2020-11-21
WIP: Proposal of the v2 config syntax

gdubicki:
See https://github.com/egnyte/gitlabform/issues/82 for more info.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -2,6 +2,10 @@
 # Please remove it if you create your config as a copy of this file.
 example_config: true
 
+# This indicates that this config file is compatible with GitLabForm 2.0+.
+# You should add it to your config after you update it to make it 2.0+ compatible.
+config_version: 2
+
 # GitLab API access config
 gitlab:
   # You can also set in your environment GITLAB_URL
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -86,16 +94,13 @@ If so then:
...
   token: "<private token of an admin user>"
-  api_version: 4
 
-group_settings:
-  my-group:
+projects_and_groups:
+  my-group":
     deploy_keys:
-      a_friendly_deploy_key_name:  # this name is only used in GitLabForm config
+      a_friendly_deploy_key_name: # this name is only used in the config
```
glensc, line 103: I prefer never to have code and comment on the same line so that if you modify comments, you don't alter lines with code. you will feel the difference when you want to `git blame` to find the commit that introduced the change for code.

gdubicki: Thanks, your advice definitely makes sense, @glensc ! I do care about `git blame` (even so much that I discouraged people from making formatting-only changes because I was not aware of the fact that [it can ignore whitespace changes and moving things around](https://coderwall.com/p/x8xbnq/git-don-t-blame-people-for-changing-whitespaces-or-moving-code)...) so I will try to fix this config and apply this in other places from now on. :)

glensc: also, recommended reading:
- https://www.freshconsulting.com/atomic-commits/
- https://chris.beams.io/posts/git-commit/

config.yml
```diff
@@ -1,98 +1,121 @@
...
-        value: "LS0tLS1CRUdJTiBSU0EgUFJJVkFURSBLRVktLS0tLQpNSUl (...)"
+# The main config starts with this key
+projects_and_groups:
+
+  # On this indentation level you choose what groups and/or projects you want to configure.
+  #
+  # Here are examples of possible dict key.
+  #
+  # This matches all the groups AND all the projects EXCEPT personal projects!
+  # "*":
```
weakcamel, line 60: Pattern matching would make the syntax very powerful indeed. The one potential issue I see here is: what if there are 2+ different patterns matching the same project or same group?

* would this be considered a "fatal error"?
* if not, which pattern wins?

config.yml
```diff
@@ -1,98 +1,121 @@
...
+  # This matches the "my-group" group AND projects in it.
+  # my-group:
+  #
+  # This matches the "my-subgroup" subgroup within the "my-group" group AND projects in it.
+  # my-group/my-subgroup:
+  #
+  # This matches all the groups that have names matching the '(foo)+' regexp AND projects in it.
+  # "~(foo)+":
+  #
+  # This matches all the subgroups of group "my-group" that have names matching the '(bar)+' regexp
```
weakcamel, line 74: > subgroups of group "my-group" that have names matching the '(bar)+' regexp  AND projects in it.

I'm guessing **projects** directly under `my-group` with matching name would be included here as well?

And if not, how could one express it?

config.yml
```diff
@@ -1,98 +1,121 @@
...
+  # This matches all the groups AND all the projects EXCEPT personal projects!
+  # "*":
+  #
+  # This matches all the personal projects.
+  # "<all-users>":
+  #
+  # This matches the "my-group" group AND projects in it.
+  # my-group:
+  #
+  # This matches the "my-subgroup" subgroup within the "my-group" group AND projects in it.
```
weakcamel, line 68: will this be recursive? would this also affect `my-group/my-subgroup/my-sub-subgroup`?

config.yml
gdubicki: Thanks for pointing this issue out. We haven't made a decision yet, but we will define some behaviour and write tests to ensure that it will be stable.
And what would be your suggestion?

weakcamel: I actually quite liked the previous 3-level priority: common, overridden on group level, overridden on project level. 
This was simple and IMO very effective.

To make it work with pattern matching, I suppose something like this should work:

```yaml
projects_and_groups:
  - path: '*'   # equivalent of "common_settings"
    group_settings:
      ...
    project_settings:
  - path: 'foogroup/*'   # equivalent of group settings (although it doesn't make it clear whether this includes sub-groups or not)
    group_settings:
      ..
  - path: 'bargroup/quxproject'  # and this is an override per project
    project_settings:
      ..
```

Those could be processed from top to bottom, with the settings at the bottom "winning" to allow to merge them.

A more powerful option (but also even more complicated):

```yaml
projects_and_groups:
  - path: '*'   # equivalent of "common_settings"
    continue: true #### <<---- let user select whether this rule should end processing or not
    group_settings:
      ...
    project_settings:
  - path: 'foogroup/*'   # equivalent of group settings (although it doesn't make it clear whether this includes sub-groups or not)
    continue: false #### <<---- let user select whether this rule should end processing or not
    group_settings:
      ..
  - path: 'foogroup/quxproject'  # and this is an override per project 
    project_settings:
      ..
```
```

config.yml
kowpatryk: That's a good question. What would be expected behaviour for you? I am currently thinking about few possible approaches:
1. Keep is as it is - no recursion. This means though that all subgroups would have to be explicitly specified (and it weakens the power of pattern matching that we would like to introduce here).
2. Make it non-recursive by default and allow recursive search to be enabled (either only on single pattern level, globally, or both).
3. Make it always recursive (with option to disable it? not sure if it makes sense though).

config.yml
kowpatryk: Another good point that makes me think we should really consider going toward recursive search...

config.yml
kowpatryk: That's a nice proposal! I believe with tree-like structure of GitLab groups and projects, this could cover most (if not all) of the cases - especially assuming that patterns are defined in the topological order (with patterns affecting more top-level groups/projects defined before the other patterns)

config.yml
weakcamel: Another option - a `recursive: true/false` property to be defined on each level.

This would allow to e.g. include only a subgroup with sub-subgroups without having to resort to a *list* of patterns.

Argh, too many options :)

config.yml
weakcamel: Wondering about pattern matching: unless resorting to "regex", it could be helpful to intrduce a distinction between `/*` and `/**` patterns where the former would mean only direct descendents of a group and the latter a recursive match.

Rationale: if recursive were to be default, it would be quite difficult to partition the groups sensibly (make any form of exclusion).

gdubicki: Hey @weakcamel! I think that it would be more efficient to discuss and come up with the best solution using our new Slack workspace. If you (or anyone else reading this and interested in it) want to join it the please use this invite link: https://join.slack.com/t/gitlabform/shared_invite/zt-ljgzvxk0-TubjJFA5lLpRkyivQWI_Qg (valid until ~21 Feb).

gdubicki:
The list of issues and PRs that require, or would benefit from the most, introducing backward incompatible changes to v1 got pretty long recently.

That’s why I plan to introduce only some changes to the syntax in v2.0 to release it faster and open the possibility to add more features based on it in v2.1+.

PR #144, opened 2020-11-24
Fix project settings save by using json payload

glensc:
Fixes https://github.com/egnyte/gitlabform/issues/114


gdubicki:
Please reformat using black. :)

glensc:
I hate those code reformatters on line length, I should be the one who decides which one is more readable. The current diff (9b00d675919a7673590d20da91412f710844b317) is more readable than every option spread to its own line (61e586c). These also have a tendency to create un-necessary diffs when attempting to run `git blame` just because the line length is 1 byte off with some change.

gdubicki:
Thanks for the contribution, @glensc ! We'll make a release with this fix within the next few days.

PR #145, opened 2020-12-01
Add example for container_expiration_policy_attributes

glensc:
Followup to https://github.com/egnyte/gitlabform/issues/114

The example keys and values are taken from GitLab docs:
- https://docs.gitlab.com/ee/user/packages/container_registry/#use-the-cleanup-policy-api


gdubicki:
Thank you, @glensc ! I was just thinking that we should also have an example how to configure this. :)

Issue #148, opened 2020-12-10
bug(pipeline_schedules) - Unexpected return code 202 on delete


grzesuav:
`
gitlabform.gitlab.core.UnexpectedResponseException: Request url='some_url/pipeline_schedules/1875/variables/SCHEDULE', method=DELETE, data='None' failed - expected code(s) [200, 201, 204, 404], got code 202 & body: 'b'{"variable_type":"env_var","key":"somekey","value":"somevalue"}''
`

grzesuav:
https://github.com/egnyte/gitlabform/pull/149

kowpatryk:
PR has been merged. I will close the ticket as soon as the fix gets rolled out in the next release.

gdubicki:
Technically fixed in v.1.19.0 but the first PyPI+DockerHub images release with this is v.1.20.0.

PR #149, opened 2020-12-11
fix(schedules #148): Add missing return code handling to delete_pipeline_variable

grzesuav:


grzesuav:
@gdubicki it should do its job

gdubicki:
It should, but can you add a test for it please? It’s the lack of tests that put us into this little mess. :(

grzesuav:
@gdubicki how to  run test at this PR ? I modified test and removed fix to check if it will fail

gdubicki:
We would have to create a schedule first as another user than root (which is used for running GitLabForm in the tests by default) and then run the test to see if the take ownership API endpoint would work... So we would have to do something new in our tests.

The problem with this is that I am not sure if we can even reproduce the 202 status code as it seems that sometimes GitLab performs operations immediately and returns 200 or 201 but sometimes it needs more time and returns 202 "promising" that it will eventually do the requested operation.

So maybe trying to reproduce it in test is not even worth it.


But after all this fighting with the status code recently I am thinking that perhaps we should make a bigger change and just start accepting all 2xx codes by default for all methods + also 404 for DELETEs..?

gdubicki:
Oh, and your code actually did test in https://travis-ci.com/github/egnyte/gitlabform/builds/207940385 but Travis wants to ~exto~make us pay for testing and building this project in a very unpleasant way:

Screenshot 2020-12-17 at 19 50 25

...and clicking on "Consume..." link gives:

Screenshot 2020-12-17 at 19 58 30

🤷‍♂️

grzesuav:
@gdubicki I can try to add support for all 2xx codes by default, is it the way we want to proceed ?

kowpatryk:
@grzesuav I believe this is something we could consider for adding in the next major version, the soonest. However I would be in favour of narrowing down the scope of this fix - and I assume adding just *202* response code would be sufficient here :)

grzesuav:
@gdubicki @kowpatryk restored to first version then

Issue #162, opened 2021-01-15
bug(pipeline_schedules) - can't create schedule without variables


trzeciakp:
I get an error:
```
  File "/gitlabform/gitlabform/gitlabform/processors/project/schedules_processor.py", line 49, in _process_configuration
    self.__set_schedule_variables(
  File "/gitlabform/gitlabform/gitlabform/processors/project/schedules_processor.py", line 107, in __set_schedule_variables
    for variable_key, variable_data in variables.items():
AttributeError: 'NoneType' object has no attribute 'items'
```

when I try to create schedule without any variables. For example the one from the docs:

```
      "Some schedule":
        ref: master
        cron: "0 * * * MON-FRI"
        cron_timezone: "London"
        active: false
```

Perhaps it should be checked if `variables` are None here: 
https://github.com/egnyte/gitlabform/blob/master/gitlabform/gitlabform/processors/project/schedules_processor.py#L107

gdubicki:
Hey @trzeciakp ! Thanks for reporting this.

Can you please provide a PR to fix this? It seems that you found the code to fix already.

gdubicki:
Actually it's strange that we did run into this bug as we do test a case with creating a schedule without variables - here is the config: https://github.com/egnyte/gitlabform/blob/main/gitlabform/gitlabform/test/test_schedules.py#L68-L80 ...

gdubicki:
Oups, we have been ignoring exceptions from test as long as the asserts were fine. :| Fixed in https://github.com/egnyte/gitlabform/commit/7a0e598b27ecc836270e28842c0d6364557a75fb . Let's see how many tests will fail now.. 😓 cc @kowpatryk

gdubicki:
The fix will be released in v1.20.1, which is being built right now.

PR #169, opened 2021-01-27
fix: #162 - Allow schedule variables to be empty

grzesuav:
Fixing https://github.com/egnyte/gitlabform/issues/162


gdubicki:
Thanks for the PR @grzesuav !

PR #171, opened 2021-01-29
Add example for typical protection of "main" (aka "master")

glensc:
I figured that a typical example of what you want for the default branch is missing.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -162,6 +162,12 @@ group_settings:
         merge_access_level: 30
         unprotect_access_level: 40
         code_owner_approval_required: true
+      branch_protected_from_changes:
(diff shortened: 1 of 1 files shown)


glensc:
I could not find what is `unprotect_access_level` but without that key, no changes were saved, so I kept it.

gdubicki:
> I could not find what is `unprotect_access_level` but without that key, no changes were saved, so I kept it.

I am not sure if I got your right, but `unprotect_access_level` is explained as one of the parameters [here](https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches).

Anyway: thanks for the contribution, @glensc !

Issue #182, opened 2021-02-19
Ability to set default branch?


llamasoft:
Maybe I missed it, but is there a way to have gitlabform set the default branch for a repository?  
We created new "main"/"develop" branches in many of our repositories, but the default branch still points to "master".  Being able to update the default branch from gitlabform would make it considerably easier to automate these changes.

Pigueiras:
This one is working for me:

```
project_settings:
  default_branch: production
```

llamasoft:
Odd, I'm not seeing that setting anywhere in the code or the example config.

llamasoft:
Ahh, disregard.  I see it falls under `project_settings` which has a link to the available values.  I'll open a quick PR to add `default_branch` to the example config as it's a setting some people might be looking for.

gdubicki:
I assume that this is fixed by #183. Please reopen if needed. :)

PR #183, opened 2021-02-22
Adds `default_branch` example to `project_config`

llamasoft:
The example config makes extensive use of "main" as the default project branch, but the setting isn't present in the example config.
It falls under the umbrella `project_config` (with a full list of settings available via GitLab documentation), but given that it's relevant to other options in the config file, I've added it here as well.

(This solves issue #182 that I opened in ignorance.)


gdubicki:
Thanks for the contribution @llamasoft ! Perhaps thanks to it someone will not get mislead in the future. 😊

PR #184, opened 2021-03-08
Disallow almost duplicate group and project names in the config

### LINKED ISSUE #160: Group Case Sensitivity Issue
When a group name has caps in its URL but lowercase is used for `gitlabform` the group is found but no projects within it are processed.

## Steps to reproduce

1. create a group with caps e.g. `gitlabformtest/FOO`
> Note: you may also need to go to the group settings and change it in the group settings (General Settings > Advanced > Change group URL)
2. create a project in the group e.g. `foo-project`
3. setup gitlabform config group_settings for the group e.g. `gitlabformtest/foo`
4. run gitlabform output:
```
$ gitlabform gitlabformtest/foo
*** # of groups to process: 1
*** # of projects to process: 0
> (1/1) Processing: gitlabformtest/foo
GitLabForm version: 1.18.3.2 (the latest)
```

## Cause

It looks like the API call to find groups works fine (case insensitive) so the group is returned and the subsequent call to get the projects returns the projects for the group.

However then when the projects are added to the list to be processed the `startswith` check will not match anything as it uses the lowercase group name and `path_with_namespace` from the API call returns the namespace with uppercase values.

See https://github.com/egnyte/gitlabform/blob/f765571ce826b3d745471ff7fbbb293589f3083f/gitlabform/gitlab/groups.py#L52

## Workaround

Pretty simple workaround is to run `gitlabform gitlabformtest/FOO` instead of `gitlabform gitlabformtest/foo` but the fact it found the group but no projects stumped me for a while.

gdubicki:
* Disallow almost duplicate group and project names in the config
* Ignore project and group paths case when looking for their configs (TODO)

This will fix #160

### DIFF
--- a/gitlabform/configuration/core.py
+++ b/gitlabform/configuration/core.py
@@ -56,6 +56,8 @@ class ConfigurationCore:
                     )
                     sys.exit(1)
 
+            self.find_almost_duplicates()
+
         except (FileNotFoundError, IOError):
             raise ConfigFileNotFoundException(config_path)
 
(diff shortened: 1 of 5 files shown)


gdubicki:
Moved to v2, release in v2.0.0 RC2 and later.

PR #188, opened 2021-03-25
Add sample config for CI CD variables scoped to environment

### LINKED ISSUE #186: Feature: Support filter parameter for secrets or CI/CD variables
Below is an example config I'm trying to use for managing some CI CD variables that are associated with specific "GitLab Environments". In this example, I am using the same name for the variable (i.e. `APP_HOST_AWS_ACCESS_KEY_ID`) but it is scoped to specific environments.

This allows me to have a simpler CI config (i.e. `.gitlab-ci.yml` file) because the appropriate variable will be exposed by GitLab depending on the "environment" that the CI is targeting.

Unfortunately, the following configuration is not working. When it is processed by gitlabform, there is only 1 variable added to the project, which is variable for "staging" environment. I believe that's because they are being processed alphabetically.

```yml
project_settings:
  group-name/project1:
    secret_variables:
      aws_access_key_id_for_deploying_in_production:
        key: APP_HOST_AWS_ACCESS_KEY_ID
        value: "prod-value-1234"
        protected: true
        masked: true
        environment_scope: "production"

      aws_access_key_id_for_deploying_in_staging:
        key: APP_HOST_AWS_ACCESS_KEY_ID
        value: "staging-value-1234"
        protected: true
        masked: true
        environment_scope: "staging"
```

I believe gitlabform is using the following endpoint for creating/updating variables.

https://docs.gitlab.com/ee/api/project_level_variables.html#update-variable

In my case, I think the update needs to happen by filtering `environment_scope`. That way appropriate variable will be updated. I don't think gitlabform currently has logic for this.

https://docs.gitlab.com/ee/api/project_level_variables.html#the-filter-parameter

Would it be possible for adding support for the filter parameter? Or, is there something I missed in my config so that I can set variable with same name but scoped to different environments?

amimas:
This change adds a sample configuration that shows how to add
multiple CI CD variables with the same name but scoped to different
deployment environments.

closes #186

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -329,6 +329,28 @@ project_settings:
         title: ssh_key_name_that_is_shown_in_gitlab_3
         can_push: false
 
+    secret_variables:
+      # CI CD variable for this project. Variables can be scoped to an "environment".
+      # Scoping to environment also allows using the same "key" name for multiple variables.
+      # This can simplify CI Config file because the appropriate variable will be exposed
+      # by GitLab to the job that is targetting a given environment. If the same "key" name
+      # is used for different environments, you'll need to add filter[environment_scope] to
+      # configuration as shown below.
+      aws_access_key_id_for_deploying_in_production:
+        key: APP_HOST_AWS_ACCESS_KEY_ID
+        value: "prod-value-1234"
+        protected: true
+        masked: true
+        environment_scope: production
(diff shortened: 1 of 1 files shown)


gdubicki:
Thanks @amimas !

Issue #164, opened 2021-01-20
Support for managing Project Environment


amimas:
Hi there,

Does gitlabform has support for configuring/managing Project Environments? I skimmed through the code but didn't see any support for it. Thought I'd check in case I missed something. Below is the API docs from GitLab on Project Environments.

https://docs.gitlab.com/ee/api/environments.html

gdubicki:
Hey @amimas !

No, we don't support this API yet. We are open to Pull Requests adding support for it. :)

ss7548:
Please see pull request #190

amimas:
PR #423 implemented ability to manage protected environment. Not sure if this should be considered as closed.

I guess regular environment config isn't available but maybe that's not a big deal or priority because those are created automatically through project's CI CD config (`.gitlab-ci.yml`). Only thing useful I can see for managing regular environments is to be able to delete envs that are not needed anymore; i.e. env created temporarily called review apps.

gdubicki:
https://github.com/gitlabform/gitlabform/pull/423 has been released in [v3.3.0](https://github.com/gitlabform/gitlabform/releases/tag/v3.3.0) today.

kaihendry:
Be good if the [Feature matrix/comparison sheet](https://docs.google.com/spreadsheets/d/1RenC5OoLW_bt8QYrijNP42w8SGBw8JTg1-7RKWgStrQ/edit#gid=0) via https://gitlabform.github.io/gitlabform/similar_apps/ could be updated with support for Protected Environments. 😬

amimas:
> I guess regular environment config isn't available but maybe that's not a big deal or priority because those are created automatically through project's CI CD config (.gitlab-ci.yml). Only thing useful I can see for managing regular environments is to be able to delete envs that are not needed anymore; i.e. env created temporarily called review apps.

Having gitlabform configure protected environments should be sufficient. Gitlab environments can be created by the CI config in a project when the pipeline runs. So there's not much to do for gitlabform. As for deleting those non protected environments as mentioned above, it might not be easy to decide because gitlabform can't know when an environment is in-use or not. Besides, gitlab CI already provides ability to stop environments from the CI pipeline.

Closing this issue for now. If there's other use cases, open a new issue with details.

PR #190, opened 2021-03-27
Implements project environments #164

ss7548:


gdubicki:
Hey @ss7548 , that’s a great looking PR, thanks!

I may have one or two nit-picking comments but I think this will land in main within a few days and gets release soon after. :)

Before I go into the code deeper, can you please just check why the tests failed?

ss7548:
@gdubicki , I guess calling a a super function overwrote or otherwise corrupted memory impacting one of the merge request tests.

gitlabform/gitlab/environments.py
```diff
@@ -0,0 +1,56 @@
...
+        # https://docs.gitlab.com/ee/api/environments.html#stop-an-environment
+        self._make_requests_to_api(
+            "projects/%s/environments/%s/stop",
+            (project_and_group_name, eid),
+            method="POST",
+            expected_codes=[200, 404],
+        )
+
+
+#    def put_environment():
```
gdubicki, line 55: Is there a specific reason this is not implemented? According to the API docs at https://docs.gitlab.com/ee/api/environments.html#edit-an-existing-environment you can edit an existing environment.

gitlabform/gitlabform/processors/project/environments_processor.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self, gitlab: GitLab):
+        super().__init__("environments")
+        self.gitlab = gitlab
+
+    def _process_configuration(self, project_and_group: str, configuration: dict):
+        logging.debug(
+            "Environments BEFORE: %s",
+            self.gitlab.get_all_environments(project_and_group),
+        )
+        e = self.gitlab.get_all_environments(project_and_group)
```
gdubicki, line 19: Nit: single-letter variables are not great except maybe iterators in loops as they are very uninformative and can be misleading. When I was reading the code below guessed that `e` is a single environment and I was wrong...

gdubicki: Also when debug output will be enabled then this line will do 2nd call to get all the environments to GitLab in a row. Check out https://github.com/egnyte/gitlabform/blob/v2/gitlabform/gitlabform/processors/project/project_push_rules_processor.py#L17-L19 where we do it only once.

gitlabform/gitlabform/processors/project/environments_processor.py
```diff
@@ -0,0 +1,64 @@
...
+        for environment in sorted(configuration["environments"]):
+            logging.info("Setting environment: %s", environment)
+            data = configuration["environments"][environment]
+            env = data["name"]
+
+            if env in map(operator.itemgetter("name"), e):
```
gdubicki, line 25: I like that you are getting all the envs and then operate only the ones that you should - that minimizes the number of calls to GitLab. 🙂 It also opens room to easily add "enforce" option - that would keep the envs exactly as defined in the code, so deleting all the ones that may have been defined manually - in the future. 👍

gdubicki (changes requested):
Thanks for fixing the tests! Please check out my review.

ss7548:
Observation on testing.  I test against my dev omnibus install, GitLab V 13.9.2-ee
After running tests, several GROUPS and PROJECTS are left in the wake.
When running the tests for the first time, the test TestMergeRequestApprovers::test__if_it_works__single_user fails
When running the tests for the second time, the aforementioned test passes, but the test TestArchiveProject::test__dont_edit_archived_project fails.

After cleaning up (i.e. removing the GROUPS left over from the tests), the same pattern occurs.

There is a problem with the MergeReqestsApprovers feature which I have seen in my production environment.

When a new project is created, there is by default an approval rule "Any eligible user" that requires 0 approvals.

When you add an approval rule with GitLabForm, it changes the 0 approvals of  "Any eligible user" to the number specified in the approval config.

This is also why the "test__if_it_works__single_user" fails the first time, After setting the approval rule, it expects that there is only one, ignoring the "Any eligible user" rule.   After the test it deletes ALL rules, so the second time you run it, the test passes because the "Any eligible user" rule has been deleted.  

The reason for this is that the PROJECT that this test creates, is left in tact when the tests complete.

I have no explanation for why the TestArchiveProject::test__dont_edit_archived_project test fails the second time through.

ss7548:
@gdubicki  any feedback?

gdubicki:
Hey @ss7548 , sorry for lack of response. I have been looking into this but haven't replied yet because this need more time from me and I wasn't able to find it recently. I will try to do it over this weekend.

ss7548:
not sure if you are waiting for something from me?

gdubicki:
Sorry, I didn't have the time and will to do this before, but I have found it today. I am now trying to overhaul the tests in GitLabForm completely - that should fix all the issues you have reported, @ss7548 . Stay tuned, I will share an update today.

gdubicki:
Unfortunately I got stuck because of #196 . :(

gdubicki:
Thanks for your patience @ss7548 !

I think that I got #196 fixed and have done the tests revamp for:
```
test_archive_project.py
test_branches.py
test_deploy_keys.py
test_files.py
```
...in the v2 branch. A final version 2 will be released within days, so it would be best if you would rebase your branch against `v2` and target `v2` with your PR. Can you do that please?

gdubicki:
Closing in favor of #203 .

Issue #172, opened 2021-01-29
Protected tags doesn't accept number 0


glensc:
```yml
    tags:
      "v*":
        protected: true
        create_access_level: 0
```

this is treated as if `create_access_level` wasn't present at all:

![image](https://user-images.githubusercontent.com/199095/106261244-fd6ea480-6229-11eb-8437-f491265dd8f7.png)

A workaround is to quote the value:

```yml
    tags:
      "v*":
        protected: true
        create_access_level: "0"
```

gdubicki:
Thanks for reporting this!

Could you add a PR with a fix, perhaps? A fresh example of code adding tests is in #170. It would be best if you would do it TDD-style - add the tests first, watch them fail on the case you noticed here, add a fix, watch the tests pass. :)

glensc:
Maybe, if you tell how to check it the proper way. Or provide an example of similar cases (I'm not a frequent python or gitlabform user)

gdubicki:
I was hoping that #170 provides a good example.. 😅 If it doesn’t then let me know what needs explaining, I’ll improve the docs/comments/code. :)

glensc:
@gdubicki I was more about how to perform this check properly:
- https://github.com/egnyte/gitlabform/pull/39#discussion_r566725404

the code must be able to make distinctions between: not set and 0, "", null

gdubicki:
Oups, sorry for the misunderstanding.
Maybe:
```
if create_access_level is not None:
```
...will do.

mcjffld:
just opened a PR for this - https://github.com/egnyte/gitlabform/pull/191 -

gdubicki:
Fixed in v2.2.0, released a moment ago. Please reopen if needed!

PR #191, opened 2021-04-08
Update to allow a '0' access for Protected tags - issue #172

mcjffld:
quick change, tested locally, now getting

{"create_access_level": 0, "name": "*"}

### DIFF
--- a/gitlabform/gitlab/tags.py
+++ b/gitlabform/gitlab/tags.py
@@ -23,7 +23,7 @@ class GitLabTags(GitLabCore):
 
     def protect_tag(self, project_and_group_name, tag_name, create_access_level):
         data = {"name": tag_name}
-        if create_access_level:
+        if create_access_level != None:
             data["create_access_level"] = create_access_level
         return self._make_requests_to_api(
             "projects/%s/protected_tags",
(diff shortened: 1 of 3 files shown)


gdubicki:
Thanks @mcjffld , but can you please add a test for this?

mcjffld:
> Thanks @mcjffld , but can you please add a test for this?

I'd love to - I did and some code that runs a test and passes, but I wasn't able to figure out how to actually test that the tag change was working - the get_tags call returned an empty list all the time

any thoughts?

gdubicki:
Sorry for keeping you waiting, @mcjffld .

You have to use `create_tags` to create some tags before protecting them.

gdubicki:
As we are close to release of v2, can you please rebase your MR to it and target that branch, @mcjffld ?

gdubicki:
Hey @mcjffld ! Do you have the time and the will do continue working on this? Because I just added the new style tests for protecting tags into the `main` branch, so it would be much easier for you to just add a new test there instead of adding a whole new set of tests in the `gitlabform.gitlab` package.

gdubicki:
I have fixed it in https://github.com/egnyte/gitlabform/pull/245/commits/2d2e36922ac15c283ca1eb1d770174a63c13ca3c as I don't want bugs in the app to linger for too long.

Thanks for your effort @mcjffld , don't hesitate to take on another bug or feature in future. 😊

PR #195, opened 2021-04-27
Make writing tests more user-friendly and the tests more robust

gdubicki:
Firstly the tests that the developer of a new/fixed/extended feature should be easier to write.

Secondly the tests should be more robust - not breaking on 2+ run (#190), not breaking if you break the test at any time.


gdubicki:
Got stuck with local testing of this code with this issue: #196

gdubicki:
Merged into v2.

PR #198, opened 2021-04-30
Do deep merges of configuration

### LINKED ISSUE #197: merge_requests/approvals is replaced instead of merged
Hi, it seems that nested configurations defined in both `common_settings` and `group_settings` are replaced, instead of merged.

For example, when defining a configuration like this:
```
common_settings:
  merge_requests:
    approvals:
      approvals_before_merge: 3

group_settings:
  "group_name":
    merge_requests:
      approvals:
        merge_requests_author_approval: true
```
and running 
```
gitlabform -v --noop -c config.yaml group_name/project_name
```
The result is
```
GitLabForm version: 1.21.0 (the latest)
Reading config from file: config.yaml
Connected to GitLab version: 13.10.3-ee (db2e358dba4)
Reading config from file: config.yaml
*** # of groups to process: 0
*** # of projects to process: 1
* [1/1] Processing: group_name/project_name
Processing merge_requests in dry-run mode.
Project group_name/project_name approvals changes:
merge_requests_author_approval: "???" => true
```
i.e. only `merge_requests_author_approval` is applied, while I would expect that both the configurations are applied:
- `approvals_before_merge` from `common_settings`
- `merge_requests_author_approval` from `group_settings`

What is the expected behaviour?
Should the `approvals` sections be merged together as I was expecting?

Pigueiras:
Fixes #197

### DIFF
--- a/gitlabform/configuration/projects_and_groups.py
+++ b/gitlabform/configuration/projects_and_groups.py
@@ -107,22 +107,18 @@ class ConfigurationProjectsAndGroups(ConfigurationCore):
         :return: merge more general config with more specific configs.
                  More specific config values take precedence over more general ones.
         """
-        merged_config = {}
+        if isinstance(more_general_config, dict):
+            for key in more_specific_config:
+                if key in more_general_config:
+                    more_general_config[key] = self.merge_configs(
+                        more_general_config[key], more_specific_config[key]
+                    )
+                else:
+                    more_general_config[key] = more_specific_config[key]
+        else:
+            more_general_config = more_specific_config
 
-        for key in more_specific_config.keys() | more_general_config.keys():
-
-            if key in more_general_config and key not in more_specific_config:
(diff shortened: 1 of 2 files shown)


gdubicki:
Thanks for the PR, @Pigueiras ! But you edited the wrong file, check the path: `build/lib/gitlabform/configuration/projects_and_groups.py`.

Pigueiras:
Oops, I fix in a sec

Pigueiras:
(It should look better now) I need to take a look to the unit tests now, but do you agree with such a pull request @gdubicki ? I think it was the intended behavior from the beginning no?

Pigueiras:
The way it's implemented in this PR it only merges dicts. I am not merging lists like you can see in the unit test I added, because otherwise if you want one approver of the common/groups to **not** appear in the project settings, you can't do that. Let me know if you're ok with it or not.

PR #203, opened 2021-05-04
 Implements project environments #164 

ss7548:
@gdubicki  Here is the pull request against V2


gdubicki:
Thanks @ss7548 ! Now please enable running your test by adding a line like https://github.com/egnyte/gitlabform/blob/v2/.github/workflows/acceptance.yml#L33 and make the tests pass.

I recommend you updating your test to the newer style, like for example https://github.com/egnyte/gitlabform/blob/v2/gitlabform/gitlabform/test/test_archive_project.py (I will soon update all remaining tests to the same style).

ep-linden:
Is there any status on whether this feature will be added to GitLabForm? The way it is configured could be more intuitive based on what my colleague @amimas suggested.

gdubicki:
Looking at how much time has passed since @ss7548 last reply here, I am afraid that he won't finish it. But the code here looks almost ready, needs only update of the tests and perhaps some configuration syntax changes, as suggested by @amimas. So if you need this feature, then I suggest you to reuse this PR, complete the implementation and file a new PR, @ep-linden. And I will review it, merge it when it's ready and release it as soon as possible.

ep-linden:
> Looking at how much time has passed since @ss7548 last reply here, I am afraid that he won't finish it. But the code here looks almost ready, needs only update of the tests and perhaps some configuration syntax changes, as suggested by @amimas. So if you need this feature, then I suggest you to reuse this PR, complete the implementation and file a new PR, @ep-linden. And I will review it, merge it when it's ready and release it as soon as possible.

Sounds good, we can definitely take over.

gdubicki:
Is it still valid that your team can take this over, @amimas? If not, then I propose to close this PR as outdated.

amimas:
> Is it still valid that your team can take this over, @amimas? If not, then I propose to close this PR as outdated.

Unfortunately I don't think we'll be able to look into this going forward @gdubicki . Also, now that gitlabform has support for "protected environments", this is probably less of a priority. Let's close this one.

PR #204, opened 2021-05-06
build & push release candidate Docker images

### LINKED ISSUE #201: Release pre-releases as a Docker image too
It just hit me that if we only don't add a tag "latest" to a pre-release Docker image, then it will be effectively like a PyPI pre-release package - people using version "latest" will keep using the latest final while people interested in pre-releases will install them by pointing to their exact versions.

weakcamel:
Closes #201

* refactor dev/build_docker_images.sh
* add a Python script to provide release type based on version string
* build & push Docker images for every release; "latest" only for public ones
* fix a few yamllint issues


weakcamel:
Test for a regular release: 
* https://github.com/weakcamel/gitlabform/actions/runs/818479154

Test for an RC release:
* https://github.com/weakcamel/gitlabform/actions/runs/818477303

dev/build_docker_images.sh
```diff
@@ -4,35 +4,49 @@
...
 #   docker_password
+#   GITHUB_REF
 
 python_version="3.8"
-tags=""
 version="$(cat version)"
 
+set -ue
+
+git_tag="${GITHUB_REF#refs/tags/v*}"
```
weakcamel, line 14: Not sure if that's something desirable to have in here; what do you think @gdubicki

dev/build_docker_images.sh
```diff
@@ -4,35 +4,49 @@
...
-  --tag "${latest}" .
-tags="$tags ${latest}"
+  --tag "${image_name}" .
+
+tags=( "${image_name}" )
 
-docker tag "${latest}" "egnyte/gitlabform:${version}-${suffix}"
-tags="$tags egnyte/gitlabform:${version}-${suffix}"
+# TODO: something more clever here maybe?
+# TODO: do we need to validate the version?
```
gdubicki, line 35: I was thinking about keeping the GitHub Actions job "Skip if this is not a final release" but modify it to pass info if this a final release or not to the bash script. This way we will use the existing Python logic for deciding if this is a final release or not and not have to reimplement it in bash.

gdubicki: That would be perfect. :)

dev/build_docker_images.sh
```diff
@@ -4,35 +4,49 @@
...
+    latest=egnyte/gitlabform:latest
+    docker tag "${image_name}" "${latest}"
+    tags+=( "${latest}" )
+  fi
 fi
 
-for image in ${tags}
-do
+echo "Images to be published: ${tags[*]}"
+for image in "${tags[@]}"; do
```
gdubicki, line 49: [Nit:](https://google.github.io/eng-practices/review/reviewer/standard.html) in these lines you interchange names "tag" and "image" while technically they are different thing. I think that this script would benefit from pulling out only the image name into a separate variable and operate everywhere else on tag and tags. Then only in `docker tag` and `docker push` we would use both.

dev/build_docker_images.sh
gdubicki: I think that we should be consistent from where do we get the version for both Docker images and the PyPI releases.

dev/build_docker_images.sh
weakcamel: Scipt is there. I find the logic of [packaging module](https://packaging.pypa.io/en/latest/version.html
) (or maybe PEP-440) quite baffling though :-)

Namely, I was in no way able to make it raise an exception:

```
$ ./dev/check_release_type.py -V 1.2.3
public
$ ./dev/check_release_type.py -V 1.2.3rc1
pre
$ ./dev/check_release_type.py -V 1.2.3dev
dev
$ ./dev/check_release_type.py -V 1.2.3post1
post
$ ./dev/check_release_type.py -V 1.2.3post1-123.123.123ABC-..
public
$ ./dev/check_release_type.py -V thisisratherweirdisntit
public
$ ./dev/check_release_type.py -V thisisratherweirdisntit?
public
$

```

dev/build_docker_images.sh
weakcamel: agreed, fixed - can you have another look and see if that now makes more sense?

gdubicki: I think that it's much better now, thanks! :)

.github/workflows/publish-to-pypi.yml
```diff
@@ -1,9 +1,10 @@
+---
 name: Publish 🐍 to PyPI
 
 on:
   push:
     tags:
-      - "v*"
+      - "NOv*"  ## TODO: Remove; temporarily changed to allow docker tests
```
weakcamel, line 7: reminder to remove after all testing is done

dev/build_docker_images.sh
gdubicki: True, it is strange.. If I am reading https://www.python.org/dev/peps/pep-0440/#version-scheme correctly then a valid PEP 440 version has to contain at least one digit... And `packaging` declares that `parse` can throw an `packaging.version.InvalidVersion` exception...

But anyway, I would rather check only if it `is_prerelease` or anything else. As in the latter category are both public as well as post release and we DO sometimes release posts and they should be treated as the latest.

weakcamel: Changed per suggestion.

weakcamel:
Issues resolved, re-tested the "public" release here: https://github.com/weakcamel/gitlabform/runs/2536128155?check_suite_focus=true

PR #231, opened 2021-06-08
Remove call to delete_legacy_approvers in gitlab versions >= 13.11

### LINKED ISSUE #220: Removed approvers endpoint used by GitLabForm
**Describe the bug**

In GitLab [13.11.0](https://gitlab.com/gitlab-org/gitlab/-/blob/master/CHANGELOG-EE.md#13110-2021-04-22) a deprecated project approvers update REST API has been removed. This endpoint is still invoked by GitLabForm to delete legacy approvers (see [merge_request_processor.py](https://github.com/egnyte/gitlabform/blob/main/gitlabform/gitlabform/processors/project/merge_requests_processor.py#L42)) if the response from GitLab contains `approvers` or `approver_groups`. Unfortunately although the endpoint has been removed above attributes are still returned by GitLab making GitLabForm call non-existent endpoint.

```
Traceback (most recent call last):
  File "/gitlabform/gitlabform/gitlabform/core.py", line 315, in process_all
    self.project_processors.process_project(
  File "/gitlabform/gitlabform/gitlabform/processors/project/__init__.py", line 57, in process_project
    processor.process(project_and_group, configuration, dry_run)
  File "/gitlabform/gitlabform/gitlabform/processors/util/decorators.py", line 42, in method_wrapper
    return method(self, project_and_group, SafeDict(configuration), dry_run)
  File "/gitlabform/gitlabform/gitlabform/processors/abstract_processor.py", line 34, in process
    self._process_configuration(project_or_project_and_group, configuration)
  File "/gitlabform/gitlabform/gitlabform/processors/project/merge_requests_processor.py", line 37, in _process_configuration
    self.gitlab.delete_legacy_approvers(project_and_group)
  File "/gitlabform/gitlabform/gitlab/projects.py", line 235, in delete_legacy_approvers
    self._make_requests_to_api(
  File "/gitlabform/gitlabform/gitlab/core.py", line 137, in _make_requests_to_api
    response = self._make_request_to_api(
  File "/gitlabform/gitlabform/gitlab/core.py", line 220, in _make_request_to_api
    raise NotFoundException("Resource path='%s' not found!" % url)
gitlabform.gitlab.core.NotFoundException: Resource path='****edited****/api/v4/projects/****edited****/approvers' not found!
```

Here's the response from `projects/:id/approvals`
```
{
  "approvers": [],
  "approver_groups": [],
  "approvals_before_merge": 1,
  "reset_approvals_on_push": true,
  "disable_overriding_approvers_per_merge_request": true,
  "merge_requests_author_approval": false,
  "merge_requests_disable_committers_approval": true,
  "require_password_to_approve": false
}
```

We're using an older version of GitLabForm but the same logic triggering the issue is still on the main branch.

**GitLabForm version**

GitLabForm version: 1.18.3.2 (the latest is 2.0.0.post1 - please update!)

Pigueiras:
Fixes #220

### DIFF
--- a/gitlabform/gitlabform/processors/project/merge_requests_processor.py
+++ b/gitlabform/gitlabform/processors/project/merge_requests_processor.py
@@ -1,6 +1,7 @@
 import logging
 
 import cli_ui
+from distutils.version import LooseVersion
 
 from gitlabform.gitlab import GitLab
 from gitlabform.gitlabform.processors.abstract_processor import AbstractProcessor
@@ -38,8 +39,10 @@ class MergeRequestsProcessor(AbstractProcessor):
                 "approvers" in approvals_settings
                 or "approver_groups" in approvals_settings
             ):
-                logging.debug("Deleting legacy approvers setup")
-                self.gitlab.delete_legacy_approvers(project_and_group)
+                # /approvers endpoint has been removed in 13.11.x GitLab version
+                if LooseVersion(self.gitlab.version) < LooseVersion("13.11"):
+                    logging.debug("Deleting legacy approvers setup")
+                    self.gitlab.delete_legacy_approvers(project_and_group)
 
             approval_rule_name = "Approvers (configured using GitLabForm)"
 
--- a/gitlabform/gitlab/core.py
+++ b/gitlabform/gitlab/core.py
@@ -41,6 +41,7 @@ class GitLabCore:
             cli_ui.debug(
                 f"Connected to GitLab version: {version['version']} ({version['revision']})"
             )
+            self.version = version["version"]
         except Exception as e:
             raise TestRequestFailedException(e)
 


Pigueiras:
Probably we should add some unit tests for the version comparison check, and move the `LooseVersion > LooseVersion` logic to a separate method, but not sure where to put it (to `gitlabform/processors/util/...`?, to a method in GitLabCore?

gitlabform/gitlabform/processors/project/merge_requests_processor.py
```diff
@@ -1,6 +1,7 @@
 import logging
 
 import cli_ui
+from distutils.version import LooseVersion
```
gdubicki, line 4: Have you considered using `packaging.version.parse()` instead? It seems to be the preferred tool for processing version in Python currently, see https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python .

We also already use it it https://github.com/egnyte/gitlabform/blob/main/dev/check_release_type.py#L3 .

gdubicki:
Thanks for your contribution @Pigueiras ! :)

> Probably we should add some unit tests for the version comparison check, and move the `LooseVersion > LooseVersion` logic to a separate method, but not sure where to put it (to `gitlabform/processors/util/...`?, to a method in `GitLabCore`?

I think that until we get another version comparison in the code it would be overengineering to do anymore more than you did.

gitlabform/gitlabform/processors/project/merge_requests_processor.py
Pigueiras: That one gives me strange results due to the `-ee` suffix in the GitLab. Maybe sanitize it first with something like 
```
self.version = re.match('\d+\.\d+\.\d+', version['version']).group(0)
```

and then use `packaging`?

```
>>> version.parse('13.11.3-ee') < version.parse('13.11')
True  # weird
```

gdubicki: I checked that and '13.11.3-ee' is parsed as `LegacyVersion` (because it's not PEP440-compliant) and `13.11` as `Version` and as [noted here](https://packaging.pypa.io/en/latest/version.html#packaging.version.LegacyVersion): 
> LegacyVersion instances are always ordered lower than Version instances.

I would not code it like that, it's a trap for the users IMHO.

In this case I would leave your code as it is then. :)

Pigueiras:
Thanks @gdubicki !

PR #232, opened 2021-06-08
Add possibility to add extra parameters in protect branches method

### LINKED ISSUE #227: Add option to allow push force in protected branches
There is no way to specify `allow_force_push` when setting up the protected branches

https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches

Pigueiras:
Fixes #227

### DIFF
--- a/gitlabform/gitlab/branches.py
+++ b/gitlabform/gitlab/branches.py
@@ -22,62 +22,23 @@ class GitLabBranches(GitLabCore):
         )
 
     # new API
-    def branch_access_level(
-        self,
-        project_and_group_name,
-        branch,
-        push_access_level,
-        merge_access_level,
-        unprotect_access_level,
-    ):
+    def branch_access_level(self, project_and_group_name, branch, protect_settings):
 
         url = "projects/%s/protected_branches?name=%s"
         parameters_list = [
             project_and_group_name,
             branch,
         ]
-        if push_access_level is not None:
-            url += "&push_access_level=%s"
-            parameters_list.append(push_access_level)
-        if merge_access_level is not None:
-            url += "&merge_access_level=%s"
-            parameters_list.append(merge_access_level)
-        if unprotect_access_level is not None:
-            url += "&unprotect_access_level=%s"
-            parameters_list.append(unprotect_access_level)
-
         return self._make_requests_to_api(
             url,
             tuple(parameters_list),
             method="POST",
-            data={},
+            data=protect_settings,
             expected_codes=[
                 200,
                 201,
                 409,
             ],  # TODO: check why is 409 Conflict accepted here :/
-        )
-
-    def branch_code_owner_approval_required(
-        self,
-        project_and_group_name,
-        branch,
-        code_owner_approval_required,
-    ):
-        data = {
-            "id": project_and_group_name,
-            "branch": branch,
-            "code_owner_approval_required": code_owner_approval_required,
-        }
-        return self._make_requests_to_api(
-            "projects/%s/protected_branches/%s",
-            (
-                project_and_group_name,
-                branch,
-            ),
-            method="PATCH",
-            data=data,
-            expected_codes=[200, 201],
         )
 
     def unprotect_branch(self, project_and_group_name, branch):
@@ -117,35 +78,6 @@ class GitLabBranches(GitLabCore):
         return self._make_requests_to_api(
             "projects/%s/protected_branches/%s", (project_and_group_name, branch)
         )
-
-    def get_only_branch_access_levels(self, project_and_group_name, branch):
-        try:
-            result = self._make_requests_to_api(
-                "projects/%s/protected_branches/%s", (project_and_group_name, branch)
-            )
-            push_access_levels = None
-            merge_access_level = None
-            unprotect_access_level = None
-            if (
-                "push_access_levels" in result
-                and len(result["push_access_levels"]) == 1
-            ):
-                push_access_levels = result["push_access_levels"][0]["access_level"]
-            if (
-                "merge_access_levels" in result
-                and len(result["merge_access_levels"]) == 1
-            ):
-                merge_access_level = result["merge_access_levels"][0]["access_level"]
-            if (
-                "unprotect_access_levels" in result
-                and len(result["unprotect_access_levels"]) == 1
-            ):
-                unprotect_access_level = result["unprotect_access_levels"][0][
-                    "access_level"
-                ]
-            return push_access_levels, merge_access_level, unprotect_access_level
-        except NotFoundException:
-            return None, None, None
 
     def create_branch(
         self, project_and_group_name, new_branch_name, create_branch_from_ref
--- a/gitlabform/gitlabform/processors/util/branch_protector.py
+++ b/gitlabform/gitlabform/processors/util/branch_protector.py
@@ -94,30 +94,39 @@ class BranchProtector(object):
         # unprotect first to reset 'allowed to merge' and 'allowed to push' fields
         self.gitlab.unprotect_branch_new_api(project_and_group, branch)
 
+        protect_rules = {
+            key: value
+            for key, value in requested_configuration.items()
+            if key != "protected"
+        }
+
         self.gitlab.branch_access_level(
             project_and_group,
             branch,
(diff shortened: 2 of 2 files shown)


Pigueiras:
It's still on going, I wanted to add more logic for the `allowed_to_merge`, `allowed_to_push`... because those are tricky https://docs.gitlab.com/ee/api/protected_branches.html#example-with-user--group-level-access

gdubicki:
Thanks for working on this!

I would deal with `allowed_to_push` and `allowed_to_merge` in a following way:

Firstly lets just expect GitLabForm users to provide in the config exactly the data that we can pass in the API call.

So [this example from GitLab API docs](https://docs.gitlab.com/ee/api/protected_branches.html#example-with-allow-to-push-and-allow-to-merge-access) would translate to such config syntax:

```
branches:
      main:
        protected: true
        allowed_to_push:
          - access_level: 30
        allowed_to_merge:
          - access_level: 30
          - access_level: 40
```

We would also still allow using fields like we do now, such as in [this example](https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches):

```
branches:
      "*-stable":
        protected: true
        push_access_level: 30
        merge_access_level: 30
        unprotect_access_level: 40
```

...BUT we would convert it internally to the above syntax, so treat it like it would be:

```
branches:
      "*-stable":
        protected: true
        allowed_to_push:
          - access_level: 30
        allowed_to_merge:
          - access_level: 30
        allowed_to_unprotect:
          - access_level: 40
```
(Perhaps this should be done somewhere in `Configuration`? 🤔 Not sure about that yet.)

Then for comparing the current protection level with the one in configuration in `BranchProtector` I would covert what we get from GitLab API, f.e.

```
    "push_access_levels": [
        {
            "access_level": 30,
            "access_level_description": "Developers + Maintainers",
            "user_id": null,
            "group_id": null
        }
    ],
    "merge_access_levels": [
        {
            "access_level": 30,
            "access_level_description": "Developers + Maintainers",
            "user_id": null,
            "group_id": null
        },
        {
            "access_level": 40,
            "access_level_description": "Maintainers",
            "user_id": null,
            "group_id": null
        }
    ],
    "unprotect_access_levels": [
        {
            "access_level": 40,
            "access_level_description": "Maintainers",
            "user_id": null,
            "group_id": null
        }
    ],
```
...like this:

* change key names like "push_access_levels" -> "allowed_to_push"
* remove "access_level_description" and fields with null values,

...and then we can do some kind of a deep compare (https://pypi.org/project/deepdiff/ maybe?) between that and what we have in the configuration.

To make it more user friendly we should also allow providing usernames and group names instead of ids, but we can add it in a separate PR.

WDYT, @Pigueiras ?

gdubicki:
Hmmm... I don't get why this API can get or return two consecutive levels like for a given type of action, for example:
```
"merge_access_levels": [
        {
            "access_level": 30,
            "access_level_description": "Developers + Maintainers",
            "user_id": null,
            "group_id": null
        },
        {
            "access_level": 40,
            "access_level_description": "Maintainers",
            "user_id": null,
            "group_id": null
        }
    ],
```
...if even the descriptions here clearly state that the former is a superset of the latter. 😐

Pigueiras:
Maybe a bug? In the GUI you can also do that (I don't think it makes a lot of sense either) 

image

The idea sounds good to me, maybe it would make more sense to do that transformation in `Configuration` but personally I don't have a strong opinion about it.

amimas:
Would the above proposal be able to handle  specifying users  or  groups under `allowed_to_merge` or `allowed_to_push`? We can do that  in the UI and it allows granular control over branch protection. For example: instead of setting `allowed_to_merge` to "Developers" role, we can select "group1" and that will make sure only members of that group can merge.

gdubicki:
> Would the above proposal be able to handle specifying users or groups under `allowed_to_merge` or `allowed_to_push`? We can do that in the UI and it allows granular control over branch protection. For example: instead of setting `allowed_to_merge` to "Developers" role, we can select "group1" and that will make sure only members of that group can merge.

For that we could add one more step to the conversion of the configuration: usernames and group names to user and group ids.

For example config like:
```
   branches:
      "main":
        protected: true
        allowed_to_push:
          - user: someuser
          - group: group
```
...would be converted to:
```
   branches:
      "main":
        protected: true
        allowed_to_push:
          - user_id: 123 # id of user "someuser"
          - group_id: 345 # id of group "agroup"
```
I would do all these conversions together in a new, intermediate step between reading the config and starting to process it.

illyaMs:
One more option might be to pass the "pure" `allowed_to_push` fields as they are described in gitlab api docs.

I mean instead of `would be converted to` just place that very same structure in the original config passed to the `gitlabform`:
```
branches:
      "main":
        protected: true
        allowed_to_push:
          - user_id: 123 # id of user "someuser"
          - group_id: 345 # id of group "agroup"
```

I actually tried it few times in our config before getting to this PR and related issue 😁

amimas:
Personally, I like what @gdubicki proposed because that way I don't have to know/use the "id" in the config. I can just refer to the group name or user name. Not sure if your approach supports this @illyaMs . If it does, that's great. If it doesn't, then it's inconsistent with other parts of the config used by gitlabform (i.e. members, project name, etc.).

Also, just for the sake of completeness, both `allowed_to_merge`  and `allowed_to_push` in Gitlab's API takes an array since you can set multiple values. What will that config look like in Gitlabform? Would they be comma separated values? Insstead of that, should the config be further nested so that it's more like native yaml syntax?

Not sure if the above is somehow breaking the existing "Raw parameter passing" feature.

gdubicki:
> One more option might be to pass the "pure" `allowed_to_push` fields as they are described in gitlab api docs.
> 
> I mean instead of `would be converted to` just place that very same structure in the original config passed to the `gitlabform`:
Sure, we can have both. And maybe even without extra flag as there are different keys for ids and names. So if the config contains `- user: foobar` it converts it to `- user_id: 123` but if it already is `-user_id: 123` then it leaves it as it is.

gdubicki:
> Also, just for the sake of completeness, both `allowed_to_merge` and `allowed_to_push` in Gitlab's API takes an array since you can set multiple values. What will that config look like in Gitlabform? Would they be comma separated values? Insstead of that, should the config be further nested so that it's more like native yaml syntax?
> 
> Not sure if the above is somehow breaking the existing "Raw parameter passing" feature.

I have taken it into account, look closely at the example in my first comment here:
```
        allowed_to_merge:
          - access_level: 30
          - access_level: 40
```

amimas:
> > Also, just for the sake of completeness, both `allowed_to_merge` and `allowed_to_push` in Gitlab's API takes an array since you can set multiple values. What will that config look like in Gitlabform? Would they be comma separated values? Insstead of that, should the config be further nested so that it's more like native yaml syntax?
> 
> I would use exactly the same syntax as GitLab API and that's why I proposed this to be an array in my first comment, here:
> 
> ```
>         allowed_to_merge:
>           - access_level: 30
>           - access_level: 40
> ```

Thanks for clarifying that. I wasn't sure if having multiple keys with the same name in gitlabform config will result in only the last config being applied. Sounds like gitlabform already handles these properly.

barryib:
We use gitlabform a lot, this is a needed feature for us. We have a premium gitlab and we want to use user or group access level protection https://docs.gitlab.com/ee/api/protected_branches.html#example-with-user--group-level-access.

Any chance to see this becoming ready and reviewed ?

gdubicki:
A friendly ping, @Pigueiras . :)

@barryib : if it will turn out that this PR got stuck, are you interested in taking it over?

Pigueiras:
Yeah, I was on holidays for a long time and right now I don't have the time to work on it to be honest 😅 . If in the future this feature is still not implemented I'll work it, but right now I can't work it sorry.

barryib:
> A friendly ping, @Pigueiras . :)
> 
> @barryib : if it will turn out that this PR got stuck, are you interested in taking it over?

We'll have a look next week and see if someone can work on this.

barryib:
@gdubicki someone in our company will work on this. He'll update this thread himself as soon as possible.

florentio:
Hi, i will be working on it.

gdubicki:
Closing in favour of #273.

PR #233, opened 2021-06-16
Add support for maintaining groups that are shared with other groups

### LINKED ISSUE #150: Enforce group sharing
I need to set up a group to be shared with another group. For example, a groupP1 containing projects will be shared with groupU1. This seems doable but if I manually share groupP1 with groupU2, gitlabform doesn't remove that manual share even if I have the enforcement option enabled. I think this is because the enforcement option currently enforces members but not the shared groups.

Can the permission enforcement option for groups be enhanced so that group sharing is enforced also?

andrewjw:
Hi,

This PR adds a new `group_shared_with` group level config, and code to maintain the list of groups that a group is shared with. It also includes acceptance tests, which pass (at least for me). The code is very much based on the code that maintains group user access.

I believe this fixes #150.

Please review the code, I'm happy to make any changes that might be required.

Many thanks,
Andrew Wilkinson (on behalf of Ocado Technology)

### DIFF
--- a/gitlabform/gitlab/groups.py
+++ b/gitlabform/gitlab/groups.py
@@ -149,3 +149,31 @@ class GitLabGroups(GitLabCore):
             "DELETE",
             expected_codes=[200, 202, 204, 404],
         )
+
+    def get_group_shared_with(self, group):
+        group = self.get_group_case_insensitive(group)
+
+        return group["shared_with_groups"]
+
+    def add_share_to_group(
+        self, group, share_with_group_id, group_access, expires_at=None
+    ):
+        data = {"group_id": share_with_group_id, "expires_at": expires_at}
+        if group_access is not None:
+            data["group_access"] = group_access
+
+        return self._make_requests_to_api(
+            "groups/%s/share",
+            group,
+            method="POST",
+            data=data,
+            expected_codes=[200, 201],
+        )
+
+    def remove_share_from_group(self, group, share_with_group_id):
+        return self._make_requests_to_api(
+            "groups/%s/share/%s",
+            (group, share_with_group_id),
+            method="DELETE",
+            expected_codes=204,
+        )
--- a/gitlabform/gitlab/groups.py
+++ b/gitlabform/gitlab/groups.py
@@ -8,7 +8,7 @@ from gitlabform.gitlab.core import GitLabCore, NotFoundException
 
 
 class GitLabGroups(GitLabCore):
-    @functools.lru_cache
+    @functools.lru_cache()
     def get_group_id_case_insensitive(self, some_string):
         # Cache the mapping from some_string -> id, as that won't change during our run.
         return self.get_group_case_insensitive(some_string)["id"]
--- a/gitlabform/gitlab/groups.py
+++ b/gitlabform/gitlab/groups.py
@@ -1,7 +1,17 @@
+import functools
+import sys
+
+import cli_ui
+
+from gitlabform import EXIT_INVALID_INPUT
 from gitlabform.gitlab.core import GitLabCore, NotFoundException
 
-
 class GitLabGroups(GitLabCore):
+    @functools.lru_cache
+    def get_group_id_case_insensitive(self, some_string):
+        # Cache the mapping from some_string -> id, as that won't change during our run.
+        return self.get_group_case_insensitive(some_string)["id"]
+
     def get_group_case_insensitive(self, some_string):
 
         # maybe "some_string" is the group's path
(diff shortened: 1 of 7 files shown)


gitlabform/gitlabform/processors/group/group_shared_with_processor.py
```diff
@@ -0,0 +1,107 @@
...
+
+        groups_to_set_by_group_id = set()
+
+        for groupname in groups_to_set_by_groupname:
+
+            try:
+                group_id = self.gitlab.get_group_case_insensitive(groupname)["id"]
+            except NotFoundException:
+                cli_ui.error(f"Group {groupname} not found.")
+                sys.exit(EXIT_INVALID_INPUT)
```
gdubicki, line 40: The pattern in our current codebase for this is to convert names to ids inside the `gitlabform.gitlab` package, see https://github.com/egnyte/gitlabform/blob/d926f5561007d82c88e1a242c927993527a61c9f/gitlabform/gitlab/projects.py#L229-L236 as an example. I think that we should do the same thing here.

gdubicki: I guess that this is done this way for performance reasons, to not re-do the same get group id requests over and over, and that makes sense, but I propose to cover this by introducing cache to wrap around some methods that we know that are safe to cache, such as those getting object ids.

gdubicki: WDYT?

gitlabform/gitlabform/test/test_group_shared_with.py
```diff
@@ -0,0 +1,164 @@
...
+            projects_and_groups:
+              {group}/*:
+                group_members:
+                  {users[0]}:
+                    access_level: 50
+                group_shared_with:
+                  {groups[0]}:
+                    group_access_level: 30
+            """,
+        ]
```
gdubicki, line 111: I think that these 2 examples would not remove the sharing even with `enforce_group_members: true`. Isn't the `group_shared_with` content in them exactly the same as at the end of the previous test:
```
            group_shared_with:
              {groups[0]}:
                group_access_level: 30
```
?

gdubicki (changes requested):
Hey @andrewjw! Thank you very much for your contribution! Please check out my review.

gitlabform/gitlabform/processors/group/group_shared_with_processor.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class GroupSharedWithProcessor(AbstractProcessor):
+    def __init__(self, gitlab: GitLab):
+        super().__init__("group_shared_with")
+        self.gitlab = gitlab
+
+    def _process_configuration(self, group: str, configuration: dict):
+        groups_to_set_by_groupname = configuration.get("group_shared_with")
+        print(groups_to_set_by_groupname)
```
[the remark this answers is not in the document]
andrewjw: Sorry, I missed a couple. They've been removed now.

gitlabform/gitlabform/processors/group/group_shared_with_processor.py
andrewjw: Yes, I was worried about the performance of many requests to convert the group name into an id. I've changed this now so the api uses group names, and internally caches the mapping to an id. Please let me know if this looks ok now.

gitlabform/gitlabform/test/test_group_shared_with.py
andrewjw: Thanks, good spot, I've corrected this.

gitlabform/gitlab/groups.py
```diff
@@ -1,7 +1,18 @@
...
+import sys
+
+import cli_ui
+
+from gitlabform import EXIT_INVALID_INPUT
 from gitlabform.gitlab.core import GitLabCore, NotFoundException
 
 
 class GitLabGroups(GitLabCore):
+    @functools.lru_cache
```
gdubicki, line 11: Check out the failed tests, @andrewjw . 
You need to add `()` for this code to work in Python 3.6 and 3.7.
Only in 3.8+ you can omit the parentheses if you don't pass any parameters.
(Btw: I learned that today. 😅 )

andrewjw: I have to admit I only tested it on Python 3.8. Hopefully all the tests will pass now.

gdubicki:
Thank you again, @andrewjw ! Please check out the new v2.1.0 release with your new feature.

Issue #236, opened 2021-07-05
Group share with another group fails on a second run


ejiek:
**The bug description**

There is an awesome new feature to add a group as a member to another group.
Added in #233 by fantastic @andrewjw 

Unfortunately it fails to recognise existing shares (created manually or by GitLabForm itself).

Steps to reproduce:

1. Create config with a group as another group member
1. Run GitLabForm (a successful run)
1. Run GitLabForm again (it fails)

Anonymized error:
```
Warning: Error occurred while processing group groupb, exception:
Request url='https://gitlab.example.com/api/v4/groups/groupb/share', method=POST, data='{'group_id': 28, 'expires_at': None, 'group_access': 40}' failed - expected code(s) [200, 201], got code 409 & body: 'b'{"message":"Shared group The group has already been shared with this group"}''
Traceback (most recent call last):
  File "/gitlabform/gitlabform/gitlabform/core.py", line 466, in process_all
    self.group_processors.process_group(
  File "/gitlabform/gitlabform/gitlabform/processors/group/__init__.py", line 36, in process_group
    processor.process(group, configuration, dry_run, output_file)
  File "/gitlabform/gitlabform/gitlabform/processors/util/decorators.py", line 42, in method_wrapper
    return method(self, project_and_group, SafeDict(configuration), *args)
  File "/gitlabform/gitlabform/gitlabform/processors/abstract_processor.py", line 47, in process
    self._process_configuration(project_or_project_and_group, configuration)
  File "/gitlabform/gitlabform/gitlabform/processors/group/group_shared_with_processor.py", line 81, in _process_configuration
    self.gitlab.add_share_to_group(
  File "/gitlabform/gitlabform/gitlab/groups.py", line 180, in add_share_to_group
    return self._make_requests_to_api(
  File "/gitlabform/gitlabform/gitlab/core.py", line 111, in _make_requests_to_api
    response = self._make_request_to_api(
  File "/gitlabform/gitlabform/gitlab/core.py", line 210, in _make_request_to_api
    raise UnexpectedResponseException(
gitlabform.gitlab.core.UnexpectedResponseException: Request url='https://gitlab.example.com/api/v4/groups/groupb/share', method=POST, data='{'group_id': 28, 'expires_at': None, 'group_access': 40}' failed - expected code(s) [200, 201], got code 409 & body: 'b'{"message":"Shared group The group has already been shared with this group"}''
```
 
[...]

PR #237, opened 2021-07-06
Switch to using group path rather than group name when sharing groups

andrewjw:
Switch to using group path rather than group name when sharing groups with other groups. This allows subgroups to be shared with.

I wasn't able to reproduce #236, but this might fix it (and is probably a good idea anyway!)

Thanks,
Andrew Wilkinson (on behalf of Ocado Technology)

### DIFF
--- a/gitlabform/gitlab/groups.py
+++ b/gitlabform/gitlab/groups.py
@@ -34,12 +34,14 @@ class GitLabGroups(GitLabCore):
                     return group
             raise NotFoundException
 
-    def create_group(self, name, path, visibility="private"):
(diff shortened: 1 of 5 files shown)


gdubicki:
LGTM, thank you for your contribution @andrewjw ! :)

gdubicki:
@andrewjw , @ejiek : sorry for the delay in releasing 2.1.1 with this PR merged but I am having issues with (seemingly) random test failures that are blocking a proper release. Until I manage to release final 2.1.1 you can try 2.1.1rc2, available on PyPI only for now.

Issue #140, opened 2020-11-21
Support for configuring group membership LDAP sync


gdubicki:

gdubicki:
Implemented in v2.2.0, released a moment ago.

PR #249, opened 2021-08-05
Add LDAP sync feature

gdubicki:
Implements #140.

Trying out a new way of defining processors.


gdubicki:
See https://github.com/egnyte/gitlabform/pull/250 instead.

PR #265, opened 2021-09-03
Fix running Gitlab container locally

gdubicki:
(as on GHA it already works well)

* make it REALLY delete GitLab container data
  on each (re)start, as it's the only way to
  make GitLab run this way stable that I have
  found,
* quote paths as we started to rm -rf "$stuff"
  and it could end up badly quickly
* make stop_gitlab work (thanks mkjmdski!)

### DIFF
--- a/dev/run_gitlab_in_docker.sh
+++ b/dev/run_gitlab_in_docker.sh
@@ -41,20 +41,28 @@ cecho b "Pulling GitLab image version '$gitlab_version'..."
 docker pull gitlab/gitlab-ee:$gitlab_version
 
 cecho b "Preparing to start GitLab..."
-existing_gitlab_container_id=$(docker ps -a -f "name=gitlab" --format "{{.ID}}")
-if [[ -n $existing_gitlab_container_id ]] ; then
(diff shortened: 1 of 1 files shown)


dev/run_gitlab_in_docker.sh
```diff
@@ -41,20 +41,28 @@ cecho b "Pulling GitLab image version '$gitlab_version'..."
...
+  fi
+}
+stop_gitlab
 
-mkdir -p $repo_root_directory/config
+mkdir -p "$repo_root_directory"/config
 if [[ -f Gitlab.gitlab-license ]] ; then
   cecho b "EE license file found - using it..."
-  cp Gitlab.gitlab-license $repo_root_directory/config/
+  cp Gitlab.gitlab-license "$repo_root_directory"/config/
```
mkjmdski, line 62:
```bash
"${repo_root_directory}/config"
```

mkjmdski (approved):

gdubicki:
Did didn't work out great in the end, closing then.

PR #268, opened 2021-09-04
Start accepting access levels names (only tests)

gdubicki:
apart from their level numbers, as defined in https://docs.gitlab.com/ee/api/access_requests.html#valid-access-levels .


gdubicki:
Incorporated into #309 with 1e1a620.

PR #274, opened 2021-09-15
Run linter and tests workflow on every PR

barryib:
@gdubicki This PR tries to solve https://github.com/egnyte/gitlabform/pull/273#issuecomment-919367567

When we use `on.event.branches` like this :

```yaml
on:
  push:
    branches:
      - '**'
  pull_request:
    types: [opened]
```

The workflow is only ran in a the base are evaluated. That means it'll only run workflow on branches in `egnyte/gitlabform` project. All unknown branch refs are ignored. The Workflow will also run when a PR is opened (just once) => will run for the `opened` event.

### DIFF
--- a/.github/workflows/linter-and-tests.yml
+++ b/.github/workflows/linter-and-tests.yml
@@ -1,6 +1,10 @@
 name: Linter and tests
 
-on: [pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - main
 
 jobs:
   black-formatting:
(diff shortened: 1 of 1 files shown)


.github/workflows/linter-and-tests.yml
```diff
@@ -5,7 +5,6 @@ on:
     branches:
       - '**'
```
barryib, line 6: Do we need to do this only on `main` branch because other branch will be merge from a PR ?

gdubicki: Let's try this first and maybe limit the runs later.

Issue #275, opened 2021-09-17
skip_groups doesn't work for subgroup


aleung:
**Describe the bug**

On our GitLab server, there is a top level group `ei` and several levels of subgroups under it.

My config file has subgroup name set into `skip_groups`:

```yml
projects_and_groups:
  "*":
    # configurations
skip_groups:
  - ei/demo
```

Run gitlabform on the group `ei`:
```
gitlabform ei
```

I expect the projects `ei/demo/*` are skipped, but all of them had be configured.

Read the source code, I saw that `_remove_skipped_groups` is called only when target is `ALL` or `ALL_DEFINED` and isn't applied if target is a group.

https://github.com/egnyte/gitlabform/blob/dbb53da60c5366295a4269fec2577b34be1a862f/gitlabform/input.py#L28

**GitLabForm version**

2.2.0

gdubicki:
Hi @aleung ! Thanks for reporting this.

I am not sure if we should treat it as a but of a feature request though. This kind of matters because I try to fix bug ASAP while the feature requests are preferred to be developed by the community with PRs. 😅 

Anyway, if you got this far that you saw the code that would need to be changes to ~fix~ implement it, perhaps you could provide a Pull Request?

gdubicki:
Implemented/fixed in v2.4.0, released a moment ago.

aleung:
#276 hasn't fixed it.

```
projects_and_groups:
  "*":
    # configurations
skip_groups:
  - ei/demo
```

If the CLI parameter points to the subgroup , it's skipped as expected.
```
gitlabform ei/demo
```

But when the parameter points to the parent group, the subgroup isn't skipped.
```
gitlabform ei
# ei/demo isn't skipped
```

I'll check if I'm able to raise a PR to address it.

Issue #276, opened 2021-09-20
`skip_group` and `skip_project` support wildcard


chris-workingmouse:
We have a lot of old/demo/test groups, subgroups and projects. The current `skip_group` and `skip_project` only does a lookup in a list which requires us to list out the hundreds of groups and projects individually to be skipped.

It would be must easier if these could be skipped using a wildcard similar to how the wildcard for `projects_and_groups` works.

gdubicki:
Hi @chris-workingmouse !

Thanks for the suggestion. It makes a lot of sense. I was thinking that „skip_*” feature needs a revamp since v2 has beed released.

This should be considered together with #275 but also #139 to design the same syntax for referencing projects, groups and subgroups everywhere.

We could start with exactly the same syntax as in `projects_and_groups` though: `foo/bar` - exact match project/subgroup, `foo/*` - all projects and subgroups under group `foo`.

As always: PR are welcome! :)

chris-workingmouse:
Hi @gdubicki,

I have submitted a MR with a simple solution for the current syntax, attempting not to disrupt too much of the existing code base.

Let me know what you think.

gdubicki:
Fixed with your PR, released in v2.4.0 a moment ago. Thanks again, @chris-workingmouse ! :)

PR #277, opened 2021-09-20
#276 Add wildcard support for skip_groups and skip_projects

chris-workingmouse:
Implements #276

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -461,7 +461,9 @@ projects_and_groups:
 skip_projects:
   - my-group/this-project-will-not-be-processed-with-gitlabform-when-running-for-my-group
   - my-group/and-this-too
+  - my-group/everything-under/*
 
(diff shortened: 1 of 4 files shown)


gdubicki:
Thanks for the instant and so nicely tested contribution, @chris-workingmouse ! ❤️

I think that the lowered coverage is caused by [introducing tests for GitLab EE here](https://github.com/egnyte/gitlabform/pull/273#issuecomment-922362258) that do not run in a fork. I will think about how to deal with that but will make a new release of the app with the new feature regardless if I fix the coverage issue or not - I don't want to hold off a new feature because of this technical issue. 😊

gdubicki:
Also fixes #275.

chris-workingmouse:
@gdubicki Thanks for reviewing and releasing so quickly!

PR #280, opened 2021-09-24
Config commit message for file operations

### LINKED ISSUE #278: Support configurable commit message for file operations
GitLabForm generates commit message for file new/update/delete operation:

https://github.com/egnyte/gitlabform/blob/ab7b8da4410767dd93713ceedddd1d5d0737edbe/gitlabform/gitlabform/processors/project/files_processor.py#L191

Our GitLab has hook configured to validate commit message format. Any commit which doesn't follow the pattern is rejected. Configurable commit message is needed in this use case.

aleung:
Fix #278


aleung:
Something wrong in my commits. Will create a new one later.

PR #281, opened 2021-09-24
Config commit message for file operations

### LINKED ISSUE #278: Support configurable commit message for file operations

aleung:
Fix #278


aleung:
No idea why my PR reduces the coverage rate. Looks like that the decrease was because `ProjectPushRulesProcessor` not run.

gdubicki:
Thanks for the contribution @aleung !

Can you please also add a use example to `config.yaml` and a test that shows that without custom commit message the default one is still used?

And I will deal with the coverage drop issue - I guess that I need to make coverage.io to not count coverage for the tests that require GitLab EE license. :(

aleung:
@gdubicki
Example and test are updated. Please have a look.

PR #284, opened 2021-09-29
fix getting members list for a project to include all members

trissanen:
…not just the first page

### DIFF
--- a/gitlabform/gitlab/members.py
+++ b/gitlabform/gitlab/members.py
@@ -50,7 +50,7 @@ class GitLabMembers(GitLabCore):
 
     def get_members_from_project(self, project_and_group_name):
         members = self._make_requests_to_api(
-            "projects/%s/members", project_and_group_name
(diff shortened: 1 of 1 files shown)


trissanen:
how did the code coverage go down on this change? Is this something I should do something about? I don't think writing a completely unrelated unit test would make this PR any better

gdubicki:
Thanks for the fix @trissanen !

gdubicki:
> how did the code coverage go down on this change? Is this something I should do something about? I don't think writing a completely unrelated unit test would make this PR any better

Sorry for that, it is not cause by your PR. The code coverage has dropped because the tests running from a fork don't have access to the secrets in this repo and we keep a GitLab license in it to run the tests using Premium features. On one hand we *do* want to know that these features are covered by the tests but on the other we *do not* want to see false drops in coverage from the PRs. I don't know how to handle this yet. :(

PR #309, opened 2021-11-26
Start accepting access levels names

gdubicki:
Implementation and tests.


gdubicki:
Probably would require https://github.com/wwkimball/yamlpath/issues/151 to be fixed in the upstream library first to add this without having to update the major version number due to YAML 1.1 -> 1.2...

PR #319, opened 2021-12-26
Vagrant using GCP as a workaround for problem with running GitLab in Docker on Apple Silicon Mac

gdubicki:


gdubicki:
I will just keep using my Intel Mac to run those tests for now.

Issue #326, opened 2022-01-14
Allow breaking configuration inheritence


amimas:
I've started using the v2 (2.10.0). Lots of improvements have been made from v1. This post is a feature request for being able to break config inheritance.

Below is an example config of what it might look like for a typical group.

```yml
config_version: 2
gitlab:
  api_version: 4

projects_and_groups:
  my-group/*:
    # Settings for ALL projects under 'my-group' group and the group itself
    group_settings:
      # Settings of the parent group
      description: Contains projects for example product.
    members:
      # Members for ALL projects.
      enforce: true
      groups:
        user-group-1:
          group_access: 30  # Developer role
        user-group-2:
          group_access: 30  # Developer role
    project_settings:
      # Settings of ALL projects
      visibility: internal
      issues_access_level: enabled
      request_access_enabled: false

  my-group/special-private-project:
    # This is a special project. It should be private and have specific user permission.
    project_settings:
      visibility: private
    members:
      # Only following users should have permission. Do not inherit permissions from common settings
      groups:
        user-group-2:
          group_access: 30
       users:
          john:
            access_level: 40
```

With the above config, the project `my-group/special-private-project` gets members from the common settings (configured in `my-group/*` section) in addition to what's configured in that project directly.

I think it makes sense that this is how it works by default. It'd be nice if I could specify in that project specific config to not inherit common settings so that I could manage "exception" scenarios. In the above example, I want to break the inheritance of the `members` section only but inherit other settings.  I guess it could be perceived that someone might want to break the inheritance in other areas/sections as well

Right now my work-around is following. Only problem with this work-around is that I need to list out every single project in the config just so that I could have exception for a single project. So, I'm not able to take full advantage/power of gitlabform's config.

```yml

config_version: 2
gitlab:
  api_version: 4

default_project_access_by_usergroup: &default-project-access-by-usergroup
  user-group1:
    group_access: 30  # Developer
  user-group2:
    group_access: 30  # Developer

projects_and_groups:
  my-group/*:
    # Settings for ALL projects under 'my-group' group and the group itself
    group_settings:
      # Settings of the parent group
      description: Contains projects for example product.
    members:
      # Members for ALL projects.
      # Leave this section empty so that each project can specify what permission to be set.
      # Otherwise we cannot configure exception using gitlabform
      enforce: true
    project_settings:
      # Settings of ALL projects
      visibility: internal
      issues_access_level: enabled
      request_access_enabled: false

  my-group/project1:
    members:
[...]

PR #339, opened 2022-02-23
Allow Breaking Configuration Inheritance #326

ep-linden:
With the hierarchical inheritance structure in GitLabForm V2, users cannot break the inheritance. To see a better example of this use case please see Issue #326.


ep-linden:
@gdubicki @amimas

ep-linden:
@gdubicki Please let us know if you had another implementation in mind.

gitlabform/configuration/core.py
```diff
@@ -162,13 +163,64 @@ def is_group_skipped(self, group) -> bool:
...
+                elif kwargs["level"] == "common":
+                    level = "common"
+                    has_invalid_flag = True
+            if has_invalid_flag:
+                fatal(
+                    "The inheritance break flag cannot be placed at the {config_level} level\n"
+                    "because {config_level} level is the highest level in the configuration file.\n".format(
+                        config_level=level
+                    ),
+                    exit_code=EXIT_PROCESSING_ERROR,
```
gdubicki, line 188: This is an input error, not a processing error.

I added commit to clarify when each exit code should be used here https://github.com/gdubicki/gitlabform/pull/339/commits/554be4dfb807d7a0ab963dfdc51dc06bd11ecc3f . Is this information enough?

gitlabform/configuration/core.py
```diff
@@ -162,13 +163,64 @@ def is_group_skipped(self, group) -> bool:
...
         :return: merge more general config with more specific configs.
                  More specific config values take precedence over more general ones.
         """
-        return dict(merge({}, more_general_config, more_specific_config))
+
+        merged_dict = merge({}, more_general_config, more_specific_config)
+
+        def break_inheritance(specific_config, parent_key=""):
+            for key, value in specific_config.items():
+                if "inherit" == key:
```
gdubicki, line 206: It's not enough to have the key `inherit` - it must also be set to boolean `False`

gitlabform/configuration/core.py
```diff
@@ -162,13 +163,64 @@ def is_group_skipped(self, group) -> bool:
     def is_skipped(self, an_array: list, item: str) -> bool:
         pass
 
+    @staticmethod
+    def validate_break_inheritance_flag(config, **kwargs):
```
gdubicki, line 167: Nit: I think that we don't need `**kwargs` here, it's making code harder to read. Plain parameters perhaps with default values would be easier.

gitlabform/configuration/core.py
```diff
@@ -162,13 +163,64 @@ def is_group_skipped(self, group) -> bool:
...
+        has_invalid_flag = False
+        level = None
+        for key, value in config.items():
+            if "inherit" == key:
+                if kwargs["parent"] == "empty" and (
+                    kwargs["level"] == "group"
+                    or kwargs["level"] == "project"
+                    or kwargs["level"] == "subgroup"
```
gdubicki, line 175: Nit: syntax `kwargs["level"] in ["group", "project", "subgroup"]` would be shorter and more readable.

gdubicki (changes requested):
Thank you for the PR, @ep-linden! Sorry if the review is premature as the PR is still in Draft. I wanted to give you some feedback though.

The great thing is that your tests show that it works! 🥳

The slightly worse is that I only scraped over the surface and have pretty trivial comments for you now. Unfortunately I don't understand your core code (yet) so I don't have comments about the logic. :(

Do you plan to change it?

gitlabform/configuration/core.py
ep-linden: I left it this way on purpose because I was unsure if this is the "inheritance-break flag" that everyone is set on. I did remember there was a bit of a discussion on what we wanted to rename it to. 

However, since I decided to go with what you suggested initially, I will add that check as well.

gitlabform/configuration/core.py
ep-linden: Yup, it's enough, thanks!

gitlabform/configuration/core.py
ep-linden: I have refactored this function so it is a lot simpler to understand. After I explained what the code was supposed to do it made me realize, I didn't need the other parameters.

gitlabform/configuration/core.py
ep-linden: After the changes, this does not exist anymore.

gitlabform/configuration/core.py
```diff
@@ -162,13 +163,64 @@ def is_group_skipped(self, group) -> bool:
...
+                ):
+                    level = kwargs["level"]
+                    has_invalid_flag = True
+                elif kwargs["level"] == "common":
+                    level = "common"
+                    has_invalid_flag = True
+            if has_invalid_flag:
+                fatal(
+                    "The inheritance break flag cannot be placed at the {config_level} level\n"
```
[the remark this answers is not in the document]
ep-linden: Made the changes.

gitlabform/configuration/projects_and_groups.py
```diff
@@ -41,15 +41,35 @@ def get_effective_config_for_project(self, group_and_project) -> dict:
         debug("Effective group/subgroup config: %s", to_str(group_config))
 
         project_config = self.get_project_config(group_and_project)
-        debug("Project config: %s", to_str(project_config))
+        debug("\nProject config: %s", to_str(project_config))
```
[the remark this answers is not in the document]
ep-linden: Yup, removed.

gdubicki:
If you want to discuss this more synchronously over chat @ep-linden @amimas @jccl, then please use https://join.slack.com/t/gitlabform/shared_invite/zt-164l3iwxd-hLqZX7v0AU4RsvVajYFtYQ link to join gitlabform's Slack. :)

tests/unit/configuration/test_inheritance_break_projects_and_groups.py
```diff
@@ -0,0 +1,337 @@
...
+        members:
+          inherit: false
+          users:
+            user3: maintainer
+            user4: maintainer
+    """
+    return Configuration(config_string=config_yaml)
+
+
+def test__get_effective_config_for_project__with_invalid_inheritance_break_set_at_common_level(
```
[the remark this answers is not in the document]
ep-linden: Good point, I did notice having to reference the fixtures became a bit overwhelming while I was creating the tests.

Issue #343, opened 2022-03-12
More Contributors / Developers / Maintainers needed!


gdubicki:
Hi Everyone!

We have a few great ideas on how to move forward but I personally just don't have the time to implement them all.

I would love to get more people cooperating on this project, especially having more Core Developers/Maintainers that would move the big things forward, like v3 design/implementation but also Feature Developers, Documentation Contributors and everyone else what wants to improve this project is very welcome to join.

Please let me know if you are interested, what you would like to do and what you need from me to get you started. A chat or a call with an introduction? - we can do it. Better docs about something specific? - can be arranged. Make some code more readable? - I will do my best. :) More permissions to the repo/project? - sure, why not.

Let me know!

jimisola:
@gdubicki How many are there that you believe are able and willing to cooperate now?

Would it be inappropriate to have a one-liner when gitlabform runs such as "The GitLabForm is in need of contributors to keep up with GitLab changes and features. Please see: <this issue>".

Have GitLab ever shown an interest in the project?

grzesuav:
hi @gdubicki , I believe local testing env would be nice, so an easy way to spin up gitlab instance locally (docker/docker-compose/shell ?) and config to run gitlabform loccaly against this instance. Is this something you have somewhere ?

gdubicki:
> hi @gdubicki , I believe local testing env would be nice, so an easy way to spin up gitlab instance locally (docker/docker-compose/shell ?) and config to run gitlabform loccaly against this instance. Is this something you have somewhere ?

Hi @grzesuav! We already do have docs and a script for that, please check out https://gitlabform.github.io/gitlabform/contrib/local_development/#running-acceptance-tests-using-gitlab-instance-in-docker.

gdubicki:
I wonder how many people would cooperate on the project if it would be using the semi-standard `python-gitlab` under the hood?

Of course I am referring to the effort that has already started in #442.

cc @nejch

nejch:
Definitely @gdubicki!

At the very least I would be happy to always work on any missing endpoints in python-gitlab, and potentially integrating them in here.

Then there's also GitLab's quirks that we are all probably working around in different ways, so solving things like https://github.com/gitlabform/gitlabform/issues?q=is%3Aissue+is%3Aopen+label%3A%22%F0%9F%A6%8Agitlab+issue%22 together is another option if it affects both projects.

adam-moss:
[...]

Issue #331, opened 2022-02-05
Placeholder for version 5 of config.yml


jimisola:
Work has started with the syntax for version ~3~ ~4~ 5 of config.yml.

This issue will be used a placeholder to link with other issues and feature request as well a providing a place for discussion (in addition to slack).

The current draft version of Configuration Syntax v5 can be found [here](https://github.com/gitlabform/gitlabform/wiki/Configuration-Syntax-v5-Draft) in our wiki.

**Requirements:**

-  [ ] YAML 1.2
  - [ ] highlight use of node anchors
- [ ] jinja2 templates (possibly with the use of dynaconf)
  - [ ] [built-in filters](https://jinja.palletsprojects.com/en/3.0.x/templates/#builtin-filters)
  - [ ] standard custom filter ENV for environment variables
  - [ ] any other custom filter?
  - [ ] what’s needed to handle different input data formats for rendering (json, yaml, xml etc)?
- [ ] work with document separator “---”
- [ ] work with multiple files, e.g. gitlabform file1 file2 file3
- [ ] prepare to minor adaptation of format when GitLab merges groups and projects into [namespaces](https://forum.gitlab.com/t/restructure-gitlab-with-namespaces-work-items-etc/64229/3)
  - [ ] can we be backwards-compatible with or will the API change so that there will have to be two versions of gitlabform (pre- [ ] and post-namespaces)?
- [ ] work with personal projects
- [ ] inheritance:
  - [ ] do-not-allow-inherit-overwrite
  - [ ] do-not-inherit

**Questions:**

- [~] Shall we plan for use of python-gitlab?
- [ ] Shall we plan for update, add and delete?
  - [ ] Currently gitlabform support update with a exception (add/delete) for keys and variables 
  - [ ] with python-gitlab it should be easier to support add and delete
- [ ] How should inheritance be combined with group/project members and “enforce”?
- [ ] GitLab mixes update and edit? Shall we use their mix or just settle for one of them? (follow-up: @jimisola )
- [ ] members: both project and groups?
  - [ ] group_ldap_links: can only be applied to groups or projects as well? only groups
  - [ ] What is this for: https://docs.gitlab.com/ee/api/groups.html#create-a-link-to-share-a-group-with-another-group ?
    - [ ] Invite members
    - [ ] Invite groups
    - [ ] sync with LDAP groups
- [ ] tags: belongs to project? **Yes.**
[...]

PR #389, opened 2022-06-25
V3

gdubicki:
Python 3.6 has been EOLed in Dec 2021 and since then a lot of our dependencies have dropped support for it.

As changing the required Python version to a higher one is a backward-incompatible change I wanted to combine it with the big configuration syntax upgrade described in #331.

However the latter is getting delayed because I can't find enough time to do it (see #343).

That's why I would like to release v3 now which will just require a Python upgrade but not any configuration syntax changes and move the planned configuration syntax change to v4.

*Update*: the scope of this release has been increased, see the last comments and commits below.


gdubicki:
Pre-release created: https://github.com/gitlabform/gitlabform/releases/tag/v3.0.0rc1

PR #392, opened 2022-07-03
HTTP2 (thanks to switch from requests to httpx)

gdubicki:


gdubicki:
I expected that for this use case switching from HTTP1 to HTTP2 might not give performance improvements but I was hoping that I was wrong and that's where this PR comes from.

Unfortunately I was not wrong and the real-life performance for Egnyte's 1001 repos in 5 groups managed with this app the processing time with HTTP2 is practically the same as for HTTP1.

But it was still an interesting experiment that made me learn that:
* switching from requests to httpx *is* very easy,
* ...but there are some things that httpx doesn't have built-in, in particular [retries](https://github.com/encode/httpx/issues/947#issuecomment-629757747) - it's not a blocker, you can use f.e. [tenacity](https://tenacity.readthedocs.io/en/latest/) to implement it on your own, but it requires some extra work.

PR #402, opened 2022-08-23
Fix Integrations documentation

L154x:
The documentation was using `services` as key which is outdated. This was recently replaced with `integrations`. The tests were already updated in a prior commit but the documentation was forgotten. See here: https://github.com/gitlabform/gitlabform/commit/5bae26f09f1dca5307530db51376f90130d8d42e


gdubicki:
Thanks for the contribution, @L154x!

Issue #388, opened 2022-06-23
MR approvers not working in some cases


weakcamel:
In several latest Gitlab releases, Gitlab doesn't support the "approvers" endpoint which gitlabform has historically been using to manage approvers... and then also approver rule (a single one).

That endpoint is no longer available on Gitlab 13, 14, 15 and I'm aware that https://github.com/gitlabform/gitlabform/pull/176 is unlikely to be merged in. It's unfortunate yet completely understandable  - if this functionality isn't essential for the author and community hasn't managed to provide a fix (well, myself included), it's only fair.

I wonder though, could it be useful perhaps to remove managing approvers from the feature list and mark using `approvers` + `approver_group` keys as obsolete? As it is, running Gitlabform accepts them but any configuration given has no effect.

gdubicki:
Hi @weakcamel!

I am surprised that you write that it doesn’t work because we are using this at Egnyte with semi-latest GitLab and we do run the acceptance tests for it with the latest GitLab.

Perhaps the problem is that we use an old GitLab license for all of these purposes and it allows only a single approver rule?

We will upgrade to the Premium license in a month or two so then I can test and fix the bug.

If you or anyone wants to help make it happen faster then you either would have to share you license (it is allowed for dev purposes) securely with this project or just provide a PR with the fix.

Whatever we will do please know that this feature is NOT abandoned and will be working again. :)

weakcamel:
Hello,

I've just checked again. Sorry for a false alert (again), it does seem to work if you use `remove_other_approval_rules` (although I was 99.99% sure I tested that with this flag as well - maybe just at the wrong level in Yaml tree?).

> Perhaps the problem is that we use an old GitLab license for all of these purposes and it allows only a single approver rule?

The `approvers` endpoint has been removed in 13.11 (https://gitlab.com/gitlab-org/gitlab/-/blob/master/CHANGELOG-EE.md#13110-2021-04-22) and I thought gitlabform was using that, but no longer - that was my wrong assumption.

I just ran gitlabform with `-d` against a project and fiddled with Gitlab's UI and I think I know a bit more about what's happening: if you remove all custom rules, there's still one applying to "Any eligible approver" which you can't remove from the UI, but can with REST API:

![Screenshot 2022-06-24 at 09 03 40](https://user-images.githubusercontent.com/2445059/175492536-4dc40dcd-595d-4d80-b134-b532a613518d.png)

[...]

Issue #290, opened 2021-10-08
Add support for dry-run to print out *only* the things that will be changed


otani88:
Hi. We are faced with a problem when we're running dry-run with verbosity and gitlabform shows diff in cases where nothing is going to be changed.

Output example:
```
                                  default_branch: "master"   => "master"
           only_allow_merge_if_pipeline_succeeds: true       => true
only_allow_merge_if_all_discussions_are_resolved: true       => true
               resolve_outdated_diff_discussions: true       => true
             printing_merge_request_link_enabled: true       => true
                           snippets_access_level: "disabled" => "disabled"
                               wiki_access_level: "enabled"  => "enabled"
                                    wiki_enabled: true       => true
                                    merge_method: "ff"       => "ff"
                                      visibility: "private"  => "private"
                remove_source_branch_after_merge: true       => true
Skipping section 'project_push_rules' - not in config.
Processing section 'merge_requests' in dry-run mode.
```

**GitLabForm version**

`2.5.0`

**GitLab version**

`14.0.0-ee`

mkryva:
up

gdubicki:
Hi @mkryva, sorry for the long response time!

This is actually how this feature has always worked...

tmeijn:
As a workaround I use this when running it in CI:

```shell
# sed explanation:
# 1. Removes all lines with not in config. It adds nothing valuable
# 2. Removes all lines stating processing in dry-run mode.
# 3. Removes all lines which indicate no change, for example true => true
# 4. Adds color coding to lines that do change
gitlabform ALL_DEFINED ${DRY_RUN:+--noop} --verbose \
  | sed -e '/not in config.$/d' \
  | sed -e '/in dry-run mode.$/d' \
  | sed -re '/:\s(\S*|".*")\s+=>\s\1/d' \
  | sed -re 's,.*=>.*,\x1B[1;33m&\x1B[0m,'

```

lmaughan-ep:
We are also trying to use `--noop` in a CI pipeline with gitlabform 3.4.1. In our case we hope to use the dry-run in branch pipelines as a way to show what might be changed, so having a way to limit the output to only the expected changes would be very useful. The last two `sed` tricks used by @tmeijn in the comment above don't work in our case because the ghcr.io Docker image for gitlabform is based on Alpine, and sed behaves differently under POSIX. 

+1 on this issue; it would be great if we had a way to see only changes in the output, with the option to colorize it, using the official Docker image.

[...]

PR #423, opened 2022-10-11
Add protected environments processor

zanella:
@gdubicki , sorry for the sloppy code, but I thought it best to ask for feedback 😬 

I left `TODO` comments on things I was not sure about or where you'll probably have a better idea on how to solve the issue I face.

I tested it against a live Gitlab installation, it works, I'll look into writing the tests next.


gitlabform/processors/project/protected_environments.py
```diff
@@ -0,0 +1,34 @@
...
+            "protected_environments",
+            gitlab,
+            list_method_name=gitlab.list_protected_environments,
+            add_method_name=gitlab.protect_a_repository_environment,
+            edit_method_name=gitlab.update_a_protected_environment,
+            delete_method_name=gitlab.unprotect_environment,
+            # TODO: I had to define "name" inside the cfg, it should get the section's name as the entity name, e.g.
+            # protected_environments:
+            #   foo:
+            #     name: "foo" <- This is redundant
```
gdubicki, line 21: Yeah, this is not perfect that you have to do this... But I would still stick to it for consistency with other sections.

gitlabform/processors/project/protected_environments.py
```diff
@@ -0,0 +1,34 @@
...
+            #
+            defining=Key("name"),
+            required_to_create_or_update=And(Key("name"), Key("deploy_access_levels")),
+        )
+
+    def _print_diff(self, project_or_project_and_group: str, entity_config: dict):
+        # TODO: yeah... I didn't get how this is supposed to work :-(
+        # Should I receive (from the super class) the "live" cfg from Gitlab ?
```
gdubicki, line 29: Don't remember also. 😅 I will get back to you on this in a day or two.

zanella:
> Hey @zanella! Thanks for the PR! I added some comments. We can try jumping on our Slack if you prefer to chat about it. Let me know!

Yeah, sounds good, do you need to generate an invite link for that ?

gitlabform/processors/abstract_processor.py
```diff
@@ -89,7 +91,15 @@ def _needs_update(
             entity_in_gitlab.keys()
         )
         for key in keys_on_both_sides:
+            # TODO -> The return from Gitlab has:
+            #  - all the keys, even if their values is None, e.g.
+            #  - keys which are not part of the request API
+            # entity_in_gitlab[deploy_access_levels] -> [{'access_level': 40, 'access_level_description': 'Maintainers', 'user_id': None, 'group_id': None, 'group_inheritance_type': 0}]
+            # != entity_in_configuration[deploy_access_levels] -> [ordereddict([('access_level', 40)])]
```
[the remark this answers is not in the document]
zanella: I mean smth like: [5254fbb](https://github.com/gitlabform/gitlabform/pull/423/commits/5254fbb4f79e1760090e783c095ffd103c31959f)

The "entity" returned by Gitlab is a list[dict], and these dicts have more data than what's defined in the cfg (yaml file), basically the key intersection that you do already in `_needs_update`.

I'm afraid of doing that recursively (0: checking if the returned entity is a list, 1: if so iterate it, goto: 0) so I added a custom "analyzer" for a case I know already will be tricky

Though now thinking about it, if this approach makes sense I'll move this function to the base class 🤔

gdubicki:
> > Hey @zanella! Thanks for the PR! I added some comments. We can try jumping on our Slack if you prefer to chat about it. Let me know!
> 
> Yeah, sounds good, do you need to generate an invite link for that ?

https://join.slack.com/t/gitlabform/shared_invite/zt-1i5h4dds0-NO4x63W5_GWL~WWrWy9DQg

gitlabform/gitlab/project_protected_environments.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+
+    def protect_a_repository_environment(
+        self, project_and_group_name: str, protected_env_cfg: dict
+    ):
+        return self._make_requests_to_api(
+            "projects/%s/protected_environments",
+            project_and_group_name,
+            method="POST",
+            json=json.loads(json.dumps(protected_env_cfg)),
```
gdubicki, line 19: Nit: doesn't much simpler `data=protected_env_cfg` do the same thing?

gitlabform/processors/multiple_entities_processor.py
```diff
@@ -16,22 +16,52 @@ def __init__(
...
+        edit_method_name: Union[str, Callable[[str, str, Any], None], None] = None,
     ):
+        """
+
+        :param configuration_name: The key in the YAML cfg file that determines the start of this Entity's section
+        :param gitlab:
+        :param list_method_name:
+        :param add_method_name:
+        :param delete_method_name:
+        :param defining: TODO -> what is this for ?
```
gdubicki, line 35: This field contains an expression that tells which fields of the entity in GitLab are _identifying_ it.

It's used for finding matches between entities in your config and the ones in GitLab.

Usually it's just the a single field like "name", "key" or "title", but sometimes it's more complicated.

An example of that case is GroupLDAPLinksProcess where we have `And(Key("provider"), Or(Key("cn"), Key("filter")))` that can be interpreted as "if you find group LDAP link in config and in GitLab that have the same value of field "provider" and the same value of at least one of the fields "cn" and "filter", then they corresponding".

If it's still not clear, please let me know. :)

gitlabform/processors/multiple_entities_processor.py
```diff
@@ -16,22 +16,52 @@ def __init__(
...
+        :param required_to_create_or_update: Entries which (according to the API documentation) are required fields
+        :param edit_method_name: TODO -> how does it handle the actual and new cfg diff ?
```
gdubicki, line 37: The `edit_method_name` method is just called with the whole config of a given entity from the config. We don't do anything smart here.

gitlabform/processors/abstract_processor.py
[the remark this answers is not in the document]
gdubicki: Thanks, now I get it.

I also only now have read the docs of this API and I realized that apart from the list of dicts in`deploy_access_levels` it also contains `approval_rules` similar to the multiple approval rules that I am only starting to working on in #388.

Please note that for both of these it would be best to make the GitLabForm syntax slightly easier than the raw API. In particular, we should allow passing usernames and group names instead of their ids. Please see how we have dealt with a similar problem for branch protection in GitLab Premium: https://gitlabform.github.io/gitlabform/reference/protected_branches/#premium-only-features

gitlabform/processors/project/protected_environments.py
gdubicki: No, this is not magic. ;) You would have to implement getting the entities from GitLab. See how it's done for a SingleEntityProcessor here https://github.com/gitlabform/gitlabform/blob/16af88b5d67faaf67a7cb90640b977168820c9c8/gitlabform/processors/single_entity_processor.py#L50-L58.

But I wouldn't worry about this for now, it's fine to not implement it.

This feature should be added globally and implemented differently to make more sense. See also #290.

amimas:
What will the config syntax be for protected environments when this is merged?

zanella:
> What will the config syntax be for protected environments when this is merged?

  ```
"group_name/project_name":
    protected_environments:
      foo:
        name: foo
        deploy_access_levels: &an_anchor_can_be_used_to_reuse #
          - access_level: 40
            #group_inheritance_type: 1
          - user_id: 123  
            #group_inheritance_type: 0
          - user: john_doe
      blah:
         name: blah
         deploy_access_levels: *an_anchor_can_be_used_to_reuse    
```

File: gitlabform/processors/multiple_entities_processor.py [hunk 7]
```diff
@@ -16,22 +16,52 @@ def __init__(
         self,
         configuration_name: str,
         gitlab: GitLab,
-        list_method_name: str,
-        add_method_name: str,
-        delete_method_name: str,
+        # TODO: Should the option of passing as str be deprecated ? It makes life harder (for Mypy and for the devs),
```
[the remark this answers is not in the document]
zanella: #424

gdubicki:
Final is not compatible with Python 3.7, @zanella...

gdubicki:
I did a fix (there would be a missing link to the new article in the docs' index) and some updates to the docs and now I think we are good to go. Thank you for all the work on this feature, @zanella!

gdubicki:
This code has been pre-released in [v3.3.0rc1](https://github.com/gitlabform/gitlabform/releases/tag/v3.3.0rc1). Please feel free to check it out before a final release with this that should come within a few days.

amimas:
I'm away from work next few weeks. So I can't test it out at the moment but really looking forward to it. Do we want visit the config syntax before this feature is released out of RC?

https://github.com/gitlabform/gitlabform/pull/423#issuecomment-1279287453

There's a redundant key name, which is treated as a label instead of the actual name of the env. Also the parent config, `protected_environments` is a departure from protected branch config, which is called `branches`.

gdubicki:
> There's a redundant key name, which is treated as a label instead of the actual name of the env.

Yes, on second thought it would be nice for this to be cleaned. We could do this relatively easy with a config transformation where:
```yaml
    protected_branches:
      foo:
        (...)
```
...would internally be converted to:
```
    protected_branches:
      foo:
        name: foo
        (...)
```
Wdyt, @zanella?

> Also the parent config, `protected_environments` is a departure from protected branch config, which is called `branches`.

I think you are right. It would be better to rename this to just `environments`, it would be more consistent.

But until we would also implement #164 it might be disappointing to some users that you cannot really manage all the envs but only protect them.

Also it's a bit of a problem with how `enforce: true` would work - now it _unprotects_ all undefined envs while if this section would be called `environments` I would expect it to _delete_ all undefined envs...

amimas:
> Also it's a bit of a problem with how enforce: true would work - now it unprotects all undefined envs while if this section would be called environments I would expect it to delete all undefined envs...

Ahh... I see. Didn't know the current implementation detail about this feature. I think you're right. I would also expect `enforce: true` under `environment` to delete all the other envs that are not in the config and I can see how that can also be unexpected. We probably need to hash out the details for regular environment config.

> But until we would also implement https://github.com/gitlabform/gitlabform/issues/164 it might be disappointing to some users that you cannot really manage all the envs but only protect them.

This made me think perhaps we should have a separate key under `environments` to indicate whether an env should be protected or not. For example: `protected: true`. This also follows the config syntax for branch protection. Protected env is a subset of overall environment feature. So, we can implement the rest of it for #164  later if/as needed.

PR #467, opened 2022-12-31
Reproduce and fix Siythrun's issue from #388

gdubicki:


gdubicki:
I got stuck and reported an issue to yamlpath library author here: https://github.com/wwkimball/yamlpath/issues/206

gitlabform/configuration/transform.py
```diff
@@ -39,82 +41,92 @@ def transform(self, configuration: Configuration) -> None:
...
-                for node_coordinate in processor.get_nodes(path):
-                    user = node_coordinate.parent.pop("user")
-
-                    node_coordinate.parent["user_id"] = self.gitlab._get_user_id(user)
-            except YAMLPathException as e:
-                # this just means that we haven't found any keys in YAML
-                # under the given path
-                pass
+        try:
+            for node_coordinate in processor.get_nodes(
```
wwkimball, line 78: When querying data -- especially when using wildcards -- be sure to set `mustexist=True` as shown in [the YAML Path documentation](https://github.com/wwkimball/yamlpath/wiki/Python-Library#getting-data-from-the-document) which illustrates how to get data out of the document.  Leaving this flag at the default will cause the engine to attempt to generate any missing path ancestry, which is unknown where wildcards are used.

PR #474, opened 2023-01-11
[Bug Fix] Fix protected branches current configuration access level fetching

BenKuhar:
The `get_current_permissions` method in `branch_protector.py` does not properly account for returned `access_levels` when they are set to `0`.

This PR just adds a verification that the `access_level` value is present, otherwise the if check becomes `if 0:` when the 0 access level is set, which equates to False, resulting in unnecessary reapplications of branch protection settings.

### DIFF
--- a/gitlabform/processors/util/branch_protector.py
+++ b/gitlabform/processors/util/branch_protector.py
@@ -277,7 +277,7 @@ class BranchProtector:
         for array_element in protected_branches_response.get(
             f"{action}_access_levels", []
         ):
-            if array_element.get("access_level"):
+            if array_element.get("access_level") is not None:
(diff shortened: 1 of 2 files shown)


gdubicki:
Hey @BenKuhar! Thank you for your contribution! But could you please also add a test using your new code that will show that the bug is fixed?

BenKuhar:
> Hey @BenKuhar! Thank you for your contribution! But could you please also add a test using your new code that will show that the bug is fixed?

Added, let me know if any other changes are needed.

gdubicki:
Released in [v3.3.4](https://github.com/gitlabform/gitlabform/releases/tag/v3.3.4).

Issue #466, opened 2022-12-29
Can't manage members of top-level group


escalate:
**Describe the bug**

I would like to manage the members of our top-level groups.
e.g. /infrastructure
I'm not able to write the correct config.yml for this task.
I have tried several variants.

Variant 1: 
Here the user was only assigned to all subgroups of /infrastructure. However, not to the actual top-level group.
```
config_version: 3

projects_and_groups:
  "infrastructure/*":
    group_members:
      users:
        user1:
          access_level: owner
```

Variant 2:
An error was issued here for a subgroup of /infrastructure.
Error: The inheritance-break flag set in "infrastructure/subgroup" under key "users" is invalid
```
config_version: 3

projects_and_groups:
  "infrastructure/*":
    group_members:
      users:
        inherit: false
        user1:
          access_level: owner
```

Variant 3:
No error was output here. However, no change was made either.
```
config_version: 3

projects_and_groups:
  "infrastructure":
    group_members:
      users:
        user1:
          access_level: owner
```

Could you please help me to find the correct configuration?

**GitLabForm version**

3.4.0rc1

**GitLab version**

14.9.4-ee

gdubicki:
Hi @escalate! Thanks for the detailed question.

Your problem seems to be the opposite of what was asked for in https://github.com/gitlabform/gitlabform/issues/434#issuecomment-1361247072 - there a user wanted to manage some settings *only* for a group, but not it's subgroups, while you want to manage subgroups, but not the group.

Unfortunately this is not yet supported but if we come up with a backward-compatible way to implement it, the chances for getting it added relatively soon are higher.

How about we do this:
```
projects_and_groups:
  "infrastructure": # settings for the "infrastructure" group, but NOT its projects or subgroups
```
and
```
projects_and_groups:
  "infrastructure/**": # settings for the projects and the subgroups of "infrastructure" but NOT that group itself
```
...while keeping the existing:
```
projects_and_groups:
  "infrastructure/*": # settings for the "infrastructure" group AND its subgroups
```
...and...:
projects_and_groups:
  "infrastructure/project": # settings for the "infrastructure/project" project
```

Does this make sense to you? cc @amimas @jimisola

escalate:
Hello @gdubicki ,

thank you for your quick response.
I will wait for this feature.

Best regards
Felix

gdubicki:
Let’s keep this open until we implement it, @escalate :)

amimas:
[...]

PR #475, opened 2023-01-15
Stop treating usernames as case-sensitive

gdubicki:
...as in fact they are not. (You cannot have 2 users in a GitLab instance for which the username differs only in the case.)

Reported in #466


amimas:
The PR #556 has taken over this update. Let's close this PR for now as the other one will be merged soon.

PR #498, opened 2023-02-18
docs: update example in homepage to reflect latest config syntax

amimas:
Some minor update in the config samples in the docs so that they reflect the latest/current merge request approval syntax.


gdubicki:
Thanks!

Issue #499, opened 2023-02-18
Create schema for config syntax


amimas:
This is a great tool and it'd be good to have some kind of schema for the config syntax. Let's take a look at a simple config example:

```yaml
projects_and_groups:
  group_1/*:
  # Config for all projects under `group_1`
  project_settings:
    # Skipping details to keep the sample short
  branches:
    # Allow merging by developers, but no direct commits
    main:
      protected: true
      push_access_level: no access
      merge_access_level: developer
      unprotect_access_level: maintainer
  group_1/project_1:
    # Special setting for `project_1`
    project_settings:
      topics:
        - foo
        - bar
    branches:
      # Keep this branch unprotected
      develop:
        protected: false
      main:
        # special permission in this project for main branch
        allowed_to_push:
          - user1
          - user2
```

I personally made config mistakes several times. For example:

- `branches` config is further indented making it a property of `project_settings`
- `branches` was mis-spelled as `branchs`

GitLabForm does not report issues in mis-configurations mentioned above. In first case, I believe it gets passed to project settings api as additional parameter which gitlab simply ignores.

Obviously the config is not applied, but mistakes are not easy to see, especially when dealing with large config file containing lots of projects configurations, comments, etc. I spent a lot of time trying to debug reasons for config not being applied. I even opened issues in here only to later find out I made config syntax mistakes.

Ideally it'd be nice to have a linter that can catch mis-configurations in config file. The linter is probably a separate feature request though using tool like "[config-lint](https://github.com/stelligent/config-lint)".

Maybe the first step we can take is to create a [json schema](https://json-schema.org/) for the configuration syntax. Secondly, add the schema to [schemastore](https://www.schemastore.org/json/). Having the schema in this public location makes it available to everyone without taking manual steps because most of the popular IDEs or their extensions/plugins automatically consume it. This gives us auto-complete and syntax issues warnings in our editors.

Extra bonus would've been if gitlab api's schemas were also there. That way gitlabform's schema could refer to gitlab's schema within the objects (i.e. `project_settings`, `branches`, etc.), I think that would've made it whole, but having a config for gitlabform might be the start.

What do you think?

gdubicki:
Hey @amimas!

I am totally for such an improvement. Actually it was already proposed in https://github.com/gitlabform/gitlabform/issues/81#issuecomment-1215040100 but I missed that comment (sorry for that, @armingerten!). I don't have the time to work on this myself now, but I would be happy to help and review PRs.

[...]

PR #504, opened 2023-02-23
chore: add json-schema for configuration syntax

amimas:
This PR adds a [JSON Schema](https://json-schema.org/) for gitlabform configuration syntax as described in #499 . Few things to note about this schema:

The schema is manually created as opposed to generating from gitlabform build. I don't know if there's anything in-place to be able to auto-generate it. That may make it too complicated and hand-crafting the schema is probably good enough since gitlabform passes most of the configuration directly to gitlab api as parameter as opposed to writing special config syntax.

In this PR, only top-level syntax's schema has been added. For example:

```yaml
config_version: 3

gitlab:
  ssl_verify: true
  timeout: 10
  token: blah-23tlws623xwett
  url: gitlab.my-domain.com

skip_groups:
  - group1/subgroupA
  - group2/*

skip_projects:
  - group3/project1
```

Goal is to merge this PR with the above config's schema so that we have a starting point. That way others can also contribute or we can review changes in smaller chunks easily.

### DIFF
--- a/.github/workflows/linter-and-tests.yml
+++ b/.github/workflows/linter-and-tests.yml
@@ -128,6 +128,26 @@ jobs:
           attempt_limit: 5
           attempt_delay: 10000
 
+  schemas-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code from the PR
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 2
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
+      - name: Install test dependencies
+        run: |
+          pip install wheel
+          pip install -e .[test]
+      - name: Run schemas tests
+        run: |
+          pytest schemas/tests
+
   security-tests:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/pr-linter-and-tests.yml
+++ b/.github/workflows/pr-linter-and-tests.yml
@@ -166,6 +166,26 @@ jobs:
           attempt_limit: 5
           attempt_delay: 10000
 
+  schemas-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code from the PR
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 2
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
+      - name: Install test dependencies
+        run: |
+          pip install wheel
+          pip install -e .[test]
+      - name: Run schemas tests
+        run: |
+          pytest schemas/tests
+
   security-tests:
     runs-on: ubuntu-latest
     steps:
--- a/docs/contrib/schema_development.md
+++ b/docs/contrib/schema_development.md
@@ -1,0 +1,49 @@
+# Schema Development
+
+Schemas for GitLabForm's configuration syntax is written using [JSON Schema](https://json-schema.org/).
+
+The `schemas/src/` directory contains various independent schemas and they are used to compose
+`schemas/src/gitlabform.json`. Each individual schemas typically correspond to variuos sections of
+GitLabForm's configuration. Having individual schemas makes it easy to unit test them with different
+sample data as well as easier to maintain.
+
+Currently the goal of these schemas is to help users compose their GitLabForm configuration. The schemas
+will be available via [schemastore catalog](https://www.schemastore.org/json/), which is supported by various
+editors. Using these schemas, the editors can provide helpful hint/tooltip, auto completion, details/example of
+configuration, etc.
+
+## Required tools
+
+- Python 3 and Pip 3 for development
+
+## Environment setup
+
+1. Create virtualenv with Python 3, for example in `venv` dir which is in `.gitignore` and activate it:
+
+```
+python3 -m venv venv
+. venv/bin/activate
+```
+
+## Developing schemas
+
+Curently the schemas are handwritten. They are not auto generated. Follow the existing schemas pattern when
+developing a new schema. Following resources maybe helpful:
+
+- [Understanding JSON Schema](https://json-schema.org/understanding-json-schema/): A good reference for JSON Schema.
+- [python-jsonschema](https://python-jsonschema.readthedocs.io/en/stable/): The library used in unit test for validating schemas.
+- [referencing](https://referencing.readthedocs.io/en/stable/): The library used in unit test that provides reference resolution between multiple schemas.
+
+### Running unit tests locally
+
+To run unit tests locally:
+
+1. Activate the virtualenv created above
+
+2. Install the dependencies for tests:
+
+    ```
+    pip install -e '.[test]'
+    ```
+
+3. Run `pytest schemas/tests` to run all the unit tests.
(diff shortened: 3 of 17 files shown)


gdubicki:
Hey, this looks good @amimas and I am totally for the incremental approach.

Can we just add some code to actually test if the test configs that you added are matching the provided schema?

amimas:
Thanks @gdubicki. I was using that test file in ny editor to try out the schema. I can look into adding automated tests but I'm kind of waiting on the decision of where it will be hosted.

I left some comment in the linked issue. If we host the schema under gitlabform domain, we'll need to implement the necessary test framework. If it is hosted in the schemastore project, I believe the test framework is already setup and we'll just need to add test cases.

Let me know how we should proceed.

gdubicki:
I don’t have experience with JSON/YAML schemas, @amimas. Let’s do it the way you recommend, so it will be better for the users. So what would that be? 😊 

Of course I can help with creating the necessary deployment flow for the schema.

nejch:
@gdubicki @amimas  quick addition from me as I've done something similar at work.

You can use the `jsonschema` package (https://python-jsonschema.readthedocs.io/en/stable/) to validate some example config files and expect success/failures. An example might look like this, of course you'll turn this into test cases and it'll need some tweaks.

```python
from pathlib import Path

import jsonschema
import jsonschema.exceptions
import jsonschema.validators


# YOUR_SCHEMA: path to your schema
# YOUR_CONFIG: path to your config

with open(Path.cwd() / YOUR_CONFIG, "r") as f:
    config = yaml.safe_load(f)

resolver = jsonschema.validators.RefResolver(
    base_uri=f"{Path.cwd().as_uri()}/", referrer=True,
)
jsonschema.validate(
    instance=config, schema={"$ref": YOUR_SCHEMA}, resolver=resolver,
)
```

This also works really well with Red Hat's yaml extension for vscode by using the following in target files, so you could document that if you like.

```yaml
---
# yaml-language-server: $schema=https://raw.githubusercontent.com/gitlabform/gitlabform/main/schema/gitlabform-v3.json
# ... start gitlabform config
```

I've only tested it with local files as we have them in the same repo but it should work. See:
https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml#associating-schemas

amimas:
Hi @gdubicki . Sorry this PR have been stale. I got a little busy lately and didn't get time to continue. I'll try and pick it up again.

Thanks for your input @nejch. This is my first time developing a json schema for something. Based on your experience, would you recommend that we host the schema in this repo and also under https://gitlabform.github.io/gitlabform/ website?
I was thinking of hosting in in the https://github.com/SchemaStore/schemastore repo. One benefit comes to mind is that, I assume, that repo already has necessary basic validation and CI CD in place; we don't need to re-invent it here. What's your thought?

amimas:
Thanks for the suggestion @adam-moss . I was starting to lean towards hosting in-repo as well. Glad to see the gitlab ci schema example. I should've checked it earlier too. Now I just need to get back to this PR... :)

amimas:
Have been learning more about jsonschema, especially around automated schema validation using [python-jsonschema](https://python-jsonschema.readthedocs.io/en/stable/) library.

The PR is now ready for review and merge @gdubicki . But, unfortunately python 3.7 stands in the way now. The above library doesn't support 3.7. That's why the python 3.7 related tests are currently failing because the library cannot be installed.

Aside from that, here're the major changes I made from my initial setup:

Split into independent schema file for different portion of configs (i.e. `gitlab`, `skip_groups`, `skip_projects`, etc.). They are all then used as a reference into the main schema file (`schemas/src/gitlabform.json`) to compose the overall schema of gitlabform configuration syntax. I believe this allows a lot more flexibility for testing and validating the schema more rigorously. It should also help to maintain the schema easily because you won't be dealing with one giant json file.

The "ID" of each schemas are hardcoded to where they will be available as a raw file from github. For example: For `schemas/src/gitlabform.json` schema, the ID is set to `https://raw.githubusercontent.com/gitlabform/gitlabform/main/schemas/gitlabform.json`. Note that this doesn't mean unit testing or local development will be using that copy of the file. It's simply a "URI" that is resolved via the [referecing library](https://referencing.readthedocs.io/en/stable/). I used the above ID (and similar for other schemas) because that's what will be submitted to [schemastore catalog](https://www.schemastore.org/json/). We could use a different ID, but then we'll have to introduce extra build process into this project that modifies schema file with IDs that is a resolvable URL. We'll then have to figure out how to distribute that updated schema files (i.e. github release, gitlabform docs site, etc.). I took the easy path 😄 .

And finally, unit testing the schemas themselves. I hope the test files are self-explanatory. I did add a pytest plugin though. For example: added `pytest-dependency` plugin so that validating sample json/config with a schema is only done if the test for validating the actual schema itself is successful. And for validating schemas with sample json/configs, I used parametrized tests.

When you get a chance, please take a look. I hope this will set the foundation for others to help build rest of the schemas.

gdubicki:
I just went through all of the code in this PR and I am supper impressed. Thank you, @amimas, for this great contribution! 😊 

I would be happy to approve but I just have a little doubt about how will this handle versioning of gitlabform.

The schema at https://raw.githubusercontent.com/gitlabform/gitlabform/main/schemas/gitlabform.json will - in the best case scenario, if we release a new app version on every schema files updates - only match the latest app version, but users sometimes stick to the older versions for a while.

Wouldn't that be a problem?

amimas:
> I would be happy to approve but I just have a little doubt about how will this handle versioning of gitlabform.
> 
> The schema at https://raw.githubusercontent.com/gitlabform/gitlabform/main/schemas/gitlabform.json will - in the best case scenario, if we release a new app version on every schema files updates - only match the latest app version, but users sometimes stick to the older versions for a while.
> 
> Wouldn't that be a problem?

Hi @gdubicki . That's a good question. I think it depends on how this schema will be used.

My proposal is to provide this schema so that gitlabform users get auto-completion, hints, validations, etc. in their editors when composing the config file. For this, 2 things are needed:

- Make the schema files available to the users
- Configure IDE/editor setting to use that schema file with certain files

This is usually simplified by making the schemas available through [SchemaStore](https://www.schemastore.org/json/) because major IDEs and/or their yaml extensions natively support it. So, making this schema available into that schemastore will be the next step after this PR is merged.

We can create new version of schema for every release or every major version of gitlabform. But the issue is that schemas are associated with filenames. That's what allows the editors to give the above mentioned functionalities. So, when we make the schema published into the above SchemaStore catalog (or even if we manually add it to the editors), what filename will we reference? I'm thinking of something like `*.gitlabform.yml` or `*.gitlabform.yaml`. Now bringing in the versioning topic, we can specify version numbers to be in the file name (i.e. `*.gitlabform-v3.yml`). But, gitlabform doesn't require special file name as long as it's a yaml file (i.e. `config.yml`), right? If gitlabform expected config file name matching the exact version (i.e. `config.gitlabform-v3.7.0.yml`), I think it would not be convenient to upgrade as users will need to rename their config files during every upgrade.

So, I think it's okay that out schemas are always about the latest version. Let's look at [GitLab CI schema](https://gitlab.com/gitlab-org/gitlab/-/raw/master/app/assets/javascripts/editor/schema/ci.json) or [GitHub Action schema](https://json.schemastore.org/github-action.json). Users can be on different versions of their self-hosted GitLab or GitHub, but they'll see the latest hints/descriptions from the above schema.

We can still publish new schema matching the version of gitlabform. But, users will have to manually configure their editor's settings to consume those schemas. Not sure how many users will do that though 😄 .

We can also be a little more expressive in the schema. For example:

- In `description`, indicate what version of gitlabform they are available from
- use `deprecated` to indicate key have been deprecated

That was a long answer to your question. Hope that made sense 😅 . Let me know what you think and if the plan for next step (filenames to be associated with and adding to schemastore catalog) makes sense.

PR #511, opened 2023-03-05
Test

gdubicki:


nejch:
@gdubicki I think I see where you're going with the auto-approve ;) as an alternative, since it's only tests, here is what we do for python-gitlab if you're interested as well. It's a bit cheeky, but intended for tests only so should be ok.

https://github.com/python-gitlab/python-gitlab/blob/main/tests/functional/fixtures/create_license.rb

gdubicki:
Thanks @nejch, I’ll check it out. 

That may be a lifesaver as frankly I’ve had enough with the clunkyness of out current workaround…

nejch:
> Thanks @nejch, I’ll check it out.
> 
> That may be a lifesaver as frankly I’ve had enough with the clunkyness of our current workaround…

I've been considering packaging some of the python-gitlab test framework/fixtures into the sdist as a pytest plugin, but it would take a while, so for the time being, if you like this, might be easier to copy the approach :)

PR #515, opened 2023-03-08
Fix docs about venv and test dependencies setup

mfriedenhagen:
* python3 directly does include the `venv` module, so no need to firstly install virtualenv in your user or even site packages anymore.
* With quotes around `.[test]` at least the `zsh` of macOS does try to evaluate this itself.


gdubicki:
Thanks for your contribution @mfriedenhagen! 😊

Issue #509, opened 2023-02-28
group_settings are not applied to groups recursively


SachinKSingh28:
**Describe the bug**

It seems that the group_settings are not applied recursively to group and all its subgroup. The configuration I'm using is as shown below - 

```yaml
config_version: 3
gitlab:
  ssl_verify: true
  timeout: 20
projects_and_groups:
  "*":
    group_settings:
      share_with_group_lock: false
``` 

The command I'm running is -

`gitlabform -c config.yaml <namespace in gitlab.com>`

The command output is - 

```
🏗 GitLabForm version: 3.4.3 = the latest stable 😊
>>> Getting requested groups or projects...
:: # of groups to process: 1
:: # of projects to process: 0
:: (# of omitted projects - archived: xxxx, empty effective config: yyyy)
@ (1/1) Processing group: <namespace in gitlab.com>
:: # of groups processed successfully: 1
:: # of projects processed successfully: 0
```

Other command I have tried is as shown below - 

`gitlabform -c config.yaml ALL_DEFINED`

Output - 
```
🏗 GitLabForm version: 3.4.3 = the latest stable 😊
>>> Getting all groups and projects defined in the configuration...
Error: Configuration does not have any groups or projects defined!
```

I tried running with "ALL" as target too but that seems to be just taking ages and doesn't complete -

`gitlabform -c config.yaml ALL`

```
🏗 GitLabForm version: 3.4.3 = the latest stable 😊
>>> Getting ALL groups and projects...
```

I was expecting that all group and subgroups will have `share_with_group_lock` set to false but have been unsuccessful with all different configurations and CLI options.

**GitLabForm version**

GitLabForm version: 3.4.3 = the latest stable 😊

**GitLab version**

https://gitlab.com

amimas:
I use a self-hosted gitlab instance, so not too sure but looking at the config, I don't think should work.

To me, the below config is to apply group settings to all groups starting from the root group ( because of `*`).

```yaml
projects_and_groups:
  "*":
    group_settings:
      share_with_group_lock: false
```

I don't think you would have access to every top level group's settings in gitlab.com instance. Although, you ran `gitlabform -c config.yaml <namespace in gitlab.com>` command, maybe there's a bug in gitlabform.

Maybe update the config to refer to the group and then use `ALL_DEFINED` when running gitlabform. For example:

```yaml
projects_and_groups:
  "namespace-or-group/*":
    group_settings:
      share_with_group_lock: false
```

[...]

PR #518, opened 2023-03-13
Fix running with ALL when using gitlab.com

gdubicki:
and improve behavior in other cases when running as
a non-admin. In that case there is no point in trying
to get all groups or projects as you need at least
Reporter role to change their configuration.

Fixes: https://github.com/gitlabform/gitlabform/issues/509

### DIFF
--- a/gitlabform/__init__.py
+++ b/gitlabform/__init__.py
@@ -562,7 +562,9 @@ class GitLabForm:
         """
 
         if target == "ALL":
-            info(">>> Getting ALL groups and projects that I can modify...")
+            info(
+                ">>> Getting ALL groups and projects that I have permission to modify..."
+            )
         elif target == "ALL_DEFINED":
             info(">>> Getting ALL groups and projects DEFINED in the configuration...")
         else:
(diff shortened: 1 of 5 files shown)


gdubicki:
I would appreciate it if someone could take a look at this, conceptually. Do you think it is okay to limit getting groups and projects like this, @amimas @jimisola?

gitlabform/processors/__init__.py
```diff
@@ -28,21 +28,10 @@ def process_entity(
     ):
         for processor in self.processors:
             if only_sections == "all" or processor.configuration_name in only_sections:
-                try:
```
gdubicki, line 31: This part of the PR just restores the code before by debugging that accidentally got pushed to `main`. It's not related with the rest of the PR.

gitlabform/__init__.py
```diff
@@ -562,9 +562,9 @@ def _show_header(
         """
 
         if target == "ALL":
-            info(">>> Getting ALL groups and projects...")
+            info(">>> Getting ALL groups and projects that I can modify...")
```
jimisola, line 565: "Getting ALL groups and projects that I have permission to modify..."

gdubicki: Thanks, corrected!

Issue #263, opened 2021-09-02
Support multiple Merge Request approval rules


Mahpara93:
Is there a way we could have multiple approval rules added using gitlabform? https://docs.gitlab.com/ee/user/project/merge_requests/approvals/rules.html#add-multiple-approval-rules

Thanks,

Mahpara

Mahpara93:
Looks like something that's already in the works https://github.com/egnyte/gitlabform/pull/176. Closing this issue.

gdubicki:
Nvm, sorry. Please see #388 for further work on this.

PR #525, opened 2023-03-31
upgrade to v3: mention MR approvals syntax change

weakcamel:
Related to #388 and #263

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -35,7 +35,9 @@ projects_and_groups:
     * Replace `services:` with `integrations:`,
     * Replace `secret_variables:` with `variables:`,
     * Replace `group_secret_variables:` with `group_variables:`,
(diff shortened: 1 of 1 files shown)


weakcamel:
After a recent Gitlab instance upgrade I finally got around to upgrading Gitlabform too. 

While porting the configs I've noticed that the current version doesn't like the `approvals_before_merge` outside of `merge_requests_approval_rules` (see below) so I'm proposing to mention this in the upgrade guide.


```
Processing section 'merge_requests_approvals'
Error: Setting the 'approvals_before_merge' in the 'merge_requests_approvals' sections is not allowed as it is not clear which rule does it apply to. Please set it inside the specific approval rules in the 'merge_requests_approval_rules' section.
```

gdubicki:
Sure, why not. Thanks for the contribution, @weakcamel!

Issue #555, opened 2023-06-22
Usernames in projects and groups should be case-insensitive


J4Numbers:
# Background

Currently, a username in GitLab will resolve down to an ID at some level. That username exists as a namespace in front of the ID number and is case-insensitive when querying the [GitLab list users API](https://docs.gitlab.com/ee/api/users.html#list-users) about it.

## Existing behaviour

(As seen on `3.6.0`)

Given that `Joe.Bloggs` already exists as a member of a group,

When running a configuration which includes a username of `joe.bloggs` against that group,

Then the GitLabForm attempts to add `joe.bloggs` to that group,

And the GitLabForm fails, stating that the ID number of `Joe.Bloggs` already exists against that group.

## Proposed behaviour

Given that `Joe.Bloggs` already exists as a member of a group,

When running a configuration which includes a username of `joe.bloggs` against that group,

Then the GitLabForm does not attempt to add the existing member to that group.

gdubicki:
Hi, @J4Numbers, thanks for reporting the issue.

It's a known one though, discovered in https://github.com/gitlabform/gitlabform/issues/466 and some time ago I started working on a fix but I only found the time to implement tests that show that for now the names are treated as case-sensitive: https://github.com/gitlabform/gitlabform/pull/475

Sadly I still have little time so if you or someone else needs this badly, you can feel free to take over my PR (or start from the scratch, if you prefer) and implement it. Some day I might complete it myself, but I cannot tell you when, sorry.

amimas:
This is now available in v3.8.0: https://github.com/gitlabform/gitlabform/releases/tag/v3.8.0

PR #556, opened 2023-06-26
fix(members): ensure group and project users are case-insensitive

J4Numbers:
Meddle with group and project membership checks to ensure that
usernames are case-insensitive. This change ensures that users
are not added again by mistake, causing an error where a given
user already exists within the repository.

### DIFF
--- a/tests/acceptance/__init__.py
+++ b/tests/acceptance/__init__.py
@@ -243,6 +243,10 @@ def get_only_tag_access_levels(project: Project, tag):
     )
 
 
+def randomize_case(input: str) -> str:
+    return "".join(random.choice((str.upper, str.lower))(char) for char in input)
+
+
 def run_gitlabform(config, target, include_archived_projects=True):
     # f-strings with """ used as configs have the disadvantage of having indentation in them - let's remove it here
     config = textwrap.dedent(config)
--- a/tests/acceptance/premium/test_branches_users_case_insensitive.py
+++ b/tests/acceptance/premium/test_branches_users_case_insensitive.py
@@ -44,7 +44,7 @@ class TestBranchesUsersCaseInsensitive:
             push_access_user_ids,
             merge_access_user_ids,
             _,
-        ) = project.protectedbranches.get(branch)
+        ) = get_only_branch_access_levels(project, branch)
 
         assert push_access_levels == [AccessLevel.MAINTAINER.value]
         assert merge_access_levels == [AccessLevel.MAINTAINER.value]
--- a/tests/acceptance/premium/test_branches_users_case_insensitive.py
+++ b/tests/acceptance/premium/test_branches_users_case_insensitive.py
@@ -1,10 +1,10 @@
 import pytest
 
-from gitlabform.gitlab.core import NotFoundException
 from gitlabform.gitlab import AccessLevel
 from tests.acceptance import (
     run_gitlabform,
     randomize_case,
+    get_only_branch_access_levels,
 )
 
 pytestmark = pytest.mark.requires_license
--- a/tests/acceptance/premium/test_branches_users_case_insensitive.py
+++ b/tests/acceptance/premium/test_branches_users_case_insensitive.py
@@ -1,0 +1,57 @@
+import pytest
+
+from gitlabform.gitlab.core import NotFoundException
+from gitlabform.gitlab import AccessLevel
+from tests.acceptance import run_gitlabform, gl, randomize_case
+
+
+class TestBranchesUsersCaseInsensitive:
+    @pytest.mark.skipif(
+        gl.has_no_license(), reason="this test requires a GitLab license (Paid/Trial)"
+    )
+    def test__users_case_insensitive(
+        self,
+        gitlab,
+        group_and_project,
+        branch,
+        make_user,
+    ):
+        first_user = make_user(AccessLevel.DEVELOPER)
+        second_user = make_user(AccessLevel.DEVELOPER)
+        third_user = make_user(AccessLevel.DEVELOPER)
+
+        config_with_more_user_ids = f"""
+        projects_and_groups:
+          {group_and_project}:
+            branches:
+              {branch}:
+                protected: true
+                allowed_to_push:
+                  - access_level: {AccessLevel.MAINTAINER.value} 
+                  - user: {randomize_case(first_user.name)}
+                  - user: {randomize_case(second_user.name)}
+                  - user: {randomize_case(third_user.name)}
+                allowed_to_merge:
+                  - access_level: {AccessLevel.MAINTAINER.value}
+        """
+
+        run_gitlabform(config_with_more_user_ids, group_and_project)
+
+        (
+            push_access_levels,
+            merge_access_levels,
+            push_access_user_ids,
+            merge_access_user_ids,
+            _,
+        ) = gitlab.get_only_branch_access_levels(group_and_project, branch)
+
+        assert push_access_levels == [AccessLevel.MAINTAINER.value]
+        assert merge_access_levels == [AccessLevel.MAINTAINER.value]
+        assert push_access_user_ids == sorted(
+            [
+                first_user.id,
+                second_user.id,
+                third_user.id,
+            ]
+        )
+        assert merge_access_user_ids == []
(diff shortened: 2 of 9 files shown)


J4Numbers:
Addresses #555 and continues the work from #475.

gdubicki:
Hey @J4Numbers, thank you for your contribution! And sorry for the CI problem, I noticed that there are HTTP 401 thrown. That's strange because it works for me locally. :/ I will try to troubleshoot it over the next few days.

gdubicki:
Ok, now it seems to be failing on something related to your code, @J4Numbers:
```
FAILED tests/acceptance/premium/test_branches_users_case_insensitive.py::TestBranchesUsersCaseInsensitive::test__users_case_insensitive - TypeError: cannot unpack non-iterable ProjectProtectedBranch object
```
Can you check it out please?

J4Numbers:
Thanks for kicking it off! Full error is below:

```
FAILED tests/acceptance/standard/test_members_enforce.py::TestMembersEnforce::test__enforce_keep_bots - assert 5 == 2
 +  where 5 = len([<ProjectMember id:37 username:gitlabform_user_maroon_4>, <ProjectMember id:40 username:gitlabform_user_maroon_5>, <Pr...:45 username:project_25_bot_016e6e53920e5925458ecb37e324d4a0>, <ProjectMember id:46 username:gitlabform_user_maroon_7>])
```

gdubicki:
> Thanks for kicking it off! Full error is below:
> 
> ```
> FAILED tests/acceptance/standard/test_members_enforce.py::TestMembersEnforce::test__enforce_keep_bots - assert 5 == 2
>  +  where 5 = len([<ProjectMember id:37 username:gitlabform_user_maroon_4>, <ProjectMember id:40 username:gitlabform_user_maroon_5>, <Pr...:45 username:project_25_bot_016e6e53920e5925458ecb37e324d4a0>, <ProjectMember id:46 username:gitlabform_user_maroon_7>])
> ```
> 
> This might be a bit of trial and error since I can't test the premium items locally...

This is fixed now.

But

J4Numbers:
Thanks for picking up the missed import!

> This is fixed now.
> 
> But

Guessing something got cut off here. But it does look like there's a race condition or some test overlap in this somewhere (with the odd failure between different python versions). Want me to raise an issue for it?

gdubicki:
> > This is fixed now.
> > But
> 
> Guessing something got cut off here. But it does look like there's a race condition or some test overlap in this somewhere (with the odd failure between different python versions). Want me to raise an issue for it?

This may be caused by the fact that many tests depend on each other, see https://github.com/gitlabform/gitlabform/pull/561#pullrequestreview-1528759477 for more info.

J4Numbers:
Fair enough. Let me know if you need me to do anything else!

tests/acceptance/premium/test_branches_users_case_insensitive.py
```diff
@@ -0,0 +1,58 @@
+import pytest
+
+from gitlabform.gitlab import AccessLevel
+from tests.acceptance import (
+    run_gitlabform,
+    randomize_case,
+    get_only_branch_access_levels,
+)
+
+pytestmark = pytest.mark.requires_license
```
amimas, line 10: What's the purpose of this test file (`tests/acceptance/premium/test_branches_users_case_insensitivity.py`)? Please correct me if I'm wrong here. I don't think username case-sensitivity is a gitlab premium/ultimate feature.

amimas (changes requested):
Unsure why there's a gitlab premium related test case. I think it doesn't belong there.

tests/acceptance/premium/test_branches_users_case_insensitive.py
J4Numbers: Added in as an additional test for the `tests/acceptance/premium/test_branches.py` file, which also exists under premium features. Protected branches have _some_ premium protections on them - this does affect the permissions for `allowed_to_merge` and `allowed_to_push` as they both do username/group name lookups.

J4Numbers: Tried moving the branch case test to standard and removing the `requires_license` fixture, and it failed due to differences with the enterprise API

amimas: > Protected branches have some premium protections on them - this does affect the permissions for allowed_to_merge and allowed_to_push as they both do username/group name lookups.

Ahh... that makes sense now. I didn't think of that. Thanks for clarifying it.

amimas:
@J4Numbers - Would you be able to update your PR branch please? Another change was merged recently and looks like now there's a bit of conflict that will need to be resolved manually.

J4Numbers:
`types` is now failing for some reason, despite me not having touched it and the relevant lines last being updated in May...

amimas:
> `types` is now failing for some reason, despite me not having touched it and the relevant lines last being updated in May...

This is coming from the last PR merge. It should be fixed in #570, I think.

amimas:
@gdubicki - Should we merge this?

amimas:
@J4Numbers - Pushed the latest update to your branch. Noticed a test failure from one of your test case. 

https://github.com/gitlabform/gitlabform/actions/runs/5767054764/job/15636113972?pr=556

Is this an intermittent test failure? If yes, would you be able to look into the test setup? It'd be good to have it fixed if possible as intermittent failures take away too much time in processing other PRs.

Hopefully we can merge this PR soon.

amimas:
@J4Numbers - Did you get a chance to look at the test failure mentioned above?

J4Numbers:
Ah, sorry. I started looking into it, but then real life got in the way slightly. From what I can tell, a lot of the code _should_ be pretty segregated due to the random group creation that the tests are separated between (something similar to `group_1234`) which then has its own projects and users - all of which follow that format of `gitlabform_{basename}_{random_xkcd_string}` which makes unpicking _why_ the tests are interdependent... fun.

I can keep looking into it if required, but the base logic for generating random groups/projects/users/etc. seems sound, which makes me suspect that it might be something to do with the fixtures holding state when they shouldn't...

J4Numbers:
So, as far as I can tell, separating the two project member tests into their own fixtures seems to have resolved my issue. And this run seems to have rolled a lucky six so far in terms of pipelines behaving...

I did notice one other intermittent failure, which was regarding a branch protection test. (See: https://github.com/gitlabform/gitlabform/actions/runs/6031288955/job/16364708681)

amimas:
> I started looking into it, but then real life got in the way slightly.

Hey @J4Numbers . Completely understandable. Hope all is well. Thanks for continuing with this PR. Really appreciate it 🙏 .

amimas:
> makes me suspect that it might be something to do with the fixtures holding state when they shouldn't...

> So, as far as I can tell, separating the two project member tests into their own fixtures seems to have resolved my issue. And this run seems to have rolled a lucky six so far in terms of pipelines behaving...

First of all, thanks for making it more reliable. I'm not an expert at this but think you're on the right path. Looks like you've split your original tests into 2 separate test class (i.e. `tests/acceptance/standard/test_project_members_case_insensitve.py` and tests/acceptance/standard/test_project_group_members_case_insensitive.py`).

I believe what was happening before is that the tests were in the same class. Both tests were using the `project` fixture whose scope is at the class level. So, when the first test's membership is updated and verified, the 2nd test's evaluation will need to keep in mind the result from the earlier test, unless that one manually did the cleanup of members so that the project is at it's initial state. Why this happens intermittently, I'm not sure. Another [similar intermittent issue](https://github.com/gitlabform/gitlabform/pull/585) was addressed by switching to `project_for_function` fixture in the test cases instead of using the `project` fixture. This has the same effect as splitting the tests into separate classes as you've done.

It would probably be nice if those tests were in the same class. Reduce some duplication and logically grouped; might be easier to maintain.

@gdubicki - Would you be able to do another review, please?

amimas:
> I did notice one other intermittent failure, which was regarding a branch protection test. (See: https://github.com/gitlabform/gitlabform/actions/runs/6031288955/job/16364708681)

Yeah, I noticed it a few times too. It's annoying 😄 .

tests/acceptance/__init__.py
```diff
@@ -65,7 +65,9 @@ def allowed_codes(codes: Union[int, List[int]]):
 
 def get_random_name(entity: str) -> str:
     random_suffix = get_random_suffix()
-    return f"gitlabform_{entity}_{random_suffix}"
+    built_random = f"gitlabform_{entity}_{random_suffix}"
+    print(f"Provisioned random name '{built_random}'")
+    return built_random
```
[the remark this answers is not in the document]
J4Numbers: Reverted

gdubicki (approved):
LGTM, thank you for the contribution @J4Numbers!

amimas (approved):

PR #554, opened 2023-06-14
fix: ignore keep_bots from users list for group membership

SachinKSingh28:
Fixes #553

PR #557, opened 2023-07-11
chore(dev/ci): set expiry date for access token

amimas:
As of GitLab v16.0, all access tokens must have an expiry date. Otherwise the access token will not be created. This wasn't being done in the script that sets up a GitLab instances with an access token so that the token can be used for all acceptance tests via API calls. This worked fine before but started failing recently when latest version of GitLab was being used in the CI because the latest version is 16 and up. As a result, all the PR validations are failing because the acceptance tests are failing with '401 Unauthorized' error.

### DIFF
--- a/dev/healthcheck-and-setup.sh
+++ b/dev/healthcheck-and-setup.sh
@@ -30,7 +30,8 @@ test -f $done || {
     printf 'gitlabform_token = PersonalAccessToken.create('
     printf 'user_id: 1, '
     printf 'scopes: [:api, :read_user], '
-    printf 'name: :gitlabform);'
+    printf 'name: :gitlabform, '
+    printf 'expires_at: 365.days.from_now);'
     printf "gitlabform_token.set_token('token-string-here123');"
     printf 'gitlabform_token.save!;'
   ) | gitlab-rails console


amimas:
Hi @gdubicki 👋, 

I'm not sure why codecov is reporting a big decrease in test coverage. This PR makes no changes to the codes that it's reporting a coverage decrease. Not too familiar with codecov, but I think this shouldn't be a blocker for this PR.

Also, because of GitLab v16 requirements of bot users must have an expiry date, couple of existing acceptance tests are failing in this PR.

```
=========================== short test summary info ============================
ERROR tests/acceptance/standard/test_group_members_users.py::TestGroupMembersUsers::test__remove_user_keep_bots - gitlab.exceptions.GitlabCreateError: 400: expires_at is missing
ERROR tests/acceptance/standard/test_members_enforce.py::TestMembersEnforce::test__enforce_keep_bots - gitlab.exceptions.GitlabCreateError: 400: expires_at is missing
============== 102 passed, 2 errors, 6 rerun in 223.73s (0:03:43) ==============
```

I believe this should be fixed as part of #554.

So, I'm leaning towards merging this PR soon even with the reported failures. It will at least resolve the issue with acceptance tests not running in the CI and unblock #554 . Thoughts?

gdubicki (approved):
BIG thank you for resolving this, @amimas ! I agree, let’s merge this first and deal with other failing tests later.

PR #561, opened 2023-07-13
feat: add enforce support for pipeline schedule

### LINKED ISSUE #539: Add `enforce` support to schedule pipeline config
Currently the scheduled pipeline configuration syntax is like this:

```yaml
   group-name/project-name:
    schedules:
      Title of the schedule:
        ref: <branch-name>
        cron: <cron-expression>
        cron_timezone: <time-zone>
        variables:
          PIPELINE_VARIABLE1:
            value: <variable-value>
```

Consider the following 2 cases:

- Rename the "Title of the schedule" to something else
- Delete the entire schedule

In the first case, a new schedule will be created; existing schedule will remain in the project. For the 2nd case, nothing happens. The schedule from previous run of `gitlabform` will remain in the project.

In most cases the above is probably not expected. It'd be good to have `enforce: true` support for the `schedules` config so that any "unconfigured schedules" are deleted.

amimas:
This change adds `enforce` support to pipeline schedule configuration. By adding `enforce: true`, unconfigured schedules can be deleted. This can help keep the gitlabform config and project's pipeline schedule's in sync.

closes #539

### DIFF
--- a/docs/reference/pipeline_schedules.md
+++ b/docs/reference/pipeline_schedules.md
@@ -17,10 +17,6 @@ There are 2 gitlabform specific keys/configs that can be set under `schedules` o
 - `delete` - set this to `true` under a specific schedule to delete that particular schedule.
 - `enforce` - set this to `true` under `schedules` so that any schedules that are not in `schedules` section are deleted.
 
-    !!! warning
-
-        The `delete` key maybe removed in future version of gitlabform as `enforce` key can be used for deleteing unconfigured schedules and keep gitlabform config cleaner.
-
 
 Example 1:
 ```yaml
--- a/docs/reference/pipeline_schedules.md
+++ b/docs/reference/pipeline_schedules.md
@@ -12,7 +12,17 @@ Additionally, under a `variables` key you can add the pipeline schedule variable
 
     * Do not set `description` attribute - see [#535](https://github.com/gitlabform/gitlabform/issues/535#issue-1678509984)
 
-Example:
+There are 2 gitlabform specific keys/configs that can be set under `schedules` or individual schedule:
+
+- `delete` - set this to `true` under a specific schedule to delete that particular schedule.
+- `enforce` - set this to `true` under `schedules` so that any schedules that are not in `schedules` section are deleted.
+
+    !!! warning
+
+        The `delete` key maybe removed in future version of gitlabform as `enforce` key can be used for deleteing unconfigured schedules and keep gitlabform config cleaner.
+
+
+Example 1:
 ```yaml
 projects_and_groups:
   group_1/project_1:
@@ -32,5 +42,18 @@ projects_and_groups:
           other_variable:
             value: another_value
       "Obsolete schedule":
-        delete: true
+        delete: true  # Delete this schedule
+```
+
+Example 2:
+```yaml
+projects_and_groups:
+  group_1/project_1:
+    schedules:
+      enforce: true  # Delete all other pipeline schedules that exists for this project
+      "Some schedule":
+        ref: main
+        cron: "0 * * * MON-FRI"
+        cron_timezone: "London"
+        active: false
 ```
--- a/tests/acceptance/standard/test_schedules.py
+++ b/tests/acceptance/standard/test_schedules.py
@@ -199,7 +199,7 @@ class TestSchedules:
         )
         assert schedule is None
 
-    def test__schedule_enforce(self, project, schedules):
+    def test__schedule_enforce_new_schedule(self, project, schedules):
         new_schedule = f"""
         projects_and_groups:
           {project.path_with_namespace}:
@@ -210,11 +210,17 @@ class TestSchedules:
                 cron: "30 1 * * *"
         """
 
+        schedules_before = project.pipelineschedules.list()
+
         run_gitlabform(new_schedule, project)
+
+        schedules_after = project.pipelineschedules.list()
 
         schedule = self.__find_pipeline_schedule_by_description_and_get_first(
             project, "One and only schedule"
         )
+        assert len(schedules_before) == 6
+        assert len(schedules_after) == 1
         assert schedule is not None
         assert schedule.description == "One and only schedule"
         assert schedule.ref == "main"
--- a/tests/acceptance/standard/test_schedules.py
+++ b/tests/acceptance/standard/test_schedules.py
@@ -199,6 +199,29 @@ class TestSchedules:
         )
         assert schedule is None
 
+    def test__schedule_enforce(self, project, schedules):
+        new_schedule = f"""
+        projects_and_groups:
+          {project.path_with_namespace}:
+            schedules:
+              enforce: true
+              "One and only schedule":
+                ref: main
+                cron: "30 1 * * *"
+        """
+
+        run_gitlabform(new_schedule, project)
+
+        schedule = self.__find_pipeline_schedule_by_description_and_get_first(
+            project, "One and only schedule"
+        )
+        assert schedule is not None
+        assert schedule.description == "One and only schedule"
+        assert schedule.ref == "main"
+        assert schedule.cron == "30 1 * * *"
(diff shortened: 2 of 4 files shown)


tests/acceptance/standard/test_schedules.py
```diff
@@ -199,6 +199,95 @@ def test__delete_schedule(self, project, schedules):
...
+        schedules_after = project.pipelineschedules.list()
+
+        schedule1 = self.__find_pipeline_schedule_by_description_and_get_first(
+            project, "New schedule to test enforce config"
+        )
+        schedule2 = self.__find_pipeline_schedule_by_description_and_get_first(
+            project, "New schedule with variables"
+        )
+        assert len(schedules_before) == 1
+        assert len(schedules_after) == 2
```
amimas, line 264: @gdubicki - It seems the tests are currently setup to be dependent on earlier test case's results. That's why this assertion looks odd to me. I expected each test case to be self-contained and not affecting other tests. I'm not sure if this was done intentionally. Thought I'd mention.

amimas:
Just realized that type checking is passing in the CI but it's failing on my local machine.

```
(venv) ➜  gitlabform git:(issue-539) mypy . || true
gitlabform/processors/project/schedules_processor.py:18: error: Item "None" of "Optional[Any]" has no attribute "pop"  [union-attr]
gitlabform/processors/project/schedules_processor.py:25: error: Argument 1 to "sorted" has incompatible type "Optional[Any]"; expected "Iterable[Any]"  [arg-type]
gitlabform/processors/project/schedules_processor.py:28: error: Value of type "Optional[Any]" is not indexable  [index]
gitlabform/processors/project/schedules_processor.py:85: error: Unsupported right operand type for in ("Optional[Any]")  [operator]
Found 4 errors in 1 file (checked 132 source files)
```

tests/acceptance/standard/test_schedules.py
gdubicki: This is a result of the fact that we are often using fixtures that have a pytest scope higher than "function". Particularly, [the fixture for a project has scope "class"](https://github.com/gitlabform/gitlabform/blob/main/tests/acceptance/conftest.py#L102-L103), so the schedules created in one method of a test class such as TestSchedules will in fact be retained in the next test method.

To fix that we would have to use "function" scope for an almost perfect separation of each test but we are are talking about tests on a real-world GitLab instance where creating and deleting projects in GitLab is pretty slow.

I just did a test of how this would work out in practice for `test_schedules.py`:

Code as-is, with project and group having scope "class":

1st run - 7.708s
2nd run - 6.280s
3rd run - 7.390s

Code with project and group scope changed to "function":

1st run - 14.556s
2nd run - 18.286s
3rd run - 18.335s

So the tests time effectively increases 2-2.5 times.

The total current time of running tests in GitHub Actions is ~3 minutes for standard and ~1.5 minutes for premium ones, so including the ~4 minutes of GitLab provisioning time the whole pipeline time would increase from ~7 minutes to ~10 minutes...

Because of that we are now using 2 ways of dealing with the issue of tests affecting each other, only where it affects the tests:

a) use "function" scoped fixtures on specific single tests - see the uses of `project_for_function` fixture,
b) make the assertions about the initial state more relaxed - see https://github.com/gitlabform/gitlabform/blob/main/tests/acceptance/standard/test_members.py#L12-L12 as an example.

Of course this is not perfect, but it's the best compromise between correctness and performance of the tests that I was able to develop at the time. If you think that we should make changes then feel free to share your ideas, @amimas. :)

gdubicki:
> Just realized that type checking is passing in the CI but it's failing on my local machine. (...)

I am having the same issue. :/ It's not the first strange behaviour of mypy that I have observed but frankly I don't feel that this mypy failures are serious enough and at the same time it was difficult to troubleshoot so I gave up. But if you want to try to get to the bottom of this, feel free! :)

docs/reference/pipeline_schedules.md
```diff
@@ -12,7 +12,17 @@ Additionally, under a `variables` key you can add the pipeline schedule variable
...
 
-Example:
+There are 2 gitlabform specific keys/configs that can be set under `schedules` or individual schedule:
+
+- `delete` - set this to `true` under a specific schedule to delete that particular schedule.
+- `enforce` - set this to `true` under `schedules` so that any schedules that are not in `schedules` section are deleted.
+
+    !!! warning
+
+        The `delete` key maybe removed in future version of gitlabform as `enforce` key can be used for deleteing unconfigured schedules and keep gitlabform config cleaner.
```
gdubicki, line 22: I am not sure if we should declare this. I don't remember, have we decided that we want to deprecate `delete`? I personally rather like it, it's like Puppet's `ensure: absent` pattern.

tests/acceptance/standard/test_schedules.py
amimas: Thanks @gdubicki for the details and the performance tests. I also realized this came down to fixture scope after looking into it a bit, but didn't have time to learn what changes are needed if the scope is changed.

I'm starting to think what you've got right now actually makes sense. These are not mocked unit tests. As you said, we're running these against a real GitLab instance. Also not sure if there's a real need for running only individual test as opposed to running all tests in a single file. That would probably make the testing and development faster, but is it worth it? So, it is a balance between isolated tests vs performance of entire test suites.

I don't have any other ideas at the moment. Being new to the code base, it was a bit confusing at the beginning. I had to go through existing test cases and then figure out if the assertions in the new test case makes sense. Maybe what might help others is if we have additional sections in the contributing docs about general setup/expectations of acceptance tests. I have only played with 1 test file through this PR so far. If I can think of any improvements to the docs/development guide in future, I'll open new PR accordingly.

amimas:
> > Just realized that type checking is passing in the CI but it's failing on my local machine. (...)
> 
> I am having the same issue. :/ It's not the first strange behaviour of mypy that I have observed but frankly to me mypy failures are not serious enough and at the same time it was difficult to troubleshoot so I gave up. But if you want to try to get to the bottom of this, feel free! :)

If you think mypy issues are not serious enough, that's fine. I'm new to this as well :). What bugs me though is the behaviour difference between my local machine vs CI. I might look into that a bit, unless you already know possible reasons.

docs/reference/pipeline_schedules.md
amimas: Ahh... You're right! We didn't discuss about deprecating `delete`. I guess I just made an assumption because I didn't see a need for it anymore. This warning was kind of a way to start the conversation 😛 .

We can keep `delete` alongside `enforce` in the future, if you think it's going to be useful.

amimas:
> > > Just realized that type checking is passing in the CI but it's failing on my local machine. (...)
> > 
> > 
> > I am having the same issue. :/ It's not the first strange behaviour of mypy that I have observed but frankly to me mypy failures are not serious enough and at the same time it was difficult to troubleshoot so I gave up. But if you want to try to get to the bottom of this, feel free! :)
> 
> If you think mypy issues are not serious enough, that's fine. I'm new to this as well :). What bugs me though is the behaviour difference between my local machine vs CI. I might look into that a bit, unless you already know possible reasons.

I couldn't figure out the discrepancy between CI and my local machine. But, I've made a small tweak to the code and `mypy` wasn't reporting issues anymore.

docs/reference/pipeline_schedules.md
amimas: I've removed this warning from the doc in latest commit.

gdubicki (approved):
LGTM now!

docs/reference/pipeline_schedules.md
gdubicki: 🙏 ☺️

tests/acceptance/standard/test_schedules.py
gdubicki: I agree, we should at least document it. I'll try to do it soon.

PR #551, opened 2023-06-04
feat: add allowed_to_create support for tag protection

amimas:
Currently `gitlabform` does not support [configuring tag protection](https://docs.gitlab.com/ee/user/project/protected_tags.html#configuring-protected-tags) using specific user(s) or group(s) as "Allowed to create". Only basic role can be set, such as: developer, maintainer, etc. Selecting individual user or group is a feature of Premium or Ultimate license.

This change adds the ability to configure tag protection to be set to individual users or groups. They can be set as a user/group ID or their names. For example with a config like below, a tag named `special-tag-blah` can only be created by either user `Jane.Doe`, or user whose ID is `123` or, users who are member of the group `foo-bar`.

```yaml
projects_and_groups:
  group_1/project_1:
    tags:
      "v*":
        protected: true
        create_access_level: developer
      "special-tag-*":
        protected: true
        allowed_to_create:
          - user: Jane.Doe
          - user_id: 123
          - group: foo-bar
      "some-old-tag":
        protected: false
```

To Dos:

- [x] Implement feature
- [x] Add acceptance test
- [x] Update documentation

closes #505

PR #570, opened 2023-07-21
chore: fix incompatible type in test fixture

amimas:
This PR is related to #551 . In that PR the GitHub CI was fine, but after merging the change found out "type" issue in test fixture reported by `mypy`. See https://github.com/gitlabform/gitlabform/actions/runs/5617292516/job/15221155542

Not sure why the issue wasn't originally caught in the PR's CI but I had also missed running the `mypy` locally to validate it.

This PR addresses the type issue so that we have a clean codebase.

### DIFF
--- a/.github/workflows/pr-linter-and-tests.yml
+++ b/.github/workflows/pr-linter-and-tests.yml
@@ -30,7 +30,11 @@ jobs:
   types:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout code from the PR
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 2
       - name: Set up Python 3.11
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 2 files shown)


amimas:
The types validation in this PR is now failing with the same error mentioned in the description:

```
conftest.py:170: error: Incompatible types in "yield" (actual type "Callable[[Any, Any], Any]", expected type "Optional[Group]")  [misc]
```

Not sure why it's reporting the same error. It's as if the checks are running on `main` branch code instead of this PR. That error doesn't exist anymore in this PR.

amimas:
> The types validation in this PR is now failing with the same error mentioned in the description:
> 
> ```
> conftest.py:170: error: Incompatible types in "yield" (actual type "Callable[[Any, Any], Any]", expected type "Optional[Group]")  [misc]
> ```
> 
> Not sure why it's reporting the same error. It's as if the checks are running on `main` branch code instead of this PR. That error doesn't exist anymore in this PR.

After going through the GitHub Actions workflow config, I realized that's exactly what was happening. For the `types` step, the PR commit wasn't being checked out.

I updated the PR workflow config, but it's not being reflected in this PR's workflow run. Is it because the workflow config is only taken from the `main` branch?

gdubicki:
> I updated the PR workflow config, but it's not being reflected in this PR's workflow run. Is it because the workflow config is only taken from the `main` branch?

Yes, exactly.

Issue #73, opened 2020-01-01
Consider migrating to `python-gitlab`


gdubicki:
When I started working on Gitlabform (as an internal Egnyte app back then) in the beginning of 2017, one existing GitLab python library I have found had a README that stated something like "Screw it, I am not maintaining this library any longer - GitLab makes too many API changes between even patch versions to keep up with it!"... 🙄 

I also remember that some other library I considered was hiding too many GitLab API internals so it was not compatible with the design decision I made to forward values from them YAML config verbatim as parameters for PUTs to GitLab API endpoints (to not have to update Gitlabform app in case of those dreaded API changes that I was warned about above).

Therefore I have started to just use `requests` and wrote my own library.

I don't remember if I have even found the [python-gitlab/python-gitlab](https://github.com/python-gitlab/python-gitlab) and why I have not considered using it...

But now as it has gained friction (1.1k stars as of now) it would be good to stop reinventing the wheel and switch to it. If it lacks something that I need, then I should write a PR to add it to it.

For now I see that is DOES provide both creating new entities by providing raw JSON of parameters as well a direct access to parameters by the same names as in the GitLab API (see examples in https://python-gitlab.readthedocs.io/en/latest/gl_objects/projects.html vs https://docs.gitlab.com/ee/api/projects.html#edit-project) so it seems to be possible relatively easily.

weakcamel:
For what it's worth, I've used python-gitlab for a few minor tasks and it usually does the job well. Maintainers do a great job.

Once or twice it did go out of sync with Gitlab's API - but that's only to be expected with a relatively unstable interface that Gitlab provides. That said, there are certain functionalities (from EE tiers) which they may not be interested in implementing (such as e.g. https://github.com/python-gitlab/python-gitlab/issues/887) so keeping a fallback option might be a good thing.

max-wittig:
Maintainer of python-gitlab here. That would totally make sense. Contributions welcome. Let me know, if you need help!

gdubicki:
Hey @max-wittig , thanks a lot for pinging us here!

I think that that we could migrate 90% of GitLabForm to python-gitlab - the exceptions are the places where we by design pass the parameters from YAML as they are to the GitLab API endpoints (see https://github.com/egnyte/gitlabform/blob/master/FEATURES_DESIGN.md#raw-parameters-passing ).

Also we could do the migration gradually, firstly when adding new features or editing exisiting ones and then at some point complete the migration.

[...]

PR #582, opened 2023-08-01
feat: ability to transfer project between groups or namespaces

### LINKED ISSUE #213: Move projects between groups and subgroups
Basically the title says all wht I'd like: can I move one project or even more than one projects from one group to another group?

amimas:
This PR implements the ability to transfer a project in Gitlab. Project can be transferred from one namespace to another including subgroups. A new config key (`transfer_from`) has been introduced. See [this comment](https://github.com/gitlabform/gitlabform/issues/213#issuecomment-1586351323) for config syntax proposal/reason.

This feature uses python-gitlab for making API calls to Gitlab instead of expanding homegrown API. A new class is added so that it can be used for existing/future features.

closes #213


gitlabform/gitlab/__init__.py
```diff
@@ -74,3 +75,25 @@ class GitLab(
...
+            token,
+            ssl_verify=ssl_verify,
+            api_version="4",
+            session=session,
+            retry_transient_errors=True,
+            timeout=timeout,
+        )
+
+    def get_gitlab(self):
+        return self._gitlab
```
amimas, line 99: @gdubicki - Not sure how to best structure/setup the python-gitlab so that it can be used going forward. It should facilitate the work of https://github.com/gitlabform/gitlabform/issues/73. Is this the best way to start or do you have other suggestions?

amimas: @nejch - FYI ☝️

amimas:
> Surprisingly the smoke test is failing in CI even though the acceptance tests are passing. The smoke test is passing on my local machine too; running on python 3.10.

Needed to set python-gitlab as runtime dependency. duh... 🤦‍♂️ . It's resolved now.

gitlabform/gitlab/__init__.py
gdubicki: Looks good to me!

gdubicki (approved):
Looks great, thank you for the contribution, and sorry for the late review, @amimas!

gitlabform/processors/project/project_processor.py
```diff
@@ -1,17 +1,74 @@
...
+                        + "' group..."
+                    )
+                    project_to_be_transferred.transfer(
+                        project_transfer_destination_group
+                    )
+                except GitlabTransferProjectError:
+                    fatal(
+                        "Encountered error transferring project. Please check if project transfer requirements were met. Docs: https://docs.gitlab.com/ee/user/project/settings/index.html#transfer-a-project-to-another-namespace"
+                    )
+                    raise
```
amimas, line 66: Thanks for the review @gdubicki . I was taking a final glance before merging and a thought occurred.

Should this be a `fatal` error? Will gitlabform continue processing other projects or will it exit here?

gitlabform/gitlab/__init__.py
nejch: Sorry folks - I was super busy recently and didn't spend much time on GitHub. Looks good to me like this. I think eventually you could get rid of another layer of wrapping here, but looks like the right place to start.

gitlabform/processors/project/project_processor.py
gdubicki: Right, you should not re-raise it.

I forgot a little how it worked but I refreshed my memory. We have 3 kinds of errors in gitlabform:

1. Regular exceptions - these will break gitlabform run if `--terminate` flag is set in cli, or will just break processing given group or project if not - see [this code](https://github.com/gitlabform/gitlabform/blob/ignore_usernames_case/gitlabform/__init__.py#L397-L427),
2. Unconditionally fatal errors - if `fatal()` is called without an `if`, for example [here](https://github.com/gitlabform/gitlabform/blob/ignore_usernames_case/gitlabform/processors/multiple_entities_processor.py#L190-L194),
3. Conditionally fatal errors - if `--strict` flag is set in cli, if will break a gitlabform run, otherwise it will not. It's restricted to some special cases of given entity, usually a branch, not existing. See [this code](https://github.com/gitlabform/gitlabform/blob/ignore_usernames_case/gitlabform/processors/project/files_processor.py#L49-L56) as an example.

It may seem more chaotic than it actually is.

There is value in having a separate flag for NOT allowing some entities to not exist (`--strict`) and allowing that by default, because in one of the main use cases of this app I used to manage tens of repositories in a single group where some had only `master` branch, some had `develop` and `master` and I could have a single config that did some things with both branch but not fail if `develop` did not exist.

But probably we could make it more consistent by checking all the error cases and maybe expanding the cases that `--strict` covers. 🤔

gitlabform/gitlab/__init__.py
amimas: Thanks @nejch for the review.

gitlabform/processors/project/project_processor.py
amimas: Thanks @gdubicki for the details. I'll look into it and update my exception handling.

amimas: Hi @gdubicki - I think in this case it makes sense for this to be handled based on `--tarminate` flag. What do you think? I've never had it enabled in my actual usage though. And to be honest, I never quite understood the `--strict` flag (i.e. why another flag in addition to `--terminate`)?. Anyways, I think that's a discussion for separate issue or possibly for next major version design.

Unfortunately I couldn't figure out how to check for the `--terminate` flag in the above exception handling. From the `ProjectProcessor`, I'm not sure how to access it. So, I've just set it as a "warning" message for now. Could you please help when you get a chance? Thank you!

gitlabform/gitlab/__init__.py
amimas: I'm going to resolve this thread. Please feel free to re-open it or comment if we need to discuss further. Thanks again.

gitlabform/processors/project/project_processor.py
amimas: @gdubicki - Would you be able to help me with this update?

amimas:
I wonder how dry-run/noop parameter will process this feature. I'm not yet fully familiar with the franework or process in place for dryrun in this project.

gitlabform/processors/project/project_processor.py
```diff
@@ -1,17 +1,76 @@
...
+                    source_project_path_with_namespace
+                )
+                destination_project_path = project_and_group.split("/")[-1]
+                # Check if the project path needs to be updated; In Gitlab, path maybe different than name
+                if destination_project_path != project_to_be_transferred.path:
+                    verbose(
+                        "Updating the source project path from '"
+                        + project_to_be_transferred.path
+                        + "' to '"
+                        + destination_project_path
```
gdubicki, line 45: Nit: single quotation mark is not closed here.

gdubicki: You might consider replacing those multiple lines with f-strings, which might make the code more readable and such mistakes more rare.

gitlabform/processors/project/project_processor.py
gdubicki: After consideration I think you should not catch the `GitlabTransferProjectError` exception at all. If you won't do that, then [this code](https://github.com/gitlabform/gitlabform/blob/main/gitlabform/__init__.py#L475-L489) with catch it and either break the run (if `--terminate` flag is set) or just break processing given project.

gdubicki (commented):
Please also see the 2 minor comments about the code.

gitlabform/processors/project/project_processor.py
amimas: Thanks for looking into it @gdubicki . In that case, wouldn't the original setup have worked because I had re-raised the exception after logging a message? Should I revert to that code or do you think simply remove the exception catching/handling?

gitlabform/processors/project/project_processor.py
amimas: Thank you. I'll look into it.

gitlabform/processors/project/project_processor.py
amimas: I agree with you. I'll push an update and remove the exception catching so that we're consistent.

We can open a separate issue with a proposal for improving the logging of exception handling in similar cases.

gitlabform/processors/project/project_processor.py
amimas: I've pushed an update to address it. Please have a look.

gitlabform/processors/project/project_processor.py
amimas: Opened a separate issue to keep track of this: https://github.com/gitlabform/gitlabform/issues/611

I've removed the `try` and `except` that was catching the exception earlier. Well... I've commented them out as "Todo". I also realized that the acceptance for it wasn't capturing the custom message. It would pass regardless of what I wrote. Instead of removing the test case, I marked it to be skipped for now. It should probably be enabled and fixed later when exception catching/handling is done with custom messages as per above issue/idea.

Please review one more time.

gdubicki (approved):

PR #585, opened 2023-08-03
test: stabilize intermittent issue with enforce_keep_bots test

### LINKED ISSUE #569: `enforce_keep_bots` test intermittently fails in the GitHub CI
I noticed the `test__enforce_keep_bots` test intermittently fails in the GitHub CI. For example:

https://github.com/gitlabform/gitlabform/actions/runs/5594056862/jobs/10228464791?pr=564

I was able to replicate it locally too. The test passes maybe 3 or 4 out of 5 times. It'd be great to stabilize the test.

@SachinKSingh28 - You recently helped fix an issue related to "keep bots". Wondering if you ran into the issue while working on that fix. Any chance you know what the issue might be?

amimas:
This PR is related to `test__enforce_keep_bots` test case that intermittently fails but it does so quite often in GitHub Actions CI.

I was able to reproduce the issue locally too. The failure happens about 3 out of 5 runs of the test. In my local run, the failure was around asserting that the project has 2 members (i.e. test was getting 3 instead).

After a bit of debugging, I noticed that this test uses the same `project` fixture as the test (`test__enforce`) that runs before it. In that test, an additional user is added for validating that the `enforce: true` is working. Since the `project` fixture is scoped to a class, this project is also available to the `test__enforce_keep_bots` test.

When the intermittent issue does show up, the project's members still contains the user that was added via gitlabform config. Not sure why this user is intermittently available for the next test though. This is confusing to me.

To keep things a bit more isolated, in this PR I updated the 2nd test case (`test__enforce_keep_bots`) to use `project_for_function` fixture. This way the tests aren't sharing the same project. To accommodate for this change, I updated the `make_project_access_token` fixture so that we can tell it which project's access token should be created, as an optional parameter. Also refactored the `bot_members` fixture because it was adding a regular user to the project in addition to a project access token, which can lead to further confusions in the future.

With my change in this PR, I ran the test locally 10 times and didn't see the intermittent failures. Hopefully in Github CI, the test passes in first attempt. We should also try running it in this PR a few times for extra validations.

closes #569

### DIFF
--- a/tests/acceptance/conftest.py
+++ b/tests/acceptance/conftest.py
@@ -281,12 +281,12 @@ def make_project_access_token(
     created_tokens: List[ProjectAccessToken] = []
 
     def _make_project_access_token(
-        level: AccessLevel = AccessLevel.DEVELOPER, scopes: List[str] = ["api"]
+        target_project: Project = project, level: AccessLevel = AccessLevel.DEVELOPER, scopes: List[str] = ["api"]
     ) -> ProjectAccessToken:
         last_id = len(created_tokens) + 1
         token_name = f"{token_name_base}_{last_id}_bot"
         expires_at = (date.today() + timedelta(days=30)).isoformat()
-        token = project.access_tokens.create(
+        token = target_project.access_tokens.create(
             {
                 "access_level": level,
                 "name": token_name,
--- a/tests/acceptance/conftest.py
+++ b/tests/acceptance/conftest.py
@@ -276,12 +276,16 @@ def make_user(
 @pytest.fixture(scope="class")
 def make_project_access_token(
     project,
-) -> Generator[Callable[[AccessLevel, List[str]], ProjectAccessToken], None, None]:
+) -> Generator[
+    Callable[[Project, AccessLevel, List[str]], ProjectAccessToken], None, None
+]:
     token_name_base = get_random_name("user")
     created_tokens: List[ProjectAccessToken] = []
 
     def _make_project_access_token(
-        target_project: Project = project, level: AccessLevel = AccessLevel.DEVELOPER, scopes: List[str] = ["api"]
+        target_project=project,
+        level: AccessLevel = AccessLevel.DEVELOPER,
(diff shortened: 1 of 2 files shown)


tests/acceptance/conftest.py
```diff
@@ -276,17 +276,21 @@ def _make_user(
...
+    Callable[[Project, AccessLevel, List[str]], ProjectAccessToken], None, None
+]:
     token_name_base = get_random_name("user")
     created_tokens: List[ProjectAccessToken] = []
 
     def _make_project_access_token(
-        level: AccessLevel = AccessLevel.DEVELOPER, scopes: List[str] = ["api"]
+        target_project=project,
+        level: AccessLevel = AccessLevel.DEVELOPER,
+        scopes: List[str] = ["api"],
```
amimas, line 288: I expected the `target_project` to be typed like this:

```suggestion
        target_project: Project = project,
        level: AccessLevel = AccessLevel.DEVELOPER,
        scopes: List[str] = ["api"],
```

Even though the tests passes, `mypy` keeps complaining with following errors:

```
tests/acceptance/conftest.py:297: error: Argument 1 to "append" of "list" has incompatible type "RESTObject"; expected "ProjectAccessToken"  [arg-type]
tests/acceptance/conftest.py:298: error: Incompatible return value type (got "RESTObject", expected "ProjectAccessToken")  [return-value]
```

Not really sure why and how to address it other than what I have right now.

gdubicki: Frankly I am fed up with constant mypy issues. I don’t recall any case where it helped while I remember many ones where it was just an nuisance. I propose to get rid of this check completely. Any objections, @amimas ?

gdubicki (approved):
I love this PR, thank you for fixing it!

tests/acceptance/conftest.py
amimas: I opened a new issue (https://github.com/gitlabform/gitlabform/issues/586) to continue the discussion.

PR #592, opened 2023-08-24
chore(python): drop support for python 3.7

### LINKED ISSUE #576: Remove support for python 3.7 and possibly other older versions
This project is currently tested using several versions of python:

- 3.7
- 3.8
- 3.9
- 3.10
- 3.11

According to [status of python versions](https://devguide.python.org/versions/), version 3.7 is already end-of-life. I don't think there's any point in testing the project in that version of python. I think we should remove this version's tests from the CI.

As for version 3.8 - 3.10, they are also not in active development state. Only security releases will be available. Should we define a smaller matrix of older versions of python to be supported? Do we need to support that many versions of python? Not sure if I missed the reason in the docs. Maybe currently supported version and one previous version is good enough? 🤷

amimas:
Python 3.7 has reached EOL (end of life) as of June 2023. See [status of python versions](https://devguide.python.org/versions/). We're already unable to update some of our dependencies as they have also dropped support for python 3.7. For example: #588, #591

This PR drops support for Python 3.7 by gitlabform.

closes #576

### DIFF
--- a/.github/workflows/linter-and-tests.yml
+++ b/.github/workflows/linter-and-tests.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", 3.11]
+        python-version: [3.8, 3.9, "3.10", 3.11]
       fail-fast : false
     steps:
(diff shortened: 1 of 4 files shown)


gdubicki (approved):
LGTM!

PR #622, opened 2023-10-26
refactor: migrate project:archive feature to use python-gitlab library

### LINKED ISSUE #621: Migrate project:archive feature to use python-gitlab library
This issue is an extension or a task for #73 .

Goal of this issue is to migrate the `project:archive` feature to use [`python-gitlab` library](https://python-gitlab.readthedocs.io/en/stable/) instead of using homegrown API in this project.

amimas:
This PR builds on top of #582 that provided the setup to use  `python-gitlab` library. Refactored the `project:archive` feature's implementation to use the `python-gitlab` library instead of internal methods that called GitLab's API endpoint.

closes #621

### DIFF
--- a/gitlabform/processors/abstract_processor.py
+++ b/gitlabform/processors/abstract_processor.py
@@ -6,6 +6,8 @@ import requests
 from cli_ui import debug as verbose
 
 from gitlabform.gitlab import GitLab
+from gitlab import Gitlab
+from gitlabform.gitlab import GitlabWrapper
 from gitlabform.output import EffectiveConfigurationFile
 from gitlabform.processors.util.decorators import configuration_to_safe_dict
 
@@ -20,6 +22,7 @@ class AbstractProcessor(ABC):
                 [str, list[dict[str, Union[str, int]]], list[dict[str, int]]], bool
             ],
         ] = {}
+        self.gl: Gitlab = GitlabWrapper(self.gitlab)._gitlab
 
     @configuration_to_safe_dict
     def process(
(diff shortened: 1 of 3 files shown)


gitlabform/processors/abstract_processor.py
```diff
@@ -20,6 +22,7 @@ def __init__(self, configuration_name: str, gitlab: GitLab):
                 [str, list[dict[str, Union[str, int]]], list[dict[str, int]]], bool
             ],
         ] = {}
+        self.gl: Gitlab = GitlabWrapper(self.gitlab)._gitlab
```
amimas, line 25: @gdubicki - I added this here so that other features can easily use the `python-gitlab` library using the `Gitlab` object created here. This should be easily accessible so that others can help migrate rest of the features to `python-gitlab` library. That was my goal. Is there a better way to do this?

Also, should I rename `self.gl` to `self.Gitlab`? I'm not sure if that would be confusing with `self.gitlab`, which is gitlabform's internal implementation of connecting with GitLab API endpoint. Please correct me if I'm mistaken.

gdubicki: Hey, I think it's a great idea. The name is fine, we can refactor it in the future when we remove the internal library completely.

gdubicki (approved):

gitlabform/processors/abstract_processor.py
amimas: Sounds good. Thanks for reviewing.

Issue #632, opened 2023-10-31
Migrate `tags` configuration feature to use python-gitlab library


amimas:
Goal of this issue is to migrate the `tags` configuration feature to use [python-gitlab](https://python-gitlab.readthedocs.io/en/stable/index.html) library instead of using homegrown API in this project.

Todo:

- Delete `gitlabform/gitlab/tags.py` as it shouldn't be needed anymore
- Update `gitlabform/processors/project/tags_processor.py`
  - To use the python-gitlab library, use `self.gl` instead of `self.gitlab`. See the library docs for more help

amimas:
This is now available in [v3.9.0 release](https://github.com/gitlabform/gitlabform/releases/tag/v3.9.0). Thanks for your contribution.

PR #634, opened 2023-10-31
fix : Migrate tags configuration feature to use python-gitlab library

subhadeepbanerjee2003:
fixes : #632 

`changes made : ` 
- Deleted gitlabform/gitlab/tags.py as it shouldn't be needed anymore
- Updated gitlabform/processors/project/tags_processor.py
 - To use the python-gitlab library, used self.gl instead of self.gitlab.


gdubicki:
Hey @subhadeepbanerjee2003, thank you for the contribution! However this PR is not enough. To migrate from gitlabform's own API to python-gitlab you need to use different API methods, see [this code fragment and the whole PR](https://github.com/gitlabform/gitlabform/pull/622/files#diff-5dfd2bde153ece096b7be3e4cdc9c426bcd151c5ed68108f5203eb46b9cd5b86R62-R70) as an example. Also please make sure that tests pass on your localhosts before you push the code. Thanks!

subhadeepbanerjee2003:
@gdubicki ok thanks for your review but  I think I will not be able to do further as per my knowledge currently. So you can close it.

PR #635, opened 2023-11-04
refactor: migrate `hooks` configuration feature to use python-gitlab library

### LINKED ISSUE #624: Migrate `hooks` configuration feature to use python-gitlab library
Goal of this issue is to migrate the`hooks` configuration feature to use [python-gitlab library](https://python-gitlab.readthedocs.io/en/stable/) instead of using homegrown API in this project. This config relates to "webhook" configuration of projects in GitLab.

Todo:

- Update `gitlabform/gitlab/projects.py` and remove hooks related methods:
  - `get_hook_id`
  - `delete_hook`
  - `put_hook`
  - `post_hook`
 - Update `gitlabform/processors/project/hooks_processor.py` to use `python-gitlab` library instead of calling above mentioned methods

TigreModerata:
Removed ProjectHook methods from gitlab Projects; 
Changed HooksProcessor to use python-gitlab methods;
Added a test_hooks_processor file to acceptance tests to check if new hooks processor can create, update, delete

closes #624


gitlabform/processors/project/hooks_processor.py
```diff
@@ -9,23 +9,26 @@ def __init__(self, gitlab: GitLab):
...
             else:
-                hook_id = self.gitlab.get_hook_id(project_and_group, hook)
+                attr = {"url": hook}
+                attr.update(configuration["hooks"][hook])
                 if hook_id:
                     debug("Changing existing hook '%s'", hook)
-                    self.gitlab.put_hook(
-                        project_and_group, hook_id, hook, configuration["hooks"][hook]
-                    )
+                    changed_hook = project.hooks.update(hook_id, attr)
```
amimas, line 29: I'm leaning the python-gitlab library too. Is there a `update` method? Could you share where you found it in their docs? Maybe I'm not looking at the right place.

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def test_hooks_processor():
+    config = textwrap.dedent(test_yaml)
+    gf = GitLabForm(
+        include_archived_projects=False,
+        target="test_group/mystery",
+        config_string=config,
+    )
+    gl = GitlabWrapper(gf.gitlab).get_gitlab()
```
amimas, line 69: You don't need to import and use `GitlabWrapper` class. All the existing acceptance tests were already migrated to use the python-gitlab library. Just the app logic wasn't migrated yet. Please take a look at the existing tests.

amimas (changes requested):
@TigreModerata - Thank you so much for this PR. It's appreciated and looks pretty good. A few feedbacks from me, but might be best to have @gdubicki to review it as well.

Let's apply types to variables (ie. `myvar: str = "hello"`).

I had no idea the project didn't already have acceptance test for hooks. Thank you for adding one 😍 . Could you simplify the test file though? Take a look at the other test files to see how they are written. Here's what I can think about off the top of my head:

- Create separate function/method for different test cases (i.e. create, delete, update, etc)
- Use existing fixtures in your test functions/methods; this way you necessary group or project will be created for you already. Existing test files might be helpful as a reference.
- Move the gitlabform config into relevant test methods

gitlabform/processors/project/hooks_processor.py
TigreModerata: I saw it in a dream... I even invented the parameters - I'm almost (!!!) certain I saw it somewhere but I can only find "save" now. Casts a dark shadow on my test :(

tests/acceptance/standard/test_hooks_processor.py
TigreModerata: sorry, I moved my test a couple of times, didn't clean up properly.

TigreModerata:
Thank you for the feedback! Will get right on it.

gitlabform/processors/project/hooks_processor.py
amimas: Haha... If this exists, that'd be very helpful for some of the other logics to be migrated to python-gitlab library. See the following discussion: https://github.com/gitlabform/gitlabform/issues/73#issuecomment-1791761118

Looks like the acceptance test passed though. So, maybe this method does exist in the library and I just don't know how to process their documentation 😄

TigreModerata: I remember going through various links from the gitlab docs and finding all these examples of how to use the library - that's how I have this idea in my brain that the first parameter is the id (I had even had it without the id at some point and indeed, the test failed).... I'll dig deeper.

TigreModerata: OK, I might not be insane after all:
https://python-gitlab.readthedocs.io/en/v1.12.1/api/gitlab.html?highlight=update#

TigreModerata:
Had some mishaps, but it should be better now. 
Is there a better way to handle deletes, without having to time.sleep? I have a 6 second sleep because at times even 5 wasn't enough (running gitlab in docker)...

gitlabform/processors/project/hooks_processor.py
amimas: That looks like a very old version of python-gitlab library. Looks like in latest version it's also available here: https://python-gitlab.readthedocs.io/en/stable/api/gitlab.html#gitlab.mixins.UpdateMixin

amimas: Reading the API usage docs below, it seems `save()` used when updating something. Would be nice if it the `update()` was also documented with example.

https://python-gitlab.readthedocs.io/en/stable/api-usage.html

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,149 @@
...
+import pytest
+import textwrap
+import time
+from typing import TYPE_CHECKING
+
+import gitlab
+from gitlab.v4.objects import ProjectHook, Project, Group
+
+from gitlabform import GitLabForm
+from gitlabform.gitlab import GitlabWrapper
```
amimas, line 11: I think these 2 imports are not needed, right?

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,149 @@
...
+    created_project: Project = create_project(created_group, project)
+    return
+
+
+def reset_gitlab(gl):
+    with allowed_codes(404):
+        gl.groups.delete("test_group")
+        time.sleep(5)
+    return
+
```
amimas, line 47: Shouldn't need these methods (i.e. `prepare_project`, `reset_gitlab`). There are some common "pytest fixtures" already in `tests/acceptance/conftest.py` file. You simply add those fixtures in your test method as parameter. Depending on the lifecycle of the fixtures, they will get cleaned up automatically.

Check out the other test files at `tests/acceptance/standard/` directory.

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,149 @@
...
+        test_group/mystery:
+          project_settings:
+            rick: pickle
+          hooks:
+            "http://plumbus.org/create":
+                push_events: true
+                merge_requests_events: true
+            "http://birdman.chirp/update":
+                push_events: true
+      """
```
amimas, line 65: If you use the existing fixtures for setting up projects, you won't need the full config or hard-code the project name here. Just the `hooks` config should be enough. Then you can move this into the relevant test method. It'd be easier to read/debug the config that is being tested if this is within the test method.

gitlabform/processors/project/hooks_processor.py
```diff
@@ -9,23 +11,26 @@ def __init__(self, gitlab: GitLab):
         super().__init__("hooks", gitlab)
 
     def _process_configuration(self, project_and_group: str, configuration: dict):
+        debug("Processing hooks...")
+        project: Project = self.gl.projects.get(project_and_group)
+        hooks_list = project.hooks.list()
```
amimas, line 16: For consistency, let's apply types to all the variables (i.e. `hooks_list`, `hook`, `hook_id`, `attr`, etc.)

amimas (changes requested):

amimas:
> Is there a better way to handle deletes, without having to time.sleep? I have a 6 second sleep because at times even 5 wasn't enough (running gitlab in docker)...

Which delete are you referring to? Deleting hooks?

tests/acceptance/standard/test_hooks_processor.py
TigreModerata: yeah, I cleaned it up, sorry (very spartan ide, I keep forgetting myself)

tests/acceptance/standard/test_hooks_processor.py
TigreModerata: yep, understood

TigreModerata:
Yes, deleting hooks here - but that won't matter once I use the fixtures you mention above, rather than hardcode my configs. 
I just wondered in general, about any of the objects, especially when you have to delete + create to update (say protectedbranches, since I was looking at that too)...

TigreModerata:
I can't get the failing test to pass, I don't know what's wrong. 
Test file should be better now, using fixtures.

gitlabform/processors/project/hooks_processor.py
```diff
@@ -9,23 +13,26 @@ def __init__(self, gitlab: GitLab):
         super().__init__("hooks", gitlab)
 
     def _process_configuration(self, project_and_group: str, configuration: dict):
+        debug("Processing hooks...")
+        project: Project = self.gl.projects.get(project_and_group)
+        hooks_list: List[RESTObject] = project.hooks.list()
```
amimas, line 18: It looks alright to me but I'm not experienced enough. Reading the error message from type check failure here in the CI: https://github.com/gitlabform/gitlabform/actions/runs/6779285076/job/18426101684?pr=635

```
gitlabform/processors/project/hooks_processor.py:18: error: Incompatible types in assignment (expression has type "RESTObjectList | list[RESTObject]", variable has type "list[RESTObject]")  [assignment]
```

Should the type for `hooks_list` be `RESTObjectList` instead?

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,63 @@
+import logging
+
+from tests.acceptance import run_gitlabform, create_group, create_project
```
amimas, line 3: Don't need to import `create_group` and `create_project` here, right? They seem to be unused in this test file.

tests/acceptance/conftest.py
```diff
@@ -109,6 +110,20 @@ def project(gl: Gitlab, group: Group):
...
+    hook = project.hooks.create({"url": f"http://hooks/{name}.org"})
+
+    yield hook
+    # When using in hook deletion test this is Not Found
+    try:
+        hook.delete()
+    except (GitlabDeleteError, GitlabHttpError) as e:
+        if e.response_code == 404:
+            pass
+
```
amimas, line 125: Would it be better to move this fixture into the `tests/acceptance/standard/test_hooks_processor.py` file? Current pattern is that this file contains fixtures that are used across different feature tests. Do you see this fixture being used in other feature tests?

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,63 @@
...
+                hooks:
+                  {project_hook.url}:
+                    push_events: {project_hook.push_events}
+                    merge_requests_events: {project_hook.merge_requests_events}
+            """
+
+        run_gitlabform(test_yaml, target)
+
+        created_hook = project.hooks.get(project_hook.id)
+        assert project_hook == created_hook
```
amimas, line 21: I think we can add one more assertion about hooks count being `1`.

Another option could be add another hook in the test config with different attributes. That way the delete hook can remove one and confirm that only the selected hook was deleted.

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,63 @@
...
+        project_hook.id = original_hook.id
+
+        run_gitlabform(test_yaml, target, include_archived_projects=False)
+
+        modified_hook = project.hooks.get(original_hook.id)
+
+        assert modified_hook.asdict() != original_hook.asdict()
+        assert modified_hook == project_hook
+
+        modified_hook.delete()
```
amimas, line 63: Not sure if it's just me, but this test seems a little unintuitive. What do you think about calling `run_gitlabform` twice with different config to the hook.

After the first call to `run_gitlabform`, retrieve the hook and then call `run_gitlabform` again with a modified yaml config. Then retrieve the hook again. Compare the before and after result.

tests/acceptance/conftest.py
```diff
@@ -109,6 +110,20 @@ def project(gl: Gitlab, group: Group):
...
+    name = get_random_name("hook")
+    hook = project.hooks.create({"url": f"http://hooks/{name}.org"})
+
+    yield hook
+    # When using in hook deletion test this is Not Found
+    try:
+        hook.delete()
+    except (GitlabDeleteError, GitlabHttpError) as e:
+        if e.response_code == 404:
+            pass
```
amimas, line 124: Is it possible something else is going wrong here? Didn't think it's expected to return 404 when trying to delete a hook.

amimas (changes requested):
Thanks again for all the updates and keep digging into this. Looking a lot cleaner now.

gitlabform/processors/project/hooks_processor.py
TigreModerata: I'll try that! it seemed like it could be either, from the error message, but I don't understand it. Maybe a mistake was to use List instead of list... but then again that's what the docs said.

tests/acceptance/standard/test_hooks_processor.py
TigreModerata: oops... again!

tests/acceptance/conftest.py
TigreModerata: I wasn't sure, I thought that maybe it didn't matter, for tidiness' sake. I'll move it - thanks for all the patience, it's really my only way to learn on here!

tests/acceptance/standard/test_hooks_processor.py
TigreModerata: Ah, I see! When separating the create and the update, because I didn't want to muddy the waters I had created the hook to modify without passing through gitlabform... I thought let's test this and only this part of run() -> hooks_processor. But I can see how this might be a bit convoluted (I can convince myself of any rationale, that's the problem).

tests/acceptance/conftest.py
TigreModerata: It happened whenever I tested deleting a hook - first it got yielded from here, I use it and delete it in my test, then when it got back to here after its scope,  to remove the yielded hook, it was now gone (at least, that's what I thought was happening). 
When using the http calls directly, when a delete gets called it comes with "accepted_codes" including 404, so I thought passing this would be ok.

TigreModerata:
Type still not happy with ListRESTObjects for list of Project Hooks... 
test_files sometimes fails on here, but seems fine "on my machine"

gitlabform/processors/project/hooks_processor.py
TigreModerata: I think I need to explicitly say List[RESTObject] | RESTObjectList ? https://python-gitlab.readthedocs.io/en/stable/api/gitlab.html#gitlab.mixins.ListMixin.list)](https://python-gitlab.readthedocs.io/en/stable/api/gitlab.html#gitlab.mixins.ListMixin.list
I suppose I just thought that an "or", or "|", would accept both, but the return type of list() is what it is...

TigreModerata:
Surprisingly ;), ultimately it was a matter of rtfm, all of it :D

gitlabform/processors/project/hooks_processor.py
amimas: I see you finally figured out the type issue. I didn't know python types could be like that. Learned something new today. Guess we just needed to trust the hint from `mypy`, eh? 😅

tests/acceptance/conftest.py
amimas: No worries. I'm learning too. 😄

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,102 @@
...
+        original_yaml = f"""
+            projects_and_groups:
+              {target}:
+                hooks:
+                  {project_hook.url}:
+                    push_events: {project_hook.push_events}
+                    merge_requests_events: {project_hook.merge_requests_events}
+            """
+
+        run_gitlabform(original_yaml, target, include_archived_projects=False)
```
amimas, line 75: Just curious.. why does this call to `run_gitlabform` need to have `include_archived_projects=False` set explicitly and not for the other tests in this file?

tests/acceptance/standard/test_hooks_processor.py
amimas, line 1: Looking at the other tests files, I think we might want to rename this test file to `test_hooks.py`. That seems to be the pattern for configs that are applicable to "projects". Configs that are for GitLab groups seem to follow the pattern of `test_group_<config-name>.py`.

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class TestHooksProcessor:
+    def project_hook(self, project):
+        name = get_random_name("hook")
+        hook = project.hooks.create({"url": f"http://hooks/{name}.org"})
+
+        return hook
+
+    def test_project_hooks_create_delete(self, gl, project):
```
amimas, line 16: I just realized that you're using the `project` fixture. If you check out that fixture, you'll see that it's scoped to the entire test class. That means, you can actually split out your tests into separate methods like this:

```python
def test_project_hooks_create(self, gl, project):
    ...

def test_project_hooks_update(self, gl, project):
    ...

def test_project_hooks_delete(self, gl, project):
    ....
```

This way, whatever hooks are "created" in the first test case, will still be available to the second test case. After the necessary changes made in the second test case, remaining hooks (if any) will be available for the third test case.

So, basically you don't need to do the "original hooks" setup in each test case/methods. 

What you are currently doing would be suitable if you were using `project_for_function` fixture instead. That fixture is scoped to the functions. So, after each test case, the project is deleted. That means in each function you'd create some hooks first and then do what you wanted to do for the tests.

I think using the `project` fixture is fine. If you agree, let's refactor/restructure the test functions? I think it should actually simplify things a bit more.

amimas (changes requested):

tests/acceptance/standard/test_hooks_processor.py
amimas: I see. Thanks for discussing it. Changes are looking good so far.

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,102 @@
...
+        assert first_hook == first_created_hook
+        assert second_hook == second_created_hook
+
+        delete_yaml = f"""
+        projects_and_groups:
+            {target}:
+              hooks:
+                {first_hook.url}:
+                  delete: true
+        """
```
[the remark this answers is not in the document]
TigreModerata: I see... I thought this was exactly what I was trying to test, that even if the second hook was not included, it would not get deleted because it doesn't come with a delete: true in the config. I didn't know they could get deleted another way. OK, I'm happy to try to add the support, when this is done.

tests/acceptance/standard/test_hooks_processor.py
TigreModerata: Originally I had kept it in all of them (had run some tests to figure out how it all worked) and I forgot about it.

tests/acceptance/standard/test_hooks_processor.py
TigreModerata: Yeah, I was explicit so as not to confuse them with group hooks, or system hooks, or whatever else exists... But here we only deal with project hooks, so there is no need.

tests/acceptance/standard/test_hooks_processor.py
```diff
@@ -0,0 +1,102 @@
...
+
+from tests.acceptance import run_gitlabform, get_random_name
+
+
+class TestHooksProcessor:
+    def project_hook(self, project):
+        name = get_random_name("hook")
+        hook = project.hooks.create({"url": f"http://hooks/{name}.org"})
+
+        return hook
```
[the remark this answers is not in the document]
TigreModerata: Yeah, that's right, I had originally wanted to keep the object explicitly so I could access its properties one by one, but I had not deleted it from gitlab after creation... I'll remove this method.

TigreModerata:
Pushed a new version, having renamed the test file and implemented the other changes.

tests/acceptance/standard/test_hooks.py
```diff
@@ -0,0 +1,101 @@
...
+        test_yaml = f"""
+            projects_and_groups:
+              {target}:
+                hooks:
+                  {first_url}:
+                    push_events: false
+                    merge_requests_events: false
+                  {second_url}:
+                    job_events: false
+                    note_events: false
```
amimas, line 33: I believe default in GitLab when setting up hooks is false. Might make sense to set them to true. Or, can be mix of both.

tests/acceptance/standard/test_hooks.py
```diff
@@ -0,0 +1,101 @@
...
+
+        update_yaml = f"""
+            projects_and_groups:
+              {target}:
+                hooks:
+                  {first_url}:
+                    id: {first_hook.id}
+                    push_events: true
+                    merge_requests_events: true
+            """
```
amimas, line 67: In this "update hooks" test, let's add/remove events too.

Also, let's add the 2nd hooks config as we discussed in another comment/thread. You're asserting that hook's existence below already, which is good. It's just might not be clear to others why that hook should still exist.

tests/acceptance/standard/test_hooks.py
```diff
@@ -0,0 +1,101 @@
...
+
+        run_gitlabform(update_yaml, target)
+        modified_hook = self.get_hook_from_url(project, first_url)
+        unchanged_hook = self.get_hook_from_url(project, second_url)
```
amimas, line 71: Very minor comment. I think the variable name `modified_hook` and `unchanged_hook` is not clear. Would be good to name them something else. Does `updated_first_hook` and `updated_second_hook` make sense?

amimas (changes requested):
Looks really good. Left couple of minor comments for you to consider.

@gdubicki - When you get a chance, would you be able to take a glance at this PR please?

tests/acceptance/standard/test_hooks.py
TigreModerata: Push events defaults to true, but yeah, all the other to false...

tests/acceptance/standard/test_hooks.py
TigreModerata: Question: updating uses a PUT, not a PATCH. It's probably obvious, but it's maybe worth emphasising here? 
I think one could create a patch update, with python-gitlab (the UpdateMixin has a _get_update_method that can be set), but is it useful?

TigreModerata:
Made the changes and resubmitted.

tests/acceptance/standard/test_hooks.py
amimas: > When changing a hook we are effectively Patching (I'm updating the hook attribute dictionary, not overwriting it completely, before the PUT).

When you call the `project.hooks.update` method from python-gitlab library, is that a `PATCH` request to gitlab api or a `PUT` request.

I was also thinking that before you call `project.hooks.update`, can we check if there's any difference between what we got from configuration vs what is already configured in GitLab? If they are same, then we don't even need to call `project.hooks.update`, right? For now it's probably fine as is too.

> Enforce true means to strictly implement only what is in the config file

Yes, that's what the `enforce: true` under `hooks` would do if it was implemented. Delete all other hooks in the project but only keep/update what's in the config file.

amimas (commented):
Latest change looks good to me. Would be great to get some feedback from @gdubicki when he has some time.

tests/acceptance/standard/test_hooks.py
TigreModerata: sorry, had not seen this reply. When calling update it does a PUT, but in principle one could use either (the update mixin has a "get update method", meaning one can set it to patch or put, default is put). 
I'll add the check before updating.

amimas: Thank you so much for the update.

amimas:
Looks good 👍

amimas:
@gdubicki -- should we merge this? Will you have time for a sanity check?

gitlabform/processors/project/hooks_processor.py
```diff
@@ -9,23 +13,31 @@ def __init__(self, gitlab: GitLab):
...
-                else:
+                hook_config = {"url": hook}
+                hook_config.update(configuration["hooks"][hook])
+                if not hook_id:
                     debug("Creating hook '%s'", hook)
-                    self.gitlab.post_hook(
-                        project_and_group, hook, configuration["hooks"][hook]
+                    created_hook: RESTObject = project.hooks.create(hook_config)
+                    debug("Created hook '%s'", created_hook)
+                elif gitlab_hook and gitlab_hook.asdict() != hook_config:
```
gdubicki, line 38: Hi @TigreModerata!

Thank you for the contribution. :)

However I suspected that this condition will always be true because in `hook_config` we have only a few fields from a hook that we get from the API (`gitlab_hook.asdict()`), f.e. we don't have the field `id`.

And it turned out to be true - I modified a test `test_hooks_update` to do the update 2 times and in the second time the update is done again.

When I runned it with `pytest tests/acceptance/standard/test_hooks.py -k 'test_hooks_create or test_hooks_update'`.

First time logs excerpt:

```
Processing hooks...
http://localhost:80 "GET /api/v4/projects/gitlabform_group_papaya%2Fgitlabform_project_stony HTTP/1.1" 200 None
http://localhost:80 "GET /api/v4/projects/8/hooks HTTP/1.1" 200 None
Changing existing hook 'http://hooks/gitlabform_hook_festival.org'
http://localhost:80 "PUT /api/v4/projects/8/hooks/9 HTTP/1.1" 200 None
Changed hook to '{'id': 9, 'url': 'http://hooks/gitlabform_hook_festival.org', 'created_at': '2023-12-03T11:41:51.849Z', 'push_events': True, 'tag_push_events': False, 'merge_requests_events': False, 'repository_update_events': False, 'enable_ssl_verification': True, 'alert_status': 'executable', 'disabled_until': None, 'url_variables': [], 'project_id': 8, 'issues_events': False, 'confidential_issues_events': False, 'note_events': True, 'confidential_note_events': None, 'pipeline_events': False, 'wiki_page_events': False, 'deployment_events': False, 'job_events': True, 'releases_events': False, 'push_events_branch_filter': None, 'emoji_events': False}'
Changing existing hook 'http://hooks/gitlabform_hook_tiring.org'
http://localhost:80 "PUT /api/v4/projects/8/hooks/10 HTTP/1.1" 200 None
Changed hook to '{'id': 10, 'url': 'http://hooks/gitlabform_hook_tiring.org', 'created_at': '2023-12-03T11:41:51.907Z', 'push_events': False, 'tag_push_events': False, 'merge_requests_events': False, 'repository_update_events': False, 'enable_ssl_verification': True, 'alert_status': 'executable', 'disabled_until': None, 'url_variables': [], 'project_id': 8, 'issues_events': False, 'confidential_issues_events': False, 'note_events': True, 'confidential_note_events': None, 'pipeline_events': False, 'wiki_page_events': False, 'deployment_events': False, 'job_events': False, 'releases_events': False, 'push_events_branch_filter': None, 'emoji_events': False}'
* (1/1) FINISHED Processing project: gitlabform_group_papaya/gitlabform_project_stony
```

Second time logs excerpt:
```
Processing hooks...
http://localhost:80 "GET /api/v4/projects/gitlabform_group_papaya%2Fgitlabform_project_stony HTTP/1.1" 200 None
http://localhost:80 "GET /api/v4/projects/8/hooks HTTP/1.1" 200 None
Changing existing hook 'http://hooks/gitlabform_hook_festival.org'
http://localhost:80 "PUT /api/v4/projects/8/hooks/9 HTTP/1.1" 200 None
Changed hook to '{'id': 9, 'url': 'http://hooks/gitlabform_hook_festival.org', 'created_at': '2023-12-03T11:41:51.849Z', 'push_events': True, 'tag_push_events': False, 'merge_requests_events': False, 'repository_update_events': False, 'enable_ssl_verification': True, 'alert_status': 'executable', 'disabled_until': None, 'url_variables': [], 'project_id': 8, 'issues_events': False, 'confidential_issues_events': False, 'note_events': True, 'confidential_note_events': None, 'pipeline_events': False, 'wiki_page_events': False, 'deployment_events': False, 'job_events': True, 'releases_events': False, 'push_events_branch_filter': None, 'emoji_events': False}'
Changing existing hook 'http://hooks/gitlabform_hook_tiring.org'
http://localhost:80 "PUT /api/v4/projects/8/hooks/10 HTTP/1.1" 200 None
Changed hook to '{'id': 10, 'url': 'http://hooks/gitlabform_hook_tiring.org', 'created_at': '2023-12-03T11:41:51.907Z', 'push_events': False, 'tag_push_events': False, 'merge_requests_events': False, 'repository_update_events': False, 'enable_ssl_verification': True, 'alert_status': 'executable', 'disabled_until': None, 'url_variables': [], 'project_id': 8, 'issues_events': False, 'confidential_issues_events': False, 'note_events': True, 'confidential_note_events': None, 'pipeline_events': False, 'wiki_page_events': False, 'deployment_events': False, 'job_events': False, 'releases_events': False, 'push_events_branch_filter': None, 'emoji_events': False}'
* (1/1) FINISHED Processing project: gitlabform_group_papaya/gitlabform_project_stony
```

This is not great as it lowers the performance of a run and may cause unnecessary audit events in the projects on each run ("Hook ... has been updated").

I have pushed my commit to show the problem here https://github.com/gitlabform/gitlabform/commit/d8b84e1a4aa5860564059a8a829e916a781ea9a7.

amimas: @TigreModerata - no rush but just wondering if you will be able to look into this feedback? I feel this PR is so close to being merged🤞 

It's a good callout from @gdubicki . I missed this.

TigreModerata:
OK I think I fixed it. I included a test, reading the debug logs to check if indeed an identical config is not getting updated.

amimas:
> OK I think I fixed it. I included a test, reading the debug logs to check if indeed an identical config is not getting updated.

Thank you @TigreModerata . appreciate it.

@gdubicki - could we get one more review please?

amimas (approved):
Looks good. Thanks again for this PR.

TigreModerata:
Cool, sorry again for all the missteps!

jimisola:
> Cool, sorry again for all the missteps!

Absolutely no reason to say sorry. Thank you for giving your time and contributing.

PR #638, opened 2023-11-06
Migrate branches gitlab4

TigreModerata:
Removed branches.py
Migrated branch_protector and files_processor where it used branches methods directly.
Acceptance tests already exist.


amimas:
Duplicated and completed by #876. Thanks for looking into it.

PR #642, opened 2023-11-15
refactor: migrate `tags` processor to use python-gitlab API

### LINKED ISSUE #632: Migrate `tags` configuration feature to use python-gitlab library

lhokktyn:
Refactors `gitlabform/processors/project/tags_processor.py` to use the equivalent [python-gitlab](https://python-gitlab.readthedocs.io/en/stable/index.html) APIs.

Closes #632


lhokktyn:
Hi @amimas. Hope you don't mind me tagging you in this thread, but noticed you'd worked on https://github.com/gitlabform/gitlabform/pull/551 which is related to this change and wondered if you wouldn't mind casting your eye over this when you get a chance? Cheers

amimas:
Hi @lhokktyn - Thanks so much for the PR. Looks like the existing tests are passing. That's a good, but would be great to add a test case for the scenario you mentioned where you ran into issue.

Also, do you think you can help with #632? That will take out issues like this related to protected tags.

lhokktyn:
> Hi @lhokktyn - Thanks so much for the PR. Looks like the existing tests are passing. That's a good, but would be great to add a test case for the scenario you mentioned where you ran into issue.

On closer inspection, I believe the `NotFoundException` was actually triggered by a call to `self.gitlab._get_user_id(config["user"])` earlier on in the function, caused by my user error trying to use the "name" of a Group Access Token bot rather than the "username" or ID!

> Also, do you think you can help with #632? That will take out issues like this related to protected tags.

I can give it a go, sure.

lhokktyn:
Have pushed an initial attempt at refactoring. However, I have no means to test the premium features locally - will this be covered elsewhere prior to any release?

amimas (changes requested):
Don't nhave access to desktop roght now. Did a quick review on mobile. Looks good but need to make sure tests passes.

In `tag_processor.py`, there's still a call to `self.gitlab` for retrieving user id. Could we use the python-gitlab library for that using `self.gl`?

amimas:
> Have pushed an initial attempt at refactoring. However, I have no means to test the premium features locally - will this be covered elsewhere prior to any release?

Yeah, testing features of paid version requires having a gitlab license. They're tested in every PR in the CI but require a manual review and approval. I could trigger the test in the CI but it'll be extremely slow process. Every commit you push, you have to wait for someone in this repo to do this for you.

I had ended up signing up for a free trial with gitlab amd used that license locally to run the tests and then make changes as needed. 

BTW - thanks for the PR and helping with this. Also, just for clarification, does the original issue exist? Not sure I followed what you said about the exception to my first reply.

gitlabform/processors/project/tags_processor.py
```diff
@@ -55,20 +57,26 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
                         if "create_access_level" in configuration["tags"][tag]
                         else None
                     )
+
                     debug("Setting tag '%s' as *protected*", tag)
                     try:
                         # try to unprotect first
-                        self.gitlab.unprotect_tag(project_and_group, tag)
-                    except NotFoundException:
+                        project.protectedtags.delete(tag)
```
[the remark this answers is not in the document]
amimas: Just had a chance to check out the gitlab api docs and python-gitlab docs. This looks alright. I just assumed `delete()` meant actually deleting the tag.

Let's close this thread.

gitlabform/processors/project/tags_processor.py
```diff
@@ -55,20 +57,26 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
+                        data["allowed_to_create"] = allowed_to_create
+                    if create_access_level is not None:
+                        data["create_access_level"] = create_access_level
+                    project.protectedtags.create(data)
                 else:
                     debug("Setting tag '%s' as *unprotected*", tag)
-                    self.gitlab.unprotect_tag(project_and_group, tag)
-            except NotFoundException:
-                message = f"Tag '{tag}' not found when trying to set it as protected/unprotected!"
+                    project.protectedtags.delete(tag)
```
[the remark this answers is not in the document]
amimas: Closing this thread. See the above thread for details.

lhokktyn:
Thanks for the feedback @amimas, I'll pick through this and get back with an update asap.

tests/acceptance/standard/test_tags.py
```diff
@@ -122,6 +122,19 @@ def test__unprotect_the_same_tag(self, project, tags):
         protected_tags = project.protectedtags.list()
         assert len(protected_tags) == 0
 
+    def test__unprotect_unknown_tag(self, project, tags):
+        config = f"""
+        projects_and_groups:
+          {project.path_with_namespace}:
+            tags:
+              "tag_unknown":
```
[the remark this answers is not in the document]
lhokktyn: No, it was caused by my own mistake :)

gitlabform/processors/project/tags_processor.py
```diff
@@ -33,13 +36,20 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
                             elif "user_id" in config:
                                 user_ids.add(config["user_id"])
                             elif "user" in config:
-                                user_ids.add(self.gitlab._get_user_id(config["user"]))
+                                users_by_username = self.gl.users.list(
+                                    username=config["user"]
+                                )
+                                if len(users_by_username) == 0:
+                                    raise NotFoundException(
```
amimas, line 43: A minor comment. Is there an appropriate exception from the python-gitlab library that can be caught by the search/list function and raised instead of this one?

gitlabform/processors/project/tags_processor.py
```diff
@@ -33,13 +36,20 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
                             elif "user_id" in config:
                                 user_ids.add(config["user_id"])
                             elif "user" in config:
-                                user_ids.add(self.gitlab._get_user_id(config["user"]))
+                                users_by_username = self.gl.users.list(
```
amimas, line 39: I believe we're looking for a single user so that wr can find their ID. Lets change the variable name to be more appropriate. Something like `user` or `gitlab_user`.

gitlabform/processors/project/tags_processor.py
```diff
@@ -33,13 +36,20 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
+                                        % config["user"]
+                                    )
+                                user_ids.add(users_by_username[0]["id"])
                             elif "group_id" in config:
                                 group_ids.add(config["group_id"])
                             elif "group" in config:
-                                group_ids.add(
-                                    self.gitlab._get_group_id(config["group"])
-                                )
+                                group_by_name = self.gl.groups.get(config["group"])
```
amimas, line 51: Similar to users, lets capture exception about group if it doesn't exist.

amimas (changes requested):
Couple more feedbacks. Also, for consistency, let's set appropriate types for each variables.

amimas:
Let's chamge the PR description and title to reflect the current state since the original issue mentioned doesn't exist.

gitlabform/processors/project/tags_processor.py
lhokktyn: There wasn't an equivalent error in the [ListMixin](https://python-gitlab.readthedocs.io/en/stable/api/gitlab.html#gitlab.mixins.ListMixin) for "404", so I've wrapped up an empty list result in a `GitlabGetError` which mirrors the exception thrown by `gl.groups.get(config["group"])` when a group is missing.

gitlabform/processors/project/tags_processor.py
lhokktyn: Changed to `gitlab_user`

gitlabform/processors/project/tags_processor.py
lhokktyn: Changed to `gitlab_group`, and caught raised `GitlabGetError`

gitlabform/processors/project/tags_processor.py
amimas: Thanks for the update. Not sure if the behaviour between listing user and group is by design in the python-gitlab library. Maybe @nejch can help clarify.

gitlabform/processors/project/tags_processor.py
amimas: Not sure if I missed it buf didn't see the exception being caught. Are we just expectung the code to reach the exception from the python-gitlab libray and not need to catch it here? Maybe we don't need to catch it here?

amimas:
@lhokktyn - acceptance test for tag protection is failing in the latest update. Also an issue in the mypy type checking. When you get a chance, could you please take a look?

gitlabform/processors/project/tags_processor.py
nejch: Hey @lhokktyn @amimas! this is expected and normal - you're using two different calls. A `.get()` may return a 404 which will trigger a `GitlabGetError`, for both `users.get(user_id)` and `groups.get(group_id`. 

But a `.list()` may simply return an empty list based on your filters, and this will not raise an exception, for both groups and users.

For this reason I always prefer working with IDs directly instead of performing these searches that need additional API calls but I understand it's more readable for end users to have this as well :+1: 

I'd maybe just use `if not user:` instead of `if len(...)` as per https://peps.python.org/pep-0008/ :sweat_smile:

amimas: Thank you @nejch for the clarification and your insight.

gitlabform/processors/project/tags_processor.py
lhokktyn: Yeh, I'm letting the `GitlabGetError` exception from either `self.gl.groups.get(config["group"])` or `self._get_user_by_username(config["user"])` bubble up to eventually be caught by the top-level `except` in `_process_configuration()`. If neither the user or group can be found the rest of the config can't be safely applied so just bails out at that point.

amimas:
Looks good. @gdubicki - will you bre available for a quick check?

gitlabform/processors/project/tags_processor.py
amimas: Hi @lhokktyn - has this been resolved already or anything still outstanding? Sorry it's been awhile since we looked at this.

amimas: Just went through the changes again and I think this is fine. Going to resolve this now.

amimas (approved):
Thanks for this PR 🤝

Issue #654, opened 2023-12-18
Use of expires_at in project members fails due to debug function


HontoNoRoger:
**Describe the bug**

When using `expires_at` for a project membership, gitlabform fails with an exception:

```
$ gitlabform -n ALL_DEFINED
🏗 GitLabForm version: 3.8.0 
= the latest stable 😊
>>> Getting ALL groups and projects DEFINED in the configuration...
Traceback (most recent call last):
  File "/usr/local/bin/gitlabform", line 8, in <module>
    sys.exit(run())
  File "/usr/local/lib/python3.10/dist-packages/gitlabform/run.py", line 5, in run
    GitLabForm().run()
  File "/usr/local/lib/python3.10/dist-packages/gitlabform/__init__.py", line 360, in run
    projects, groups = self._show_header(
  File "/usr/local/lib/python3.10/dist-packages/gitlabform/__init__.py", line 5[90](https://gitlab.sirius.supplyon.com/sirius/platform/cicd/gitlab/gitlab-configuration/-/jobs/528042#L90), in _show_header
    self.groups_and_projects_filters.filter(groups, projects)
  File "/usr/local/lib/python3.10/dist-packages/gitlabform/lists/filter.py", line 25, in filter
    self.omit_empty_configs.filter(groups, projects)
  File "/usr/local/lib/python3.10/dist-packages/gitlabform/lists/filter.py", line 77, in filter
    if self._project_has_empty_effective_config(project):
  File "/usr/local/lib/python3.10/dist-packages/gitlabform/lists/filter.py", line [99](https://gitlab.sirius.supplyon.com/sirius/platform/cicd/gitlab/gitlab-configuration/-/jobs/528042#L99), in _project_has_empty_effective_config
    config_for_project = self.configuration.get_effective_config_for_project(
  File "/usr/local/lib/python3.10/dist-packages/gitlabform/configuration/projects.py", line 53, in get_effective_config_for_project
    debug("Project config: %s", to_str(project_config))
  File "/usr/local/lib/python3.10/dist-packages/gitlabform/util.py", line 7, in to_str
    return json.dumps(a_dict, sort_keys=True)
  File "/usr/lib/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.10/json/encoder.py", line 257, in iterencode
[...]

PR #656, opened 2023-12-20
fix: json data with dates not serialized correctly

HontoNoRoger:
The datetime.date object used for access
expiration etc. is not supporting JSON serialization.
Trying to do this for debug output reasons
ends with an exception. The json.dumps() function
offers to specify a default handling of
unsupported types which is utilized here.

Resolves: #654

### DIFF
--- a/gitlabform/util.py
+++ b/gitlabform/util.py
@@ -4,4 +4,4 @@ import json
 def to_str(a_dict: dict) -> str:
     # arguably the most readable form of a dict in a single line
     # is JSON with sorted keys
-    return json.dumps(a_dict, sort_keys=True)
+    return json.dumps(a_dict, sort_keys=True, default=str)
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -1,0 +1,20 @@
+import datetime
+
+from gitlabform.util import to_str
+
+
+def test_date_stringification():
+    example_project_config = {
+        "members": {
+            "enforce": True,
+            "users": {
+                "project_204_bot": {"access_level": 40},
+                "project_204_bot_211c9f071a88910e7ed518cc5d81436a": {
+                    "access_level": 40,
+                    "expires_at": datetime.date(2024, 12, 12),
+                },
+            },
+        }
+    }
+    output = to_str(example_project_config)
+    assert type(output) is str


amimas (commented):
Thanks for the PR. I'm not a python expert. Still learning. Could you please share a doc around the `default=str` in the `json.dumps`? 

I ended up browsing to the following doc that says default should be a function. Maybe I'm not looking at the correct info.

https://docs.python.org/3/library/json.html#json.JSONEncoder

Also, thanks for adding the unit test. Could we also include acceptance test. Or do you think that's unnecessary?

amimas:
I edited the PR title. Please feel free to update if I didn't capture it properly.

HontoNoRoger:
> Thanks for the PR. I'm not a python expert. Still learning. Could you please share a doc around the `default=str` in the `json.dumps`?
> 
> I ended up browsing to the following doc that says default should be a function. Maybe I'm not looking at the correct info.
> 
> https://docs.python.org/3/library/json.html#json.JSONEncoder
> 
> Also, thanks for adding the unit test. Could we also include acceptance test. Or do you think that's unnecessary?

The document you reference is correct. Especially this part is of interest:

> To extend this to recognize other objects, subclass and implement a [default()](https://docs.python.org/3/library/json.html#json.JSONEncoder.default) method with another method that returns a serializable object for o if possible, otherwise it should call the superclass implementation (to raise [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError)).

It expects a method that returns a serializable object for whatever it gets. The `str` we use here as value to the default parameter points to the str class/type: https://docs.python.org/3/library/stdtypes.html#str. For a given object, it handles conversion based on the `__str__()` representation if it exists or uses `repr()`. That way, more types can be handled by our `to_str()` function, including the `date`.

```python
In [11]: datetime.date(2024, 12, 12).__str__()
Out[11]: '2024-12-12'
```

As for acceptance testing, I believe it is not required in this case, as we are not fixing functionality that requires GitLab itself. It's just about fixing the handling of Python internal types used in a helper function for debugging purposes.

amimas:
Thank you @HontoNoRoger for the detailed explanation. It makes sense now.

amimas (approved):
Looks good.

amimas:
@HontoNoRoger - looks like all the acceptance tests are failing after merging this PR. Same tests are also failing after updating some of the dependabot PRs after this one was merged. Not sure why everything passed in this PR before merging it.

When you get a chance, could you please take a look? Thanks.

HontoNoRoger:
@amimas did a quick cross-check, I believe this is due to the newer version of GitLab being tested against on the main branch.

GitLab version in acceptance-tests (3.9) before merge (3 days ago):

`{"version":"16.6.2-ee","revision":"fddcadf1869","kas":{"enabled":true,"externalUrl":"ws://localhost/-/kubernetes-agent/","version":"v16.6.0"},"enterprise":true}`

GitLab version in acceptance-tests (3.9) after merge (1 hour ago):

`{"version":"16.7.0-ee","revision":"9e7d34f7ff1","kas":{"enabled":true,"externalUrl":"ws://localhost/-/kubernetes-agent/","version":"v16.8.0-rc1"},"enterprise":true}`

amimas:
Thabks for taking a quick look at it.

amimas:
Just confirmed the above by re-running tge acceptance test on python 3.12 from the previous commit on the main branch. It was successful before and now failing with latest version of GitLab (16.7) 😔

https://github.com/gitlabform/gitlabform/actions/runs/7177097321/job/19914875671

amimas:
I think we should make a minor release. There's at least one unreleased feature merged since last release. At the moment not sure when the release can happen. Hopefully soon 🤞 .

amimas:
This is now available in [v3.9.0 release](https://github.com/gitlabform/gitlabform/releases/tag/v3.9.0). Thanks for your patience.

PR #669, opened 2024-01-17
Do not add entities that are marked as delete

KrystianOcado:
Hi!

Current version of gitlabform tries to add entities (e.g. variables) that are present only in config, even if they are marked as delete.

This means, that on first run, gitlabform will delete entity "foo" and on the next run it will try to add it again.

This should fix this issue.


amimas:
Hi. Thanks for this PR. Is there an existing open issue for it? Could you please reference it or create a new issue? I didn't completely follow the problem you mentioned. Also it seems the acceptance tests are failing.

amimas:
Hi @KrystianOcado - Did you get a chance to look into what I mentioned previously ☝️ ?

KrystianOcado:
Hi @amimas !

Sorry about the delay, I've added an issue here: https://github.com/gitlabform/gitlabform/issues/684

Please let me know if you need more info.

gdubicki:
Hey @KrystianOcado! We would love to get your contribution merged, but the tests need to pass first. Can you find the time to revisit this PR and fix them please?

Issue #620, opened 2023-10-24
Enable automerge for dependency updates by dependabot


amimas:
This project PRs by dependabot with updates to dependencies. Processing these PRs are manual.

The PRs need approval for running the acceptance tests that requires GitLab license. Once the approval is given and all the CI jobs completed successfully, the PR needs to be merged.

Since the project has pretty good test coverage/validation, we should enable auto merging these dependency updates.

gdubicki:
I am totally for. Its just that it's not a switch in the config - we would need a bit of code to do that: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions

amimas:
Have you used [renovate](https://docs.renovatebot.com/)? I'm not familiar enough with dependabot. So, can't say if renovate would be more useful instead. Thought I'd mention it in case it's something worth trying out.

amimas:
I suggest we keep this open. Even when we move to gitlab.com, this will still need to be worked on.

PR #691, opened 2024-02-21
Lower the amount of PRs created by Dependabot

gdubicki:
We only want security updates PRs +  updates for some key packages

But this change means that occasionally we should check our other dependencies for updates manually!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,11 +1,14 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
-# Please see the documentation for all configuration options:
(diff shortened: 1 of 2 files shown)


amimas:
> We only want security updates PRs + updates for some key packages

I actually feel we should stay up to date on all of our dependencies. If needed, we could reduce the frequency for how often dependabot openns PR but I think it's fine as is? What would be great or help out is to implement #620 .

As you said, this will require manually checking and updating our other dependencies. Given the lack of resource or dedicated maintainer for this project, I'm afraid we'll just end up with outdated dependencies. Eventually we might even face a situation where significant effort might be needed for updating one of the dependencies which might have issues or vulnerabilities.

Issue #364, opened 2022-05-10
Updating pipeline schedules is not working


stuart-warren:
**Describe the bug**

Similar to #361 updating pipeline schedules seems to have the same issue

```py
Processing section 'schedules'
Error: Error occurred while processing project foo/gitlab/gitlabform, exception:
Request url='https://gitlab.foo.tech//api/v4/projects/foo%2Fgitlab%2Fgitlabform/pipeline_schedules/2464', method=PUT, data='{"active": true, "cron": "0 10 * * MON-FRI", "cron_timezone": "London", "ref": "master"}' failed - expected code(s) [200, 201], got code 403 & body: '{"message":"403 Forbidden"}'
Traceback (most recent call last):
  File "/gitlabform/gitlabform/core.py", line 430, in run
    self.project_processors.process_entity(
  File "/gitlabform/gitlabform/processors/__init__.py", line 30, in process_entity
    processor.process(
  File "/gitlabform/gitlabform/processors/util/decorators.py", line 42, in method_wrapper
    return method(self, project_and_group, SafeDict(configuration), *args)
  File "/gitlabform/gitlabform/processors/abstract_processor.py", line 48, in process
    self._process_configuration(project_or_project_and_group, configuration)
  File "/gitlabform/gitlabform/processors/project/schedules_processor.py", line 38, in _process_configuration
    updated_schedule = self.gitlab.update_pipeline_schedule(
  File "/gitlabform/gitlabform/gitlab/schedules.py", line 39, in update_pipeline_schedule
    return self._make_requests_to_api(
  File "/gitlabform/gitlabform/gitlab/core.py", line 127, in _make_requests_to_api
    response = self._make_request_to_api(
  File "/gitlabform/gitlabform/gitlab/core.py", line 218, in _make_request_to_api
    raise UnexpectedResponseException(
gitlabform.gitlab.core.UnexpectedResponseException: Request url='https://gitlab.foo.tech//api/v4/projects/foo%2Fgitlab%2Fgitlabform/pipeline_schedules/2464', method=PUT, data='{"active": true, "cron": "0 10 * * MON-FRI", "cron_timezone": "London", "ref": "master"}' failed - expected code(s) [200, 201], got code 403 & body: '{"message":"403 Forbidden"}'
```

**GitLabForm version**

`2.11.0`

**GitLab version**

F.e. `14.9.4-ee`

stuart-warren:
[...]

PR #706, opened 2024-03-13
refactor: migrate `schedules_processor` to use python-gitlab

### LINKED ISSUE #631: Migrate `schedules` configuration feature to use python-gitlab library
Goal of this issue is to migrate the `schedules` configuration feature to use [python-gitlab library](https://python-gitlab.readthedocs.io/en/stable/) instead of using homegrown API in this project.

Todo:

- Delete `gitlabform/gitlab/schedules.py` as it shouldn't be needed anymore
- Update `gitlabform/processors/project/schedules_processor.py`

TimKnight-Opencast:
- closes #631


amimas:
Thanks for this PR @TimKnight-Opencast . Based on PR description and file changes, it seems branch protection processor is also included. Really appreciate the help converting both schedule processor and branch protection processor. But, it's a lot easier to review and safer if we keep the scope of changes to minimum. Could you please open a separate PR for branch protection processor related changes?

TimKnight-Opencast:
@amimas done I've cut a new branch for branch_processor changes.

Will raise a PR once the group_members refactor is in so can re-use the common python-gitlab code I'm adding in there

TimKnight-Opencast:
(same person, I just work for a consultancy at DWP, and due to stuff, needed to contribute from this account instead last week 😸 )

gitlabform/processors/project/schedules_processor.py
```diff
@@ -42,102 +44,149 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
+        active = self._get_active_state_from_config(data)
+
+        schedule_data = {
+            "ref": data.get("ref"),
+            "description": schedule_description,
+            "cron": data.get("cron"),
+            "cron_timezone": cron_timezone,
+            "active": active,
+        }
+        debug("Creating pipeline schedule using data: '%s'", schedule_data)
```
amimas, line 119: Same comment as above regarding passing all data from config.

amimas (changes requested):
It mostly looks good to me. Thanks for taking care of this. Only major issue I see is how the schedule is being created/updated.

gitlabform/processors/project/schedules_processor.py
```diff
@@ -42,102 +44,149 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
+                    schedule.variables.create(
+                        {
+                            "key": variable_key,
+                            "value": variable_data.get("value"),
+                        }
+                    )
 
     @staticmethod
-    def __group_schedule_ids_by_description(schedules) -> Dict[str, List[str]]:
+    def _group_schedule_ids_by_description(schedules) -> Dict[str, List[str]]:
```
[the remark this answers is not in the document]
TimKnight-Opencast: will push as two commits, if I recall @amimas , the accurate notation `schedules: List[ProjectPipelineSchedule] | RESTObjectList` is incompatible with python 3.9

gitlabform/processors/project/schedules_processor.py
```diff
@@ -42,102 +48,124 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
-                debug(f"processing {schedule_id}: {schedule_description}")
-                if schedule_description not in configured_schedules:
-                    debug(
-                        "Deleting pipeline schedule named '%s', because it is not in gitlabform configuration",
-                        schedule_description,
-                    )
-                    self.gitlab.delete_pipeline_schedule(project_and_group, schedule_id)
+            self._create_schedule_with_variables(
+                configured_schedules, project, schedule_description
+            )
```
amimas, line 93: Why do we need to delete the existing schedule and create a new one? Can't we just perform an update? The API supports a `PUT` operation. That's what the current logic is uses.

https://github.com/gitlabform/gitlabform/blob/69cdd7ab49a0b1516ddba5be4ada5760dc742041/gitlabform/gitlab/schedules.py#L36-L45

Creating a new schedule might be problematic. Each schedule has a separate ID. Imagine running gitlabform periodically. There'll be a ton of records of schedule in the database because of constant delete and create new. Doing this will also make the following API useless to end users because they will lose the history.

https://docs.gitlab.com/ee/api/pipeline_schedules.html#get-all-pipelines-triggered-by-a-pipeline-schedule

So, I don't think updating schedule should require deleting the current one and creating a new one. Let me know if there's something I'm unaware of that requires doing this.

gitlabform/processors/project/schedules_processor.py
```diff
@@ -42,102 +48,124 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
                     debug(
                         "Changing existing pipeline schedule '%s'", schedule_description
                     )
```
amimas, line 50: The latest logic of `_update_existing_schedule` is performing checks to see if update is needed. This log should be in there instead. Otherwise the log might say "Changing schedule..." but actually that might not be happening.

Another option, move the logic here to see if update is needed and then call the update method. Maybe we don't even need the `_update_existing_schedule` function since the logic is not that extensive.

gitlabform/processors/project/schedules_processor.py
```diff
@@ -42,102 +48,124 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
-            debug(
-                "Deleting variables for pipeline schedule '%s'", schedule["description"]
-            )
+            debug("Deleting variables for pipeline schedule '%s'", schedule.description)
 
             for variable in existing_variables:
-                self.gitlab.delete_pipeline_schedule_variable(
-                    project_and_group, schedule_id, variable.get("key")
-                )
+                schedule.variables.delete(variable.get("key"))
```
amimas, line 121: I know deleting the existing variable was being done before but I wonder if that's really needed. Going through the flow of code in your change, the `_set_schedule_variables` being called only when a pipeline is being created. So, how would there be existing schedule?

What do you think?

gitlabform/processors/project/schedules_processor.py
```diff
@@ -42,102 +48,124 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
+                            "variable_type": variable_data.get("variable_type"),
+                        }
+                    )
+                else:
+                    schedule.variables.create(
+                        {
+                            "key": variable_key,
+                            "value": variable_data.get("value"),
+                        }
+                    )
```
amimas, line 140: Do you think we can just pass the entire `variable_data` to `schedule.variables.create`? Just wondering if that's doable and if it is, maybe it'll simplify the logic here.

amimas (changes requested):

gitlabform/processors/project/schedules_processor.py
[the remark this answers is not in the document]
amimas: Thanks for looking into it.

gitlabform/processors/project/schedules_processor.py
TimKnight-DWP: hmm yeah good point, I will need to have a look and see if there's a way to pass all the data in the update request

amimas: Hi @TimKnight-Opencast . Will you have opportunity to look into this? You're helping to refactor the group_member processor in another PR that's also in progress. Appreciate it btw.

I feel that one is a bit more sensitive/complex than this one and might take a bit of time to review. Maybe we can get this completed and then focus on the other one.

To me this one seems very close to be done. That's why I'm suggesting to finish it first. But if you prefer to come back to this later, that's fine too. Thanks again for everything.

TimKnight-DWP:
@amimas I'll try get back onto this one today 👍

gitlabform/processors/project/schedules_processor.py
TimKnight-Opencast: Hmm yeah even in the current update flow, the old schedule gets deleted and a new one created, so this isn't needed. If I get update working without delete/create will need to call _set_schedule_variables from there

gitlabform/processors/project/schedules_processor.py
TimKnight-Opencast: @amimas - on this now, going to update the tests to verify when updating a schedule the updated schedule has the same id as the existing schedule (i.e. wasn't deleted and then recreated)

gitlabform/processors/project/schedules_processor.py
TimKnight-Opencast: @amimas okay this is all done now - there are definetly problems with the test suite making it hard to add new test cases. They all modify the same project and don't clean up pipelineschedules at the end of each test run, so there's lots of awkward interactions when running the full suite vs running a test individually.

Probably need to raise an issue to fully re-write the tests, I don't have the brain to do it at the moment (spent about 4hrs today trying to unpick what was happening)

gitlabform/processors/project/schedules_processor.py
amimas: > going to update the tests to verify when updating a schedule the updated schedule has the same id as the existing schedule (i.e. wasn't deleted and then recreated)

Awesome! Thank you so much! Looking forward to the update.

gitlabform/processors/project/schedules_processor.py
amimas: > even in the current update flow, the old schedule gets deleted and a new one created

Not sure if we're looking at the same thing but I believe the above happens in a very specific scenario. You'll notice the schedules are first grouped by their descriptions. If there are multiple schedules with the same description, then the above happens (i.e. delete all these schedules and create a new single schedule with given description). Otherwise I believe the delete and create new does not happen right now.

> They all modify the same project and don't clean up pipelineschedules at the end of each test run, so there's lots of awkward interactions when running the full suite vs running a test individually.

Yeah, I know what you're referring to. I think it has both pros and cons. Sometimes that setup is useful and probably makes subsequent tests faster and easier. But sometimes that does get in the way.

You'll notice the tests are using `project` fixture. This is scoped to the entire test class. That's why the same project is available to every test case/function. There's another fixture named `project_for_function`. Sounds like this is probably what you were looking for. It's being used in other test cases where needed.

gitlabform/processors/project/schedules_processor.py
amimas: Let's close this. Thanks for the refactor and updating the tests.

gitlabform/processors/project/schedules_processor.py
amimas: @TimKnight-Opencast - I didn't review the updates to the test cases. Do you want to refactor your test updates using the `project_for_function` fixture if needed?

gitlabform/processors/project/schedules_processor.py
```diff
@@ -29,115 +37,130 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
                 if schedule_ids and len(schedule_ids) == 1:
-                    debug(
-                        "Changing existing pipeline schedule '%s'", schedule_description
-                    )
-                    schedule_id = schedule_ids[0]
-                    self.gitlab.take_ownership(project_and_group, schedule_id)
-                    self.gitlab.update_pipeline_schedule(
-                        project_and_group,
-                        schedule_id,
-                        configuration.get("schedules|" + schedule_description),
```
amimas, line 50: Previously updating an existing schedule had involved taking ownership of that schedule first and then performing the update. In the current code, I don't see the "taking ownership" logic anymore. I can't tell what the impact of this is because we didn't have any particular test case around it and I'm not sure what python-gitlab library does when calling `project.pipelineschedules.update` function. I believe taking ownership of schedule needs to be specifically called but I could be wrong. Could you please look into this?

amimas (changes requested):
@TimKnight-OpenCast - Logic refactor looks good except I'm uncertain about schedule ownership as I explained in a thread below.

Aside from that, I found it took awhile to review because a lot of changes in the test codes since the refactor logic was getting close to be completed when I last reviewed. I don't think it was necessary to change every single test case's fixtures to `project_for_function` or change the scope of `schedules` fixture from `class` to `function`.

As I mentioned before, class level fixture also has some advantage. With latest change, now every test will have to create a new project in gitlab, setup schedules from scratch. This affects the performance of the tests and also setup of some tests. This increases the overall execution of the tests. It would've been easier or less time consuming if existing test codes didn't change. Would you be open to revert your refactor of the tests and maybe only add the new test case? If not, that's okay. Will try to work through it but maybe let's keep it mind for other refactors that you're willing to help out with; keep the scope small 😄 .

One important logic to the processor included in this refactor is to check whether or not an update is necessary. Could you please add a new test case for it? For example: update an existing schedule but capture the log that says no update was needed.

TimKnight-Opencast:
@amimas these all look good comments, I'll try get back to it next week, things have got a bit hectic here

TimKnight-Opencast:
@amimas - reverted the test changes so it's just a subset using project_for_function (i.e. my new tests) and added that test for validating the logs when no changes to an existing schedule are made

gitlabform/processors/project/schedules_processor.py
amimas: @TimKnight-Opencast - latest changes looks good to me. I feel this is the only thread/item outstanding.

I wasn't sure why "take ownership" logic was there. I browsed the older commits and traced the logic being added via [this commit](https://github.com/gitlabform/gitlabform/commit/dd5086b88cdcb642a55585a5ec5d3333eea54998), which seem to have addressed the issue reported in #364 . From one of the comments in that thread, I traced the issue in GitLab's [change in version 14.10.1](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/changelogs/archive-14.md#14101-2022-04-29).

> Prevent maintainers from editing PipelineSchedule (merge request)

I believe we should re-add this logic back, unless we can confirm that GitLab allows maintainers/other users to be able to edit pipeline schedule.

TimKnight-DWP: Will add this logic back in, thanks 👍 
Was thinking I was missing comments somewhere

TimKnight-Opencast: Looks like we still need take ownership: https://docs.gitlab.com/ee/ci/pipelines/schedules.html#take-ownership

TimKnight-Opencast: @amimas re-added

amimas: Thanks for adding the "take ownership" logic back.

I don't think that section of the document actually explains it. The document mentions "how to take ownership" but it doesn't explain why or even the fact that modifying someone else' schedule requires "taking ownership". Or maybe I'm just misunderstanding the doc 😂 .

amimas (approved):
Looks good to me. Thanks again for all the work.

Issue #399, opened 2022-08-22
Add creation of groups and projects


gdubicki:
I have been told by one person that they would use this app if it could automatically create groups and projects.

(This is also the last remaining unimplemented feature mentioned in #6.)

weakcamel:
For me this is not a dealbreaker at all (apparently!), but it would be a nice thing to have, especially for "access" groups (we currently manage project/group access by sharing with project-less groups, e.g. "foo-team", "bar-team" which emulates a role-based access control).

amimas:
Same here.. What @weakcamel  said.

It's not a deal breaker for me because we don't expect to create random projects on a daily or hourly basis. We also created a subgroup where everyone has access to create random projects for their POC. Have a more relaxed config for that subgroup being managed by gitlabform.

I'm also using separate project-less groups as @weakcamel  mentioned. It helps with finer codeowner, approval, etc.

kriko:
Would really-really appretiate this functionality as well!

Looking to implement GitLabForm, to roll out and deploy preconfigured GitLab (sub)groups and projects for a GitOps solution, however the only current limitation seems to be the unability to create groups and projects. Wouldn't want to use the Terraform solution or custom API calls to solve this.

Looking at the code, it seems that both the Groups API ([create_group method](https://github.com/gitlabform/gitlabform/blob/main/gitlabform/gitlab/groups.py#L33)) and Projects API implementations ([create_project method](https://github.com/gitlabform/gitlabform/blob/main/gitlabform/gitlab/projects.py#L29)) fully support the creation of these, but it has not yet been implemented. I could try to make a patch for it, but not sure, if I'd be able to follow the current coding practice and design.

As per configuration goes and since the current definition of GitLab path type objects (groups, subgroups, projects) simply states that each group and project are explicitly defined (we don't have group settings that apply to sub-groups of that group), then it would be easy:

Objects (groups/projects) are created based on their order of appearance (or perhaps it's a good idea to sort them before handling, to get the order of creation correct), thus root groups can be created first, then their subgroups and finally projects nested underneath.

A configuration tag would enable backwards compatible behaviour, so "create_if_not_found: true" would enable the creation of the objects.

Otherwise, having gone through the GitLab API v4 definition for both Groups and Projects, very little is different between creation and update requests, with the exception of a few fields, from my research:

### Groups API - group creation
```
    name: "" # Type: string # Required: yes # The name of the group.
    path: "" # Type: string # Required: yes # The path of the group.
[...]

PR #708, opened 2024-03-14
Feature/create groups and projects

MihkelKillo:
A small addition to automatically create groups, subgroups and projects.

Only usable when using target "ALL_DEFINED" and creates the elements in Gitlab when key "create_if_not_found" with value "true" is present. Doesn't affect any already existing parts when this key is not present (or is not truthy).

Solves issue https://github.com/gitlabform/gitlabform/issues/399 as proposed by user [kriko](https://github.com/kriko)


kriko:
@amimas or @gdubicki - could we request a review and a merge, if it's okay?

kriko:
@amimas or @gdubicki - we have fixed the tests, requesting a review. :)

kriko:
@amimas , can you review please?

kriko (approved):

amimas:
Thanks for the PR. Let's finish the discussions and suggestions in the issue first.

gitlabform/lists/projects.py
```diff
@@ -130,12 +130,27 @@ def _verify_if_projects_exist_and_get_archived_projects(
...
+                    )
         return archived
 
+    def _create_project_if_needed(self, project: str):
+        project_config = self.configuration.get_effective_config_for_project(project)
+        if project_config.get("create_if_not_found", False):
+            path_elements = project.split("/")
+            parent_name, project_name = "/".join(path_elements[:-1]), path_elements[-1]
+            parent_id = self.gitlab.get_group_id_case_insensitive(parent_name)
+            self.gitlab.create_project(
```
TimKnight-DWP, line 148: Should aim to use python gitlab apis rather than the existing GLF implementations

TimKnight-DWP (changes requested):

PR #710, opened 2024-03-18
docs: fix branches keyword in MR approval rules

tmeijn:
It ain't much but it's honest work.


amimas:
Thank you very much for this update.

Issue #689, opened 2024-02-20
Transfer project does not work when gitlabform run with ALL_DEFINED or ALL as the target


amimas:
**Describe the bug**

Version 3.9.0 added a feature for transferring a project from one namespace to another. But it only seems to work when `gitlabform` is run with the project being the direct target. Consider following config sample:

```yaml
# test-config.yml

config_version: 3

projects_and_groups:
  group-bar/project-transfer-test:
    project:
        transfer_from: group-foo/project-transfer-test
```

If we run `gitlabform --config test-config.yml group-bar/project-transfer-test`, the above config is processed successfully and the project gets transferred from `group-foo/project-transfer-test` to `group-bar/project-transfer-test`. The dry-run (`--noop`) also seems to complete without errors.

But, if we run `gitlabform --config test-config.yml ALL_DEFINED` following error is returned:

```
Error: Configuration contains project group-bar/project-transfer-test but it cannot be found in GitLab!
```

 If we run `gitlabform --config test-config.yml ALL` (with or without `--noop`), following message is returned indicating there's nothing to do. Running `gitlabform --config test-config.yml group-bar` (using group name as the target), produces same result.

```
🏗 GitLabForm version: 3.9.0 = the latest stable 😊
>>> Getting ALL groups and projects that I have permission to modify...
:: # of groups to process: 0
:: (# of omitted groups - empty effective config: 2)
:: # of projects to process: 0
:: (# of omitted projects - empty effective config: 1)
:: Nothing to do. 
```

**GitLabForm version**

Output of `gitlabform -V`

🏗 GitLabForm version: 3.9.0 = the latest stable 😊

**GitLab version**

F.e. `16.8.2-ee`

long-wan-ep:
My team is also running into this, I can take a stab at it.

Issue #613, opened 2023-10-12
when running for a single group, its subgroups should be included


exaV:
There is currently a unintuitive behaviour when it comes to running config for a single group.

Consider the following config file:

```yaml
csi/*:
        project_settings:
            allow_merge_on_skipped_pipeline: false

csi/my-subgroup/*
      group_members:
          users:
              myself:
                  access_level: maintainer

csi/my-subgroup/project-a:
            allow_merge_on_skipped_pipeline: true
csi/project-b:
            allow_merge_on_skipped_pipeline: true
```

Here we define some settings for projects inside the top level `csi` group. We also specifically add some users as group members to the group `csi/my-subgroup`. Unfortunately settings for that subgroup are only applied when we run with `ALL_DEFINED` or `ALL` or `csi/my-subgroup` but not when we run for `csi`. Considering the behaviour of `ALL` I find this a bit counter-intuitive.

Example output:
`pipx run gitlabform csi`:

🏗 GitLabForm version: 3.8.0 = the latest stable 😊
>>> Getting requested groups or projects...
:: # of groups to process: 1 <---------------------------------------- subgroup is excluded
:: # of projects to process: 2

`pipx run gitlabform ALL_DEFINED`:
>>> Getting requested groups or projects...
:: # of groups to process: 2 <---------------------------------------- subgroup is included
:: # of projects to process: 2

**What should happen:**
Running `pipx run gitlabform csi` should also update all subgroups of csi.

exaV:
As far as I can tell this could be accomplished by using the https://docs.gitlab.com/ee/api/groups.html#list-a-groups-descendant-groups api after having loaded the group here https://github.com/gitlabform/gitlabform/blob/main/gitlabform/lists/groups.py#L28

We would also have to consider "excluded" groups afterwards.

amimas:
It sounds like a bug and I agree with you about the expected behaviour. I've been using `ALL_DEFINED` so far. Maybe @gdubicki can help confirm the issue or suggest something.

PRs are welcome. @exaV - will you be able to help with this?

amimas:
This came up during the above PR review.

[...]

PR #714, opened 2024-03-19
fix: transfer project not working for ALL_DEFINED and ALL

long-wan-ep:
Fix for https://github.com/gitlabform/gitlabform/issues/689.

Adds logic to check for projects being transferred when the target is set to `ALL` or `ALL_DEFINED`.


gitlabform/lists/projects.py
```diff
@@ -113,6 +117,39 @@ def _get_projects(self, target: str, groups: Groups) -> Projects:
...
+            projects_from_configuration = self.configuration.get_projects()
+
+            for project in projects_from_configuration:
+                try:
+                    project_transfer_source = self.configuration.config[
+                        "projects_and_groups"
+                    ][project]["project"]["transfer_from"]
+                except KeyError:
+                    # ignore projects that don't contain transfer_from key
+                    continue
```
amimas, line 133: Correct me if I'm wrong. I think we can use either `continue` or `pass`, right? I think everywhere else in this repo, `pass` is being used. Maybe follow the pattern for consistency?

gitlabform/lists/projects.py
```diff
@@ -113,6 +117,39 @@ def _get_projects(self, target: str, groups: Groups) -> Projects:
...
+
+                if project_transfer_source:
+                    if project_transfer_source in projects_from_groups:
```
amimas, line 136: I think we don't need the outer condition.

amimas (changes requested):
Thanks @long-wan-ep for the PR. Have couple questions/requests.

Could you please add 2 more test cases that covers the scenario where gitlabform's target is a group or subgroup. It might be an issue for that type of target too, although not sure if this has been caught in the related github issue.

There are some code duplications. For example: checking if it's a project that needs to be transferred, finding project transfer source. Do you think we can reduce the duplication? Maybe move the common code in a function?

long-wan-ep:
Thanks for the feedback, looking into reducing duplication and adding tests.

gitlabform/lists/projects.py
long-wan-ep: I've done some refactoring to reduce code duplication, and this is no longer used.

gitlabform/lists/projects.py
long-wan-ep: This has been refactored.

long-wan-ep:
@amimas I've refactored the code a bit to reduce duplication and added 2 more tests for group and subgroup targets. Ready for review again.

gitlabform/lists/projects.py
```diff
@@ -112,6 +100,21 @@ def _get_projects(self, target: str, groups: Groups) -> Projects:
...
+
+                if project_transfer_source:
+                    if self._get_source_project(project, project_transfer_source):
+                        projects.add_requested([project])
+                    else:
+                        fatal(
+                            f"""Configuration contains project {project} to be transferred from {project_transfer_source}
+                                but the source project cannot be found in GitLab!""",
+                            exit_code=EXIT_INVALID_INPUT,
+                        )
```
amimas, line 117: We have the same logic in `_get_single_project` function. But we're not setting similar log and exit code there. Should it be done in that function too? If not, could you clarify why?

Also, couldn't we just call that function for each projects in the loop?

gitlabform/lists/projects.py
```diff
@@ -130,10 +133,31 @@ def _verify_if_projects_exist_and_get_archived_projects(
                 if project_object["archived"]:
                     archived.append(project_object["path_with_namespace"])
             except NotFoundException:
-                fatal(
-                    f"Configuration contains project {project} but it cannot be found in GitLab!",
-                    exit_code=EXIT_INVALID_INPUT,
+                debug("Could not find '%s'", project)
+                debug(
+                    "Checking if it's a project that needs to be transferred from elsewhere"
```
amimas, line 138: Do you know why we need to do this here? Seems like repeat logic that was already done but maybe I'm missing something.

tests/acceptance/conftest.py
```diff
@@ -99,6 +99,17 @@ def subgroup(gl: Gitlab, group: Group):
         gl.groups.delete(f"{group.full_path}/{subgroup_name}")
 
 
+@pytest.fixture(scope="class")
+def other_subgroup(gl: Gitlab, group: Group):
+    subgroup_name = get_random_name("subgroup")
```
amimas, line 104: Let's add the following comment. Looking at the other fixtures above, it seems duplication is already known. Not sure what the fix is though.

```
# TODO: deduplicate this - it's a copy and paste from the above fixture
```

tests/acceptance/standard/test_transfer_project.py
```diff
@@ -299,3 +299,93 @@ def test__transfer_without_satisfied_prerequisites(
...
+              transfer_from: {project_in_other_subgroup.path_with_namespace}
+        """
+
+        run_gitlabform(config, f"{group.path}/{other_subgroup.path}")
+        projects_in_destination_after_transfer = group.projects.list()
+
+        assert (
+            len(projects_in_destination_after_transfer)
+            == len(projects_in_destination_before_transfer) + 1
+        )
```
amimas, line 391: Trying to understand what this test case is doing. Given we have following:

- top level group `foo`
- a subgroup `foo/bar`
- a project in the subgroup `foo/bar/project-a`

Our config looks like this:

```yaml
projects_and_groups:
  foo/project-a:
    project:
      transfer_from: foo/bar/project-a
```

And we run gitlabform like this: `gitlabform --config <config> foo/bar`. This is where I'm getting confused. Our config doesn't contain any "project" reference from `foo`. We only have `foo/project-a`. So, shouldn't this be run as `gitlabform --config foo`, but this kind of doesn't follow the intention of the test case, I think. Something seems not right. When gitlabform targets a subgroup, it will process projects within that subgroup, right? The `transfer_from` could come from anywhere.

What do you think @long-wan-ep ?

gitlabform/lists/projects.py
amimas: Thanks. Let's close this thread. Will open new discussion if needed.

gitlabform/lists/projects.py
amimas: Thanks for the update.

amimas (changes requested):

gitlabform/lists/projects.py
```diff
@@ -1,5 +1,5 @@
 from typing import Tuple
-from logging import debug
+from logging import debug, info
```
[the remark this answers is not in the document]
long-wan-ep: Oops, was using it for testing. Removed.

gitlabform/lists/projects.py
```diff
@@ -165,3 +189,28 @@ def _get_skipped_projects(self, projects: list) -> list:
...
 
         return skipped
+
+    def _get_project_transfer_source(self, project: str) -> str:
+        try:
+            project_transfer_source = self.configuration.config["projects_and_groups"][
+                project
+            ]["project"]["transfer_from"]
+        except KeyError:
+            return ""
```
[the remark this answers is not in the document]
long-wan-ep: Updated and updated method return type.

gitlabform/lists/projects.py
```diff
@@ -53,24 +53,14 @@ def _get_single_project(self, target: str) -> Projects:
...
-                    debug(
-                        "Found a project '%s' to be transferred to '%s'",
-                        maybe_project["path_with_namespace"],
-                        target,
-                    )
+                source_project = self._get_source_project(
+                    target, project_transfer_source
+                )
+                if source_project:
                     projects.add_requested([target])
```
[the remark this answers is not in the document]
long-wan-ep: I see what you mean about it being disconnected. I've updated the debug messages to be more suited for the methods.

gitlabform/lists/projects.py
```diff
@@ -53,24 +53,14 @@ def _get_single_project(self, target: str) -> Projects:
...
-                        project_transfer_source
-                    )
-                    debug(
-                        "Found a project '%s' to be transferred to '%s'",
-                        maybe_project["path_with_namespace"],
-                        target,
-                    )
+                source_project = self._get_source_project(
+                    target, project_transfer_source
+                )
```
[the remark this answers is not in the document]
long-wan-ep: Renamed them to be more descriptive.

gitlabform/lists/projects.py
long-wan-ep: I moved the second debug message to the _get_project_transfer_source_from_config method, so that it's not duplicated in 2 places. We do need the `except NotFoundException` here since this method is checking if a project exists which it won't for projects being transferred.

gitlabform/lists/projects.py
```diff
@@ -165,3 +189,28 @@ def _get_skipped_projects(self, projects: list) -> list:
...
+            maybe_project = self.gitlab.get_project_case_insensitive(
+                project_transfer_source
+            )
+            debug(
+                "Found a project '%s' to be transferred to '%s'",
+                maybe_project["path_with_namespace"],
+                project,
+            )
+        except NotFoundException:
+            return {}
```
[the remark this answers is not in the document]
long-wan-ep: I think `None` makes sense, updated and updated method return type. The tests are passing.

tests/acceptance/conftest.py
long-wan-ep: Added.

tests/acceptance/standard/test_transfer_project.py
long-wan-ep: Ah I think I misunderstood the intention of targeting a group/subgroup. I thought that if a group/subgroup is targetted, it should process all projects in that group/subgroup + projects being transferred from that group/subgroup. Let me revisit this.

long-wan-ep: Updated the logic for group target so that only projects transferring into the targeted group are processed. Updated the test as well to transfer into a subgroup.

gitlabform/lists/projects.py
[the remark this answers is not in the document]
amimas: Thanks for the update.

gitlabform/lists/projects.py
```diff
@@ -50,27 +50,16 @@ def _get_single_project(self, target: str) -> Projects:
...
-                    debug(
-                        "Found a project '%s' to be transferred to '%s'",
-                        maybe_project["path_with_namespace"],
-                        target,
-                    )
+                source_project = self._find_project_transfer_source_in_gitlab(
+                    project_transfer_source
+                )
+                if source_project:
                     projects.add_requested([target])
```
amimas, line 62: I just learned about python's walrus operator (`:=`). We can re-write the nested `if` conditions like this:

```py
            if (
                project_transfer_source := self._get_project_transfer_source_from_config(
                    target
                )
            ) and self._find_project_transfer_source_in_gitlab(project_transfer_source):
                projects.add_requested([target])
```

What do you think about this? Would it make the code harder to read/understand? If it does, then we can leave it as-is.

gitlabform/lists/projects.py
```diff
@@ -112,6 +99,36 @@ def _get_projects(self, target: str, groups: Groups) -> Projects:
...
+            # in all other cases, we also need to look for projects in the config
+            # that are being transferred to a different namespace
+            for project in projects_from_configuration_not_from_groups:
+                # if the target is a group, and the project is not in the group
+                if (
+                    len(groups.get_effective()) == 1
+                    and target != project.rsplit("/", 1)[0]
+                ):
+                    debug(
+                        "Ignore project '%s', since it's not in target group '%s",
```
amimas, line 112: Should we remove the term `group`? This will be applicable when we do `gitlabform ALL`, right? Not sure if mention of `group` in that scenario gets confusing. Thoughts?

gitlabform/lists/projects.py
```diff
@@ -112,6 +99,36 @@ def _get_projects(self, target: str, groups: Groups) -> Projects:
...
             )
+        else:
+            # in all other cases, we also need to look for projects in the config
+            # that are being transferred to a different namespace
+            for project in projects_from_configuration_not_from_groups:
+                # if the target is a group, and the project is not in the group
+                if (
+                    len(groups.get_effective()) == 1
+                    and target != project.rsplit("/", 1)[0]
+                ):
```
amimas, line 110: What happens when a group contains a subgroup? For example:

```py
>>> project='group1/subgroup1-1/project1'
>>> print(project.rsplit("/", 1)[0])
group1/subgroup1-1
```

When we do `gitlabform group1`, does this mean projects in `subgroup1-1` will not be processed? I feel like it should be processed. What do you think @long-wan-ep .

This does seem like related to issue #613 . Don't want to bloat this PR with multiple issues, but if you agree with what I said above and if it's a simple couple lines of changes, should we just do it? On the other hand, not sure if this is considered as a breaking change. If yes, then maybe we go with what you have.

@gdubicki - would you mind chiming in here if available?

gitlabform/lists/projects.py
```diff
@@ -112,6 +99,36 @@ def _get_projects(self, target: str, groups: Groups) -> Projects:
...
+                        if self._find_project_transfer_source_in_gitlab(
+                            project_transfer_source
+                        ):
+                            projects.add_requested([project])
+                        else:
+                            fatal(
+                                f"""Configuration contains project {project} to be transferred from {project_transfer_source}
+                                    but the source project cannot be found in GitLab!""",
+                                exit_code=EXIT_INVALID_INPUT,
+                            )
```
amimas, line 131: I think we can simplify the code using the walrus operator mentioned above.

gitlabform/lists/projects.py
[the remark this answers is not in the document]
amimas: Thanks for the update.

gitlabform/lists/projects.py
amimas: @long-wan-ep - did you get a chance to look into this?

gitlabform/lists/projects.py
amimas: Okay, let's close this thread. If needed, will open new discussion from latest code.

tests/acceptance/standard/test_transfer_project.py
amimas: Thanks for refactoring the code. I think it makes sense now.

amimas (changes requested):

gitlabform/lists/projects.py
```diff
@@ -165,3 +202,47 @@ def _get_skipped_projects(self, projects: list) -> list:
                 skipped.append(project)
 
         return skipped
+
+    def _get_project_transfer_source_from_config(self, project: str) -> Optional[str]:
```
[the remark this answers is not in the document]
long-wan-ep: I had that before and it fails for python 3.8.

gitlabform/lists/projects.py
long-wan-ep: You're right, we can reuse it here and also the error handling here should be in that method, updated.

gitlabform/lists/projects.py
long-wan-ep: I think we can use that operator though I don't think it can be as simplified as that due to the error handling, updated it.

gitlabform/lists/projects.py
long-wan-ep: This is specifically for group targets, updated the if condition to be more explicit.

gitlabform/lists/projects.py
long-wan-ep: Yes, I believe only projects in the group, not subgroup, are processed. This is the default GitLab API behaviour.

gitlabform/lists/projects.py
long-wan-ep: Updated.

gitlabform/lists/projects.py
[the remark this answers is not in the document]
amimas: oh.. darn! I forgot this project still supports python 3.8 and 3.9 😞

gitlabform/lists/projects.py
amimas: Thanks for clarifying.

gitlabform/lists/projects.py
```diff
@@ -50,27 +50,19 @@ def _get_single_project(self, target: str) -> Projects:
...
                     projects.add_requested([target])
-                except NotFoundException:
-                    # it's a group or a subgroup - ignore it here
-                    pass
+                else:
+                    fatal(
+                        f"""Configuration contains project {target} to be transferred from {project_transfer_source}
+                            but the source project cannot be found in GitLab!""",
+                        exit_code=EXIT_INVALID_INPUT,
+                    )
```
amimas, line 65: Previously/currently if the project's transfer source wasn't found in GitLab, we didn't do a fatal log or set an exit code.

https://github.com/gitlabform/gitlabform/blob/fe8d11e015107ece9291767f29e496405c05228e/gitlabform/lists/projects.py#L60-L73

Do you think that was incorrect? Based on the old code comment in that scenario ("it's a group or a subgroup - ignore it here"), wondering if we should keep it that way. To be honest, that comment is not clear to me.

gitlabform/lists/projects.py
amimas: Thanks for the update.

gitlabform/lists/projects.py
amimas: Thanks for the update.

gitlabform/lists/projects.py
amimas: Thanks for the update.

amimas (commented):
Mostly looks good. Left couple more comments to discuss. Also, could you please add couple more tests to capture the scenarios where:

- project doesn't exist and is not configured to be transferred
- project doesn't exist but configured to be transferred and transfer source project is not found

gitlabform/lists/projects.py
```diff
@@ -130,10 +140,28 @@ def _verify_if_projects_exist_and_get_archived_projects(
...
+                        fatal(
+                            f"""Configuration contains project {project} to be transferred from {project_transfer_source}
+                                but the source project cannot be found in GitLab!""",
+                            exit_code=EXIT_INVALID_INPUT,
+                        )
+                else:
+                    fatal(
+                        f"Configuration contains project {project} but it cannot be found in GitLab!",
+                        exit_code=EXIT_INVALID_INPUT,
+                    )
```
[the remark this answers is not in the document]
long-wan-ep: Sounds good. Thanks.

gitlabform/lists/projects.py
long-wan-ep: I think it was incorrect. Looking at the chain of logic that leads to this method, `_get_single_project` should only run when the target is a project, so I don't think we should assume it's a group or subgroup if it's not found.

long-wan-ep:
> * project doesn't exist and is not configured to be transferred
> * project doesn't exist but configured to be transferred and transfer source project is not found

Added test for "project doesn't exist but configured to be transferred and transfer source project is not found", the `test__ALL_DEFINED` test should cover the other case.

gitlabform/lists/projects.py
amimas: Okay, I think that makes sense. I think this may have been overlooked when the code was refactored after this commit.

https://github.com/gitlabform/gitlabform/blob/486132b7b1233076b5f5a52454b523b846f0e6a1/gitlabform/input.py

gitlabform/lists/projects.py
amimas: > This is the default GitLab API behaviour.

While that is true, role of gitlabform is automation. It processes individual groups and projects individually using corresponding API endpoint anyways, but iterates through list or target projects to process them all or the ones defined. So, I feel the subgroups should also be included. Otherwise it's an unintuitive behaviour. If anyone doesn't want a subgroup to be processed, they can use the `skip_groups` key in the config to control what subgroup to be excluded. Or, maybe an additional CLI flag to indicate all subgroup processing should be excluded.

This feels like a potentially breaking change. So, best not to handle it in this PR.

PR #717, opened 2024-03-30
Replace deprecated code with Python 3.8-compatible one

gdubicki:
as suggested in https://setuptools.pypa.io/en/latest/pkg_resources.html
plus a tiny refactoring (simplify some names)


gdubicki:
It's not really bothering anyone with asking for a review, I created the PR just to make sure that all of the CI checks are passing. :)

PR #719, opened 2024-04-04
runs container as non-root

lkwg82:
Runs the container as user different from root for security reasons.

See https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-2-set-a-user

### DIFF
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -14,3 +14,5 @@ RUN cd gitlabform \
     && apk --purge del build-base
 
 WORKDIR /config
+
+USER 1000 # do not run as root


gdubicki (approved):
Thanks for the contribution, @lkwg82!

gdubicki:
Actually this needed a fix to work, see https://github.com/gitlabform/gitlabform/commit/ef578151f4b6faadd7097221521ff894694a5242 @lkwg82.

gdubicki:
Released in [v3.9.3](https://github.com/gitlabform/gitlabform/releases/tag/v3.9.3).

PR #726, opened 2024-04-09
fix: dry run failure when project transfer is configured along with other additional configs

### LINKED ISSUE #725: Project transfer failing with --noop
**Describe the bug**

The `project: transfer_from` config is failing when run with `--noop`, if there are `project_settings` defined in the config. This is due to the project_settings processor getting a 404 when trying to fetch the settings from GitLab API(since the project doesn't get moved in dry-run).

**GitLabForm version**

Output of `gitlabform -V`

🏗 GitLabForm version: 3.9.3 = the latest stable 😊

**GitLab version**

v16.9.2-ee

long-wan-ep:
Fixes #725.

Project transfer config can fail when run with the `--noop` flag, if a processor's `_print_diff` method makes a GitLab API call for the not yet existing project. This PR adds logic to check for `transfer_from` and makes the API call using the correct project path depending on the situation.

### DIFF
--- a/gitlabform/processors/abstract_processor.py
+++ b/gitlabform/processors/abstract_processor.py
@@ -53,11 +53,11 @@ class AbstractProcessor(ABC):
                 )
                 try:
                     project_transfer_source = configuration["project"]["transfer_from"]
-                    project_or_project_and_group = project_transfer_source
                     verbose(
                         f"""Project {project_or_project_and_group} is configured to be transferred, 
                         diffing config from transfer source project {project_transfer_source}."""
                     )
+                    project_or_project_and_group = project_transfer_source
                 except KeyError:
                     pass
 
--- a/gitlabform/processors/abstract_processor.py
+++ b/gitlabform/processors/abstract_processor.py
@@ -51,7 +51,20 @@ class AbstractProcessor(ABC):
                 verbose(
                     f"Processing section '{self.configuration_name}' in dry-run mode."
                 )
-                self._print_diff(project_or_project_and_group, configuration)
+                try:
+                    project_transfer_source = configuration["project"]["transfer_from"]
+                    project_or_project_and_group = project_transfer_source
+                    verbose(
+                        f"""Project {project_or_project_and_group} is configured to be transferred, 
+                        diffing config from transfer source project {project_transfer_source}."""
+                    )
+                except KeyError:
+                    pass
+
+                self._print_diff(
+                    project_or_project_and_group,
+                    configuration.get(self.configuration_name),
+                )
             else:
                 verbose(f"Processing section '{self.configuration_name}'")
                 if self._can_proceed(project_or_project_and_group, configuration):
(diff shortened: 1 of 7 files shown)


gitlabform/processors/abstract_processor.py
```diff
@@ -51,6 +51,16 @@ def process(
                 verbose(
                     f"Processing section '{self.configuration_name}' in dry-run mode."
                 )
+                try:
+                    project_transfer_source = configuration["project"]["transfer_from"]
+                    project_or_project_and_group = project_transfer_source
+                    verbose(
+                        f"""Project {project_or_project_and_group} is configured to be transferred,
```
amimas, line 58: The value of `project_or_project_and_group` is already modified to the same value as `project_transfer_source`. I think we need to save the original value for logging unless we can refer to it from another variable.

amimas (changes requested):
Latest change looks good. Just one small change needed I think.

gitlabform/processors/abstract_processor.py
amimas: Or... move the log one line above so that the change happens after the log.

long-wan-ep: Moved the re-assignment below the verbose, is that ok?

amimas: I think that should be fine. Thanks for the update.

tests/acceptance/standard/test_transfer_project.py
```diff
@@ -9,6 +9,28 @@
...
+            f"{other_group.path}/{project_for_function.name}"
+        )
+        projects_in_destination_before_transfer = other_group.projects.list()
+
+        config = f"""
+        projects_and_groups:
+          {project_new_path_with_namespace}:
+            project:
+              transfer_from: {project_for_function.path_with_namespace}
+        """
```
amimas, line 25: @long-wan-ep - I forgot about this yesterday. Since this bug only showed up when you attempted to use the transfer feature that contained other sections/settings (i.e. `project_settings`), we should include that in this test case. In fact, I suggest we add all the sections that are currently available via gitlabform (i.e. `branches`, `schedules`, `integrations`, etc.). They can be minimal config but should be present in the config, I think.

amimas (changes requested):

tests/acceptance/standard/test_transfer_project.py
long-wan-ep: I debating if we should include all sections here, it feels kind of odd to be testing all of them in transfer project. Since we move the logic to `abstract_processor`'s `process` method,  wouldn't testing one processor be enough(ie. `project_settings` since it makes a call to gitlab) since they all use the same `process`?

amimas: Hmm... That's a good point. Thanks for the update.

amimas (approved):
Looks good. Thank you for the fix.

PR #729, opened 2024-04-16
fix: error when running with --noop and --output-file parameter and config contains project to be transferred

### LINKED ISSUE #728: Error when output file flag is used with dry-run project transfer
**Describe the bug**

https://github.com/gitlabform/gitlabform/pull/726 introduced a bug where an error will occur when the project transfer config is set and the `--output-file` flag is used together with `--noop`. This is due to `effective_configuration` trying to access configs using the project's  `transfer_from` value instead of its actual name.

**GitLabForm version**

Output of `gitlabform -V`

🏗 GitLabForm version: 3.9.4 = the latest stable 😊

**GitLab version**

v16.9.2-ee

long-wan-ep:
Fixes #728.

Update `abstract_processor.py` to not overwrite `project_or_project_and_group` when determining if a project is being transferred

### DIFF
--- a/tests/acceptance/__init__.py
+++ b/tests/acceptance/__init__.py
@@ -247,7 +247,9 @@ def randomize_case(input: str) -> str:
     return "".join(random.choice((str.upper, str.lower))(char) for char in input)
 
 
-def run_gitlabform(config, target, include_archived_projects=True, noop=False):
+def run_gitlabform(
+    config, target, include_archived_projects=True, noop=False, output_file=None
+):
     # f-strings with """ used as configs have the disadvantage of having indentation in them - let's remove it here
     config = textwrap.dedent(config)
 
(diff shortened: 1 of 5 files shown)


tests/acceptance/__init__.py
```diff
@@ -265,5 +267,6 @@ def run_gitlabform(config, target, include_archived_projects=True, noop=False):
         config_string=config,
         target=target,
         noop=noop,
+        output_file=output_file,
```
amimas, line 270: Can we add a generic test case to validate this. Let's try to assert that the generated file is valid. I guess it should be okay as long as it's a valid yaml file?

amimas (changes requested):

tests/acceptance/__init__.py
amimas: Thanks for adding the additional test.

amimas (approved):
Looks good. Thanks for the fix and tests.

PR #733, opened 2024-04-24
test: add additional merge_request_approvals functional tests

TimKnight-DWP:
- Covers some use-cases in case of regression


TimKnight-DWP:
Ignore this, was checking my branches and realised that this one whilst interesting relies on currently experimental-features so we couldn't go onto develop features ontop. One I'll be keeping an eye on with upcoming GL releases

PR #734, opened 2024-04-24
refactor: use new convenience wrapper and caching

TimKnight-DWP:


TimKnight-DWP:
Little refactor following on from the PR which added the python-gitlab wrapper, to try use the common functions (and caching) where possible.

I think caching on get of group/project is _fine_ in these scenarios but there will be some where we need to most up to date state, so we shouldn't use cached result all the time.

What do you think of naming the functions get_x_by_name_cached ? to make it clearer

gitlabform/processors/project/project_processor.py
```diff
@@ -20,15 +20,15 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
 
             # Check if the project was already transfered (i.e. in previous run) or a project with same path already exists
             try:
-                project_in_config: Project = self.gl.projects.get(
+                project_in_config: Project = self.gl.get_project_by_name(
                     project_path_with_namespace
                 )
```
amimas, line 25: Looking at the usage/code here, wondering if it might be more accurate to rename `get_project_by_name` to `get_project_by_path`. Technically we can have 2 projects with same name in GitLab but the path must be unique. If that make sense, maybe same for `get_group_by_name`?

gitlabform/processors/project/tags_processor.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, gitlab: GitLab, strict: bool):
         self.strict = strict
 
     def _process_configuration(self, project_and_group: str, configuration: dict):
-        project = self.gl.projects.get(id=project_and_group, lazy=True)
+        project = self.gl.get_project_by_name(name=project_and_group, lazy=True)
```
amimas, line 16: I wonder why it needed `lazy=True` while the other processors above didn't need to do that. Why is this special 🤔

amimas (commented):
Thanks for this refactor. It looks good to me. Left couple of minor comments.

Could you add a brief description to the PR and/or edit the title? I'm thinking about what should be in the changelog when it gets released. The changelog will just link to this PR. So, would be good to make it self-explanatory. I guess this PR makes small performance improvements and sets the same stage for future refactors or features?

gitlabform/processors/project/tags_processor.py
amimas: yeah, that makes sense.

TimKnight-DWP:
@gdubicki @amimas - updated following comments

amimas (approved):
Looks good to me. Thanks for updating the doc too.

Issue #732, opened 2024-04-22
Processing group fails if it has more than 20 members


long-wan-ep:
**Describe the bug**

Processing a group with > 20 members is failing with the following messages in the logs:
```
/gitlabform/gitlabform/processors/group/group_members_processor.py:153: UserWarning: Calling a `list()` method without specifying `get_all=True` or `iterator=True` will return a maximum of 20 items. Your query returned 20 of 147 items. See https://python-gitlab.readthedocs.io/en/v4.4.0/api-usage.html#pagination for more details. If this was done intentionally, then this warning can be supressed by adding the argument `get_all=False` to the `list()` call. (python-gitlab: /usr/local/lib/python3.12/site-packages/python_gitlab-4.4.0-py3.12.egg/gitlab/client.py:956)
  debug("Group shared with AFTER: %s", group_being_processed.members.list())
/gitlabform/gitlabform/processors/group/group_members_processor.py:266: UserWarning: Calling a `list()` method without specifying `get_all=True` or `iterator=True` will return a maximum of 20 items. Your query returned 20 of 147 items. See https://python-gitlab.readthedocs.io/en/v4.4.0/api-usage.html#pagination for more details. If this was done intentionally, then this warning can be supressed by adding the argument `get_all=False` to the `list()` call. (python-gitlab: /usr/local/lib/python3.12/site-packages/python_gitlab-4.4.0-py3.12.egg/gitlab/client.py:956)
  members = group.members.list()
Warning: Error occurred while processing group my-group, exception:
409: Member already exists
```

This started happening after upgrading to 3.9.5.

**GitLabForm version**

Output of `gitlabform -V`

🏗 GitLabForm version: 3.9.5 = the latest stable 😊

**GitLab version**

16.9.2-ee

PR #731, opened 2024-04-22
fix: process all schedules and not only the first 20

nudgegoonies:
* process all schedules and not only the first 20
* get_all takes all schedules without using paging

Fixes this problem with our project with 57 schedules:
```
/gitlabform/gitlabform/processors/project/schedules_processor.py:26: UserWarning: Calling a `list()` method without specifying `get_all=True` or `iterator=True` will return a maximum of 20 items. Your query returned 20 of 56 items.
```

PR #735, opened 2024-04-26
fix: iterates through group members

SachinKSingh28:
Fixes https://github.com/gitlabform/gitlabform/issues/732

### DIFF
--- a/gitlabform/processors/group/group_members_processor.py
+++ b/gitlabform/processors/group/group_members_processor.py
@@ -152,7 +152,7 @@ class GroupMembersProcessor(AbstractProcessor):
 
         debug(
             "Group shared with AFTER: %s",
-            group_being_processed.members.list(iterator=True),
+            group_being_processed.members.list(get_all=True),
         )
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


amimas (commented):
Thanks for looking into this bug and the PR. I can't add a comment from the exact line. On line 265, there's call to list group members (`group.members.list()`). Any reason why that isn't updated also or was it just a miss?

Also, are you able to verify if this patch works? Wondering if we should add a special acceptance test for this.

gdubicki:
> Thanks for looking into this bug and the PR.

Thanks @SachinKSingh28! And thanks for the review, @amimas. 😊 

> I can't add a comment from the exact line. On line 265, there's call to list group members (`group.members.list()`). Any reason why that isn't updated also or was it just a miss?

I fixed it. Also changed the calls to be the same like as we did in https://github.com/gitlabform/gitlabform/pull/731.

> Also, are you able to verify if this patch works? Wondering if we should add a special acceptance test for this.

To do it it comprehensively we should do it for all the places where we call `lists()`, and that would be many. :/

I don't this should be done on the gitlabform level, but rather in the library. That's why I created https://github.com/python-gitlab/python-gitlab/issues/2843.

(Note that we have this problem solved in a single place in the code when we are still using our home-grown library - see [here](https://github.com/gitlabform/gitlabform/blob/7ca0b2d36ebd425c6df6f5e0ec4d2f88772ada5e/gitlabform/gitlab/core.py#L147-L178).)

Issue #596, opened 2023-09-04
`test__custom_commit_message` acceptance test intermittently fails


amimas:
The acceptance test  located at `tests/acceptance/standard/test_files.py::TestFiles::test__custom_commit_message` intermittently fails. Examples:

- https://github.com/gitlabform/gitlabform/actions/runs/6031288955/job/16364708681
- https://github.com/gitlabform/gitlabform/actions/runs/6072962483/job/16473939160

```
Processing file 'README.md' in branch 'gitlabform_branch_liqueur'
----------------------------- Captured stderr call -----------------------------
Error: Operation add on file README.md in branch gitlabform_branch_liqueur not permitted, but we don't have a branch protection configuration provided for this branch. Breaking as we cannot unprotect the branch as we would not know how to protect it again. 1
------------------------------ Captured log call -------------------------------
...
=========================== short test summary info ============================
FAILED tests/acceptance/standard/test_files.py::TestFiles::test__custom_commit_message - SystemExit: 1
============== 1 failed, 106 passed, 3 rerun in 255.63s (0:04:15) ==============
```

It'd be nice to stabilize the test. Otherwise processing PRs require extra time.

amimas:
Haven't seen this issue for a long time. Closing this.

PR #744, opened 2024-05-02
test: refactor test_files, test_branches and conftest to support better test encapsulation

TimKnight-DWP:
- Should reduce chances of tests clashing with project/branch names etc


amimas (commented):
Hi @TimKnight-DWP  - Thanks for the PR. I haven't come across or seen any issues with tests clashing. Could you share some details please?

I'm not yet familiar with the `test_file*.py` and where it's used. Probably need some time to take a closer look. But just not sure what is being solved.

If @gdubicki is available, it might be quick review for him.

TimKnight-DWP:
It's one of those infrequent, but annoying to debug issues, as it's next to impossible to reproduce consistently, this change is applying some common strategies I've used to mitigate these sorts of issues

amimas:
Thanks for explaining @TimKnight-DWP . I haven't experienced that, but maybe because I don't actually run the full test suite locally. Usually only run the test cases related to a specific processor. I know there is a known issue with tests randomly failing in GitHub Actions involving committing a readme file to project. See issue #596. Is this PR potentially going to resolve that or have you seen other kinds of test issues?

I briefly skimmed through the changes and I think it's good to merge. All the tests are still passing, which is good.

amimas:
Hi @TimKnight-DWP - looks like the there's a test failure in latest pipeline at the moment.

https://github.com/gitlabform/gitlabform/actions/runs/8925985636/job/24515984764?pr=744

I feel this is a flaky test and probably will pass with a rerun. But I haven't seen this before. Since you're trying at stabilize the tests with this PR, do you want to take a look at the current failure?

gdubicki:
Frankly I do not get why this change would help with test flakiness, but it does make the code less duplicated so I am for at least for that reason. Thank you, @TimKnight-DWP !

gdubicki (approved):

TimKnight-DWP:
@gdubicki my gut feeling is there is some thread/handover going on somewhere, which can result in the yield from `project` fixture not being invoked before the next test class executes, meaning it ends up using the same project and has clashing file contents. Moving to by fixture, a) as you say reduces duplication and b) should hopefully result in a cleaner test execution, especially with projects getting slightly more unique names

TimKnight-DWP:
@amimas if my gut feeling is right than inheritance_break.py will be the same issue, but additionally having now read the test it is not clear exactly what it does, so I will rename and then try fold into a more appropriate test file

TimKnight-DWP:
This may relate to #596 (intermittent errors with README etc)

amimas (approved):
Thanks @TimKnight-DWP again for looking into this. I noticed on several occasions recently where tests are failing in github actions with following error:

```
FAILED tests/acceptance/standard/test_files_protected.py::TestFilesProtected::test__set_file_protected_branches - AssertionError: assert 'Content for all branches' == 'Default README content.'
```

https://github.com/gitlabform/gitlabform/actions/runs/8993858128/job/24706391920?pr=748

Is this what you have been running into? Not sure why this started happening recently. Wouldn't be surprised if this is somehow caused by latest version of GitLab, since that's what we use in our tests.

Anyways, I agree with @gdubicki . I think this PR is addressing some chores related to tests and if this does help with test flakiness, that's even better. Would love to see this being merged. I'll leave it to you to merge.

TimKnight-DWP:
@amimas yes that was the test failure I spotted. I know in GL 16 a lot more actions (like delete) happen asynchronously from my testing re-work on python-gitlab, tldr Deletes get put onto a Queue. So it makes it even harder to tidy up effectively, ideally we'd want to be even more sure we got a new branch everytime, but we're using a random string generator so it's hard to do without making the test object names harder to read by say putting in a UUID

PR #746, opened 2024-05-03
refactor: migrate group settings processor to python-gitlab library 

mahadevan-karthi-dwp:
Use python gitlab lib for group settings. 

Ran existing acceptance tests in standard folder and they pass.

### DIFF
--- a/gitlabform/processors/group/group_settings_processor.py
+++ b/gitlabform/processors/group/group_settings_processor.py
@@ -1,12 +1,30 @@
+from logging import info, debug
+from typing import Dict
+
 from gitlabform.gitlab import GitLab
-from gitlabform.processors.single_entity_processor import SingleEntityProcessor
+from gitlab.v4.objects.groups import Group
+from gitlabform.processors.abstract_processor import AbstractProcessor
 
 
-class GroupSettingsProcessor(SingleEntityProcessor):
+class GroupSettingsProcessor(AbstractProcessor):
     def __init__(self, gitlab: GitLab):
-        super().__init__(
-            "group_settings",
-            gitlab,
-            get_method_name="get_group_settings",
-            edit_method_name="put_group_settings",
-        )
+        super().__init__("group_settings", gitlab)
+
+    def _process_configuration(self, group: str, configuration: Dict):
+        configured_group_settings = configuration.get("group_settings", {})
+
+        gitlab_group: Group = self.gl.get_group_by_path_cached(group)
+
+        if self._needs_update(gitlab_group.asdict(), configured_group_settings):
+            info(f"Updating group settings for group {gitlab_group.name}")
+            self.update_group_settings(gitlab_group, configured_group_settings)
+        else:
+            debug("No update needed for Group Settings")
+
+    @staticmethod
+    def update_group_settings(gitlab_group: Group, group_settings_config: dict):
+        for key in group_settings_config:
+            value = group_settings_config[key]
+            debug(f"Updating setting {key} to value {value}")
+            gitlab_group.__setattr__(key, value)
+            gitlab_group.save()
--- a/gitlabform/gitlab/groups.py
+++ b/gitlabform/gitlab/groups.py
@@ -97,20 +97,3 @@ class GitLabGroups(GitLabCore):
                 project_and_groups_in_group_namespace,
                 key=lambda x: x["path_with_namespace"],
             )
-
-    def get_group_settings(self, project_and_group_name):
-        try:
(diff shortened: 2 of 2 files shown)


gitlabform/processors/group/group_settings_processor.py
```diff
@@ -1,12 +1,30 @@
...
-        )
+        super().__init__("group_settings", gitlab)
+
+    def _process_configuration(self, group: str, configuration: Dict):
+        configured_group_settings = configuration.get("group_settings", {})
+
+        gitlab_group: Group = self.gl.get_group_by_path_cached(group)
+
+        if self._needs_update(gitlab_group.asdict(), configured_group_settings):
+            info("Updating group_settings  '%s'", gitlab_group.name)
```
gdubicki, line 19: Nit: we should use only the f-strings syntax everywhere in the new code. It's the cleanest. :)

gdubicki (approved):
Thank you for the contribution, @mahadevan-karthi-dwp! Just 1 nit-picking comment, you may follow it or not, as you prefer, and then merge!

gitlabform/processors/group/group_settings_processor.py
```diff
@@ -1,12 +1,30 @@
...
+        else:
+            debug("No update needed for Group Settings")
+
+    @staticmethod
+    def update_group_settings(gitlab_group: Group, group_settings_config: dict):
+        for key in group_settings_config:
+            value = group_settings_config[key]
+            debug(f"Updating setting {key} to value {value}")
+            gitlab_group.__setattr__(key, value)
+            gitlab_group.save()
```
amimas, line 30: Instead of going through a loop, is it possible to use the `update()` method for a group and pass the entire config dictionary? I'm assuming this is available from python-gitlab library. Here's an example of how this is done for the "schedules processor".

https://github.com/gitlabform/gitlabform/blob/96a39eee118b6c7e00d6dd1d47d845213876ffb3/gitlabform/processors/project/schedules_processor.py#L86-L103

Would be good to follow a consistent pattern, if this makes sense. Not sure which way has more performance benefit 🤔 .

TimKnight-DWP: I think calling `.update` would be best, _if_ it is available (sometimes py-gitlab docs are a bit confusing as to what is available where)

TimKnight-DWP:
Needs rebase

gitlabform/processors/group/group_settings_processor.py
mahadevan-karthi-dwp: I could not spot `update()` method on project or group. 
unless I see if thats available in api version 6 or 7 of gitlab.

amimas: > sometimes py-gitlab docs are a bit confusing as to what is available where

Oh good. So I'm not the only one 😂 

> hi @amimas I could not spot update() method on project or group.

Thanks for checking @mahadevan-karthi-dwp . I have difficulty following python-gitlab docs. Not sure if I'm mis-reading this, but here it says the "GroupManager" is based on `CRUDMixin`, which  is based on `UpdateMixin` among others. Seems like `update()` method should be available.

https://python-gitlab.readthedocs.io/en/stable/api/gitlab.v4.html#gitlab.v4.objects.GroupManager

I'm still trying to get familiar with python-gitlab. So I could be wrong here. I think the important thing here is to make sure the entire setting/config gets passed to gitlab, which I think your loop does. I'm not sure if it will work if the config included nested dictionary, map, or array. Looking at GitLab's [update group setting api](https://docs.gitlab.com/ee/api/groups.html#update-group), there are couple of attributes that expects array type input (i.e. `unique_project_download_limit_allowlist`, `unique_project_download_limit_alertlist`). Maybe these don't work today and in that case your refactoring won't break anything that currently works. Would you be able to test it out or better yet... add a test case for it?

Anyways, I will leave it to @TimKnight-DWP as I know he's been working with python-gitlab lately. So he may have more insights. Thanks for helping with this refactoring.

gdubicki: > > sometimes py-gitlab docs are a bit confusing as to what is available where
> 
> Oh good. So I'm not the only one 😂
> 
Omg, I’m

TimKnight-DWP: @amimas @mahadevan-karthi-dwp  - I've previously tested this style loop and __set_attrs__ with an array input, which works perfectly, I've never tried it with a nested object/map etc because I've not seen one of those on GitLab

TimKnight-DWP (approved):

PR #750, opened 2024-05-06
Release only when tests pass

gdubicki:
and revamp the way we organize GitHub Actions code.

The "_" YAMLs are the workflows we run for the main branch, the PRs and for releases (tags beginning with "v").

The other YAMLs are being used in the above ones according to "reusable workflows" pattern
(https://docs.github.com/en/actions/using-workflows/reusing-workflows).

It's not perfect, but I think it's easier to comprehend than what we had before.


gdubicki:
Note: it's not tested, because it seems to be impossible without merging it. 😝

.github/workflows/_releases.yml
```diff
@@ -0,0 +1,137 @@
...
+      - name: Build
+        run: |
+          pip install wheel setuptools
+          python setup.py build sdist bdist_wheel
+      - name: Publish distribution 📦 to PyPI
+        uses: pypa/gh-action-pypi-publish@v1.5.0
+        with:
+          password: ${{ secrets.PYPI_API_TOKEN }}
+
+  publish-to-dockerhub:
```
jimisola, line 82: I'm unable to find the publish to dockerhub. There is a login to ghcr on line 122.
I can't see other registries in "Publish image to registries" job.

.github/workflows/linters.yml
```diff
@@ -0,0 +1,33 @@
...
+  workflow_call:
+
+jobs:
+  black-formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Black Code Formatter
+        uses: gdubicki/black-action@main
```
jimisola, line 13: Perhaps better to use the "official" github action?

https://black.readthedocs.io/en/stable/integrations/github_actions.html

gdubicki: Good point, switched to it!

.github/workflows/_releases.yml
gdubicki: True, we use GHCR. Corrected the name.

.github/workflows/_prs.yml
```diff
@@ -0,0 +1,102 @@
...
+      - main
+
+jobs:
+
+  approve:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Approve
+        run: echo For security reasons, all pull requests need to be approved first before running any automated CI.
```
TimKnight-DWP, line 15: @gdubicki 
```
echo For security reasons, all pull requests need to be approved first before running the Premium Acceptance Tests in CI.
```

.github/workflows/_prs.yml
```diff
@@ -0,0 +1,102 @@
...
+          mkdocs build
+
+  linters:
+    uses: ./.github/workflows/linters.yml
+
+  tests-standard:
+    uses: ./.github/workflows/tests-standard.yml
+
+  # you cannot use reusable workflows with an "environment" key,
+  # so we have to copy "tests-premium.yml" here :(
```
TimKnight-DWP, line 59: Damn, this I think would work fine in GitLab CI 😂

TimKnight-DWP (changes requested):

TimKnight-DWP:
@gdubicki - just seen on https://github.com/gitlabform/gitlabform/pull/746, if the Acceptance Tests have not yet completed, BUT we have triggered a Premium run and those pass -> GitHub will do an auto-merge if one has enabled auto-merge.

I was expecting it to wait for everything to be done. 

Worth having a think about how we resolve that scenario? Only trigger Premium tests once all Standard Tests done and the Environment deploy is approved? OR prevent auto-merge whilst any check incomplete?

.github/workflows/_prs.yml
TimKnight-DWP: will resolve in: https://github.com/gitlabform/gitlabform/pull/753

TimKnight-DWP:
> @gdubicki - just seen on #746, if the Acceptance Tests have not yet completed, BUT we have triggered a Premium run and those pass -> GitHub will do an auto-merge if one has enabled auto-merge.
> 
> I was expecting it to wait for everything to be done.
> 
> Worth having a think about how we resolve that scenario? Only trigger Premium tests once all Standard Tests done and the Environment deploy is approved? OR prevent auto-merge whilst any check incomplete?

This is a setting in the "Branch Protections" I am going to set it to require tests to pass

TimKnight-DWP (approved):

PR #751, opened 2024-05-07
feat: manage Project and Group Labels via gitlabform

TimKnight-DWP:


TimKnight-DWP:
I am not yet certain if there is python-gitlab support for Group Labels: https://python-gitlab.readthedocs.io/en/stable/gl_objects/labels.html, I will test, but if it is a non-starter, then I'll raise a PR for that and come back to raise a separate PR here to add Group Label support

TimKnight-DWP:
The bulk of the processor code is identical between ProjectLabels and GroupLabels, I will see if I can refactor

gitlabform/processors/util/labels_processor.py
```diff
@@ -0,0 +1,97 @@
...
+        parent_object_type: str,
+    ):
+        debug(f"{label_name} not in configured labels")
+        # only delete labels when enforce is true, because user's maybe automatically applying labels based
+        # on Repo state, for example: Compliance Framework labels based on language or CI-template status
+        if enforce:
+            info(f"Removing {label_name} from {parent_object_type}")
+            full_label.delete()
+
+    def update_existing_label(
```
amimas, line 68: The parameters of this function is a bit unusual. We're going to process/update a specific label. It's being called in a loop. Does it need all the labels in the config and all the existing label? Feels like the caller should simply pass the label that should be updated and their current config.

gitlabform/processors/util/labels_processor.py
```diff
@@ -0,0 +1,97 @@
...
+        full_label: Union[GroupLabel, ProjectLabel],  # GroupLabel | ProjectLabel
+        label_name: str,
+        needs_update_method: Callable,
+        parent_object_type: str,
+    ):
+        info(f"Updating {label_name} on {parent_object_type}")
+        configured_label = configured_labels.get(label_name)
+        existing_label_names.append(label_name)
+
+        if needs_update_method(full_label.asdict(), configured_label):
```
amimas, line 81: Might be helpful to have logs if update is not needed.

tests/acceptance/standard/test_group_labels.py
```diff
@@ -0,0 +1,125 @@
...
+          {group.full_path}/*:
+            group_labels:
+              test_label:
+                color: red
+                description: this is a label
+        """
+
+        run_gitlabform(config_for_labels, group_for_function)
+
+        updated_group = gl.groups.get(group.id)
```
amimas, line 21: Do we need this? Can't we refer to `group`?

tests/acceptance/standard/test_project_labels.py
```diff
@@ -0,0 +1,173 @@
...
+        assert new_label.description == "this is a label"
+        assert new_label.priority == 2
+
+        # text color gets converted to HexCode by GitLab ie blue -> #0000FF
+        assert new_label.color == "#0000FF"
+
+        # validate same id is being used
+        assert created_label.id == new_label.id
+
+    def test__can_set_project_labels_at_group_level(
```
amimas, line 143: I can't seem to follow or match the logic of the test to what's being described here. Is it trying to test that label can be set for all projects under a group? If yes, I think it's not necessary. That's more like testing if common/inheritance is working.

amimas (commented):
Hey @TimKnight-DWP , this is great. Thanks for looking into it and the PR. Left a few comments. I haven't dealt with labels enough. Can label be deleted if it's still in use by issues or merge requests? I wonder if it would be useful to add an option to not delete label in case it's in use, even if enforce is set. Maybe that's a separate feature.

gitlabform/processors/util/labels_processor.py
```diff
@@ -0,0 +1,97 @@
...
+    def process_labels(
+        self,
+        configured_labels: Dict,
+        enforce: bool,
+        group_or_project: Union[
+            Group, Project
+        ],  # Group | Project -> |: operand not supported in 3.8/3.9
+        needs_update_method: Callable,
+    ):
+        existing_labels = group_or_project.labels.list()
```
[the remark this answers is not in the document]
gdubicki: Yes, for now this needs explicit `get_all=True`, see https://github.com/python-gitlab/python-gitlab/issues/2843.

tests/acceptance/standard/test_project_labels.py
TimKnight-DWP: Yes that sis what it was testing, happy to drop it, but I did want to test inheritance worked as I expected it to 😄

TimKnight-DWP: I was roughly sketching out TDD locally when writing this

amimas: Ahh.. ok. It's probably not needed but doesn't hurt either I think. Will leave it to you to decide. Maybe rename the test case for clarity if you decide to keep it.

gitlabform/processors/util/labels_processor.py
TimKnight-DWP: @amimas - good points, over-enthusiastic method extraction from me, should be this way round so it's more legible

tests/acceptance/standard/test_group_labels.py
TimKnight-DWP: @amimas - we need to get the fresh state of the group after the test run, my personal preference is to assign to a fresh parameter, I prefer to avoid mutation wherever I can to make debugging easier. We could assign it to `group` I'm easy if you would prefer to stick with that convention used elsewhere in the tests

TimKnight-DWP:
> Hey @TimKnight-DWP , this is great. Thanks for looking into it and the PR. Left a few comments. I haven't dealt with labels enough. Can label be deleted if it's still in use by issues or merge requests? I wonder if it would be useful to add an option to not delete label in case it's in use, even if enforce is set. Maybe that's a separate feature.

@amimas my read of: https://docs.gitlab.com/ee/user/project/labels.html#delete-a-label 

Suggests it will delete the label and all references to it (so presumably issues etc reference it via id). It is plausible we want to disambiguate `enforce` into two options 
1. Delete if not in config _and_ not in use on issues etc
2. Always delete

But I don't have a good grasp on a good language for that, I would suggest we leave it as an option for people to request as an enhancement later?
Code-Logic wise it shouldn't be too hard to implement but don't want to over-engineer this solution up front

tests/acceptance/standard/test_group_labels.py
amimas: Ahh... Ok. Thanks for explaining. It's probably fine.

amimas:
I think that's fine @TimKnight-DWP . Like you said, it doesn't need to be over engineered right now. Hopefully the users understand the implication. It's more about knowing gitlab than anything else in my opinion.

tests/acceptance/standard/test_group_labels.py
TimKnight-DWP: (it comes from learned pain when I was debugging a Threaded Java app and it kept mutating variables under me in other threads 😂 )

amimas:
Not on my computer right now and it's too hard to review the changes on mobile. Will try to get to it later unless @gdubicki gets to it first.

BTW - this PR has 100% coverage, I think. Nice work!! 👏 😉

gdubicki (approved):
Thank you for the contribution, @TimKnight-DWP ! 😊 👏

PR #753, opened 2024-05-08
build: automate release process and enforce conventional-commits

TimKnight-DWP:
Relates to: https://github.com/gitlabform/gitlabform/pull/750


TimKnight-DWP:
Links used as a guide when developing:
- https://github.com/google-github-actions/release-please-action
- https://dev.to/natilou/automating-tag-creation-release-and-docker-image-publishing-with-github-actions-49jg

gdubicki (changes requested):
I think that we need to change more for the new release process.

What about the `version` and `tbump.toml` files? The former needs to be replaces with reading the version from the last git tag and the latter probably just removed.

gdubicki (approved):

Issue #673, opened 2024-01-22
Add support for member_role_id


adam-moss:
Add support for `member_role_id` in https://github.com/gitlabform/gitlabform/blob/main/gitlabform/gitlab/members.py

This will allow for the configuration of [custom roles](https://docs.gitlab.com/ee/development/permissions/custom_roles.html)

Thanks.

amimas:
Hey @adam-moss - Shouldn't this be supported already since the `member_role_id` is just another attribute of [group/project members API](https://docs.gitlab.com/ee/api/members.html#edit-a-member-of-a-group-or-project).

I didn't check the code yet. Since gitlabform follows the raw parameter passing pattern, in theory it should already be supported. I'm not able to validate it. Would you be able to try?

amimas:
@adam-moss - did you get a chance to try it out yet?

amimas:
Thank you for verifying. I just took a quick peak at the code for project membership. Looks like "all the data" from the config is not being passed to the API like I had assumed. What's being passed to the API right now is `access_level` and `expires_at`...

https://github.com/gitlabform/gitlabform/blob/f7c592205ac9fe480eacb7340de60987d90541e2/gitlabform/processors/project/members_processor.py#L142-L152

Looking forward to the update. I guess this will only take care of adding an existing "custom role" as member. Maybe we should open a separate feature request to have gitlabform to manage the custom roles?

amimas:
In my opinion, it makes sense that this is available via python-gitlab library. There's an open issue about it too 

https://github.com/python-gitlab/python-gitlab/issues/1357

Of course I can't speak for that library. Hopefully maintainers of the library also agree. Even then not sure if they have the capacity to provide that option.

Don't know if there's another equivalent library. The really last option I think would be to just implement it in gitlabform but my suggestion would be to create a separate project that basically builds the library and then consume it here.

TimKnight-DWP:
@amimas @adam-moss - I'll look at adding `member_role_id` into the relevant places on the REST APIs but if we want/need to support creation, then yeah I think adding initial GraphQL support to python-gitlab would probably be the way to go

PR #759, opened 2024-05-14
feat: add support for supplying member_role_id in project members

TimKnight-DWP:
- Adds basic support for custom roles, by allowing Ultimate users to specify role Ids for their Members
- Relates to #673


TimKnight-DWP:
Can't actually `assert` the change is successful because GitLab don't expose the `member_role_id` in a get/list call. :(

TimKnight-DWP:
**Known bugs**
- Adding a member_role_id which is not present results in a 500
- When testing we must follow the steps to create a custom role at instance level: https://docs.gitlab.com/ee/user/custom_roles.html#create-a-custom-role

**Additional features**
- REST APIs look like they exist for creating Custom roles on both SaaS and Dedicated: https://gitlab.com/groups/gitlab-org/-/epics/11851 (WIP)
  - We would need to document clearly **how** users can create Custom Roles in GL using gitlabform
  - Currently there is no python-gitlab support for `member_roles` or `group/member_roles`

TimKnight-DWP:
I can fudge the test locally to add user with member_role_id, but creating a custom role on my local instance

TimKnight-DWP:
Only available in Ultimate, so I have provided a new action in the workflows to support this, and added a secret with an EE Ultimate License I got recently from GitLab using the handbook process.

TimKnight-DWP:
Not sure how to test this as workflows only get run if they exist on `main` -> the tests run correctly on my machine with the same Ultimate License I put in a secret used to stand up a local container

docs/reference/members.md
```diff
@@ -32,6 +32,26 @@ projects_and_groups:
...
+    * user `access_level` MUST still be supplied and MUST match the base_access_level of the custom role
+
+```yaml
+projects_and_groups:
+  group_1/project_1:
+    members:
+      users:
+        my-user:
+          access_level: maintainer
+          member_role_id: 2
```
adam-moss, line 49: The API expects an id but for usability purpose should we not allow the Name attribute of the role to be used and the id looked up from that?  Could be an issue with .com usage but for self-managed is probably preferable?

.github/workflows/_prs.yml
```diff
@@ -100,3 +100,41 @@ jobs:
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install test dependencies
+        run: |
+          pip install wheel
+          pip install -e .[test]
+      - name: Run acceptance tests that require Ultimate license
+        env:
+          # requested via: https://handbook.gitlab.com/handbook/marketing/developer-relations/contributor-success/community-contributors-workflows/#contributing-to-the-gitlab-enterprise-edition-ee
```
amimas, line 130: I had done something similar when working on a gitlabform feature locally. If I'm not mistaken, this gives a temporary license that will expire soon. For CI purposes, we need something that won't expire or at least not after few weeks. This will potentially block other PRs from merging, unless we mark this as "not required". Maybe that's what we do for now? But if the license does expire, we can't confirm other changes didn't break this feature. We really need to get onto gitlab's open source program and migrate to gitlab.com.

docs/reference/members.md
amimas: I agree. It could be added in a separate PR if it'll make this PR too large.

docs/reference/members.md
```diff
@@ -58,3 +78,25 @@ projects_and_groups:
...
+projects_and_groups:
+group_1/*:
+    group_members:
+      groups:
+        another-group:
+          group_access: no access
+      users:
+        my-user:
+          access_level: owner
+          member_role_id: 2
```
amimas, line 99: In this example, shouldn't this be under the `another_group`?

tests/acceptance/conftest.py
```diff
@@ -86,6 +86,11 @@ def other_group(gl: Gitlab):
         gl.groups.delete(group_name)
 
 
+@pytest.fixture(scope="function")
+def random_string() -> str:
```
amimas, line 90: Why do we need this if we already have another function for random string? We're calling that one in this function anyway.

amimas (approved):
I'm not familiar with member processors as much as I'd like but it seems good to me. Unsure about the gitlab license and PR workflow but that's leading to another discussion, I think.

docs/reference/members.md
TimKnight-DWP: I'll look up how we do it with project names etc

.github/workflows/_prs.yml
TimKnight-DWP: @amimas - agreed that would be the perfect situation, I'll address Adam's comments and also try 

Currently that is a 90/day, 100/user trial which I can repeatedly get extended -> I'll raise a ticket and talk to our Customer Support Manager, hopefully can leverage some of our weight as an Ultimate customer

tests/acceptance/conftest.py
TimKnight-DWP: as far as I could tell, the other random_string function is not exposed on the scope for Tests to use @amimas

docs/reference/members.md
TimKnight-DWP: @amimas -> from reading https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project, member_role_id is only support for a per user basis. I'll have to test and see if you can set a whole group to have a member_role_id

tests/acceptance/conftest.py
TimKnight-DWP: Confirmed trying to use get_random_name directly results in:
```
fixture 'get_random_name' not found
```

docs/reference/members.md
TimKnight-DWP: @adam-moss @amimas I've pushed a commit which should support names as well as Ids, the difficulty for resolving between the two is that GitLab SaaS will use a different API endpoint for custom roles to Self-Managed/Dedicated

.github/workflows/_prs.yml
adam-moss: What I've done previously is use GitLab's open source programme to get Ultimate.  See https://about.gitlab.com/solutions/open-source/join/

It's a trivial process to revalidate once per year although would necessitate hosting on gitlab itself.  Could set up automated mirrors back to here though

gitlabform/gitlab/python_gitlab.py
```diff
@@ -50,3 +51,46 @@ def get_user_by_username_cached(self, username: str) -> RESTObject:
...
+            # Self-Managed & Dedicated
+            path = f"/member_roles"
+
+        return self.http_get(path)
+
+    @functools.lru_cache()
+    def get_member_role_cached(self, name_or_id, group_id: Union[int, None]):
+        member_roles = self.get_member_roles_cached(group_id)
+        for member_role in member_roles:
+            if member_role["id"] == name_or_id or member_role["name"] == name_or_id:
```
adam-moss, line 80: should we consider normalising the case here?

adam-moss (approved):

Issue #764, opened 2024-05-16
GitLab 17.0 Issues


TimKnight-DWP:
**Describe the bug**

Issue to track issues when running GLF vs GitLab 17.0 (release 16/05/24)

**GitLabForm version**

3.11.1

**GitLab version**

17.0.0-ee

TimKnight-DWP:
**test_schedules.py** -> Is now failing on all branches, suggesting something within Schedules has changed in 17.*

amimas:
@TimKnight-DWP - Do we still need to keep this issue open? If not, let's close it. Also, looks like this was specifically around scheduled pipeline related changes, which has been addressed by above PR? If yes, would be good to update the title and description of this issue for future references.

TimKnight-DWP:
@amimas no I think this was fixed?

Closing now

PR #766, opened 2024-05-17
test: update pipeline_schedules tests for GitLab 17

TimKnight-DWP:
- Tests required to be updated due to enforcement fixes in GL 17
- Documentation updated; pre-existing rules on GL API are now being enforced correctly
- relates to #764

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,4 +17,4 @@
 
 Please see <a href="https://gitlabform.github.io/gitlabform/">the project site</a> for more information.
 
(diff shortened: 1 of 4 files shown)


amimas (changes requested):
Overall it looks good. Have a small request to take out the specific version of gitlab reference in the readme.

Thanks for fixing the test. Speaking of which, I feel the commit message and the PR title has the wrong "commit type". This PR is not really fixing gitlabform's functionalities. The scheduled pipeline configuration remains same as before. The PR is simply updating the automated test based on new behaviour from GitLab v17. I feel something like following would be appropriate commit message.

> test(fix): use correct branch reference in assertion of scheduled pipeline test

I used the `test` as the commit type and added `fix` as additional scope for clarity. Although the change includes some docs update, I figured updates to the test is higher priority here.

TimKnight-DWP:
@amimas - great points, have adjusted docs and commit message

amimas (approved):
Thanks for the update. Looks good to me.

PR #774, opened 2024-05-29
docs: Add info about and link to pyenv

jimisola:
Minor update about ability to use pyenv if one's OS does not have a required version of Python.


amimas (approved):
Good idea. This is what I use too for managing different versions of python as needed for r different projects.

TimKnight-DWP (approved):

PR #775, opened 2024-05-30
fix: bug in custom roles implementation

TimKnight-DWP:


TimKnight-DWP:
Won't get back to this until Monday if there are comments 👍

TimKnight-DWP:
Damn these were running locally - sorry out of time for now, `main` is currently also broken :( 

Will come back Mon

TimKnight-DWP:
Annoyingly I cannot reproduce this locally when running a "Standard" GitLab instance

TimKnight-DWP:
Looks like GitHub workflows driven by on_call are not picking up the global environment variable being set :( 
So the tests are still checking out code from main

PR #781, opened 2024-06-04
build(deps): update to codecov@v4 and use secret for token

TimKnight-DWP:
Token will be regenerated and rotated post-merge

### DIFF
--- a/.github/workflows/_main.yml
+++ b/.github/workflows/_main.yml
@@ -21,6 +21,7 @@ jobs:
     uses: ./.github/workflows/tests-standard.yml
     with:
       BRANCH_REF: ${{ github.ref }}
+    secrets: inherit
 
   tests-premium:
     uses: ./.github/workflows/tests-premium.yml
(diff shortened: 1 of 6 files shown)


gdubicki:
Sorry for a late reply.

I obviously applaud the action upgrade but is token really a secret for public repos? I am sure that we made it plain text in the code on purpose, because the docs recommended so back then.

Now it seems to have changed but if it’s a secret, then it won’t be available in forks. Won’t this break some codecov analysis in the CI/CD for pull requests, @TimKnight-DWP?

shoddyknight:
> Sorry for a late reply.
> 
> I obviously applaud the action upgrade but is token really a secret for public repos? I am sure that we made it plain text in the code on purpose, because the docs recommended so back then.
> 
> Now it seems to have changed but if it’s a secret, then it won’t be available in forks. Won’t this break some codecov analysis in the CI/CD for pull requests, @TimKnight-DWP?

Good point about the Forks, I hadn't considered that! @gdubicki

PR #787, opened 2024-06-12
fix: path for member_roles in saas

TimKnight-DWP:
- can't be tested by our test suite as standing up a container counts as self-managed not saas


adam-moss (approved):
lgtm

Issue #754, opened 2024-05-08
Drop python < 3.12


TimKnight-DWP:
3.8 EOL 10-2024
3.9 EOL 10-2025

We will need to drop them anyway by the above dates

If we dropped these earlier we could support `Obj | Obj` typing for example `Group | Project` which is going to become more and more useful as we adopt python-gitlab.

amimas:
> we could support a secondary release branch for hot fixes to the previous major version

Personally I'd avoid multiple version support. It'll bring additional maintenance that I think we don't need or can afford.

gdubicki:
I wanted to share with others who might be reading this issue that we discussed this within the maintainers team and agreed to stop supporting all version of Python except the latest stable.

The benefits would be:
* enabling new syntax for optimal development experience,
* less tests to run - less chance of running into flakiness, lower CI/CD use which could help if we move to GitLab and GitLab CI, lower resource use is better for the environment 🌱
* optimal performance as new Python versions are faster (but it's not a very significant factor in our case),

The downsides:
* some users will have to switch to run the app with Docker if they are stuck with old Python but a) [the switch is easy](https://gitlabform.github.io/gitlabform/installation/#run-without-installing), b) they will have to switch or upgrade Python at some point anyway,
* we may alienate some potential contributors who are stuck with old Python on their dev envs? - not sure if that’s really a realistic scenario, is anyone really using something like Centos 7 or Ubunto 18.04 on a desktop?

If anyone has anything to add here, in particular something that would make us change our minds, please share your thoughts!

PR #789, opened 2024-06-16
chore!: stop supporting Python older than 3.12 

gdubicki:
BREAKING CHANGE: but note that the next older Python
version support drop (when 3.13 stable will be released)
will NOT be treated as a breaking change anymore.

https://github.com/gitlabform/gitlabform/issues/754

### DIFF
--- a/.github/workflows/tests-standard.yml
+++ b/.github/workflows/tests-standard.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [3.8, 3.9, "3.10", 3.11, 3.12]
+        python-version: [3.12]
       fail-fast : false
     steps:
       - uses: actions/checkout@v4
@@ -50,7 +50,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [3.8, 3.9, "3.10", 3.11, 3.12]
+        python-version: [3.12]
       fail-fast : false
     steps:
       - uses: actions/checkout@v4
@@ -103,7 +103,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [3.8, 3.9, "3.10", 3.11, 3.12]
+        python-version: [3.12]
       fail-fast : false
     steps:
       - uses: actions/checkout@v4
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -1,6 +1,6 @@
 # Requirements
 
-* Docker or Python 3.8-3.12
-    * you can use [pyenv](https://github.com/pyenv/pyenv) if your OS does not have a required Python version
+* Python 3.12
+    * you can use Docker image or [pyenv](https://github.com/pyenv/pyenv) if your OS does not have a required Python version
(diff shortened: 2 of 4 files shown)


TimKnight-DWP:
@gdubicki - how do we feel about combining this with the GitLab move?

Make the first distinct release via GitLab be this breaking change, marking a clear boundary between where we are hosting

TimKnight-DWP (approved):

jimisola (approved):

gdubicki:
> @gdubicki - how do we feel about combining this with the GitLab move?
> 
> Make the first distinct release via GitLab be this breaking change, marking a clear boundary between where we are hosting

Sounds good! (I wanted to combine it with implementing https://github.com/gitlabform/gitlabform/wiki/Current-and-proposed-way-of-selecting-targets but I don’t think I will manage to do it in a reasonable time.)

TimKnight-DWP:
@gdubicki cool - I am going to have time this week to look into the GitLab CI and hopefully to synch the MRs/Issues over into GitLab from GitHub

TimKnight-DWP:
Rebasing and will plan a major release @gdubicki @jimisola as discussing in Slack 👍

marcelstoer:
> but note that the next older Python version support drop (when 3.13 stable will be released) will NOT be treated as a breaking change anymore.

@gdubicki could you please share your thoughts on why you choose this pattern?

When I realized the only significant change in 4.0.0 was the Python precondition, I wondered what you'd do about Python 3.13, 3.14 and so on. To me your "stop supporting all version of Python except the latest stable"-policy calls for a never-or-always-breaking pattern.

Side note, will you now need to rename existing "v4" wikis and issues such as https://github.com/gitlabform/gitlabform/wiki/Configuration-Syntax-v4-Draft#current-draft-configuration-syntax-v4 or #331?

gdubicki:
> @gdubicki could you please share your thoughts on why you choose this pattern?
>
> When I realized the only significant change in 4.0.0 was the Python precondition, I wondered what you'd do about Python 3.13, 3.14 and so on. To me your "stop supporting all version of Python except the latest stable"-policy calls for a never-or-always-breaking pattern.

Hey @marcelstoer, the idea was that the change of policy regarding Python is the breaking change itself. However I mostly assumed that when we will release it along with v4, we will have other, obviously breaking changes released in it, like https://github.com/gitlabform/gitlabform/wiki/Current-and-proposed-way-of-selecting-targets.

It didn't turn out like that though, sorry for that. In hindsight, I think we should have only updated the minor or patch version or even just the "post-release" number as we have not really changed the app code here? I think we should discuss this @TimKnight-DWP @amimas @jimisola.

(And the reasoning why the Python version requirement update itself is not a breaking change is the following: 1. it doesn't force you to change the config syntax and the app behavior will not change for the same config, 2. if you run the app using Docker image you won't even have to change anything more than a number in the version tag (`gitlabform:3` -> `gitlabform:4`).)

gdubicki:
> Side note, will you now need to rename existing "v4" wikis and issues such as https://github.com/gitlabform/gitlabform/wiki/Configuration-Syntax-v4-Draft#current-draft-configuration-syntax-v4 or #331?

Thanks, updated!

Issue #455, opened 2022-12-02
Evenly distribute load of scheduled pipelines by variable cron expressions


pho3nixx:
When configuring schedules via GitLabForm for a larger number of projects using wildcards at `projects_and_groups`, all generated pipeline schedules share the exact same cron expression. When these pipelines get triggered, it puts heavy load on the Gitlab-CI runners due to all pipelines running simultaneously.

To distribute the load over time, one could configure every project's pipeline schedule cron pattern individually, which is cumbersome and in part against what GitLabForm strives for: managing multiple projects with minimal work.

Looking at good old Jenkins CI, they offer a variable `H` ("hash") to be used within the cron expression to achieve a random equal distribution of e.g. daily/nightly jobs: https://www.jenkins.io/doc/book/pipeline/syntax/#cron-syntax

It would be very helpful, if the same or a similar mechanism was available in GitLabForm.

gdubicki:
Hi @pho3nixx! Thanks for the suggestion.

A complete implementation of this seems quite difficult to be done right, see https://gitlab.com/gitlab-org/gitlab/-/issues/17799. But perhaps this is also the reason that issue in GitLab is open for more than 5 years now...

So a simpler solution to automate replacing `H`s with a random value from the appropriate range based on a seed from the group/project name (+ maybe the schedule name?) seems not that hard and could be done in GitLabForm.

If you or someone else would provide a working implementation of this feature in a PR, we will definitely review it and if it works, merge it.

mfriedenhagen:
Hi @gdubicki I think my MR is now ready to be reviewed, it would be nice if you could take a look :-)

gdubicki:
Thank you again for contributing your implementation, @mfriedenhagen! 😊 Released in [v3.14.1](https://github.com/gitlabform/gitlabform/releases/tag/v3.14.1) a moment ago. 🎊

PR #800, opened 2024-07-06
feat: enhanced cron expressions

mfriedenhagen:
* Allow an extended syntax for minutes, hours and weekdays in definitions.
* Replace `H(N1-N2)` for minutes and hours with a project stable number between N1 and N2 including both values.
* Specifying letter `H` only is the same as `H(0-59)` for minutes, `H(0-23)` for hours and H(0-6)' for weedays.
* This allows to specify one single cron expression which may be reused in all projects of a group.
* Add some common aliases like '@nightly' etc.

Refs: #455

### DIFF
--- a/docs/reference/pipeline_schedules.md
+++ b/docs/reference/pipeline_schedules.md
@@ -61,3 +61,25 @@ projects_and_groups:
         cron_timezone: "London"
         active: false
 ```
+
+!!! info
+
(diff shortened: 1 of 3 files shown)


gdubicki:
Thank you for the contribution, @mfriedenhagen! LGTM.

PR #801, opened 2024-07-09
fixup! feat: enhanced cron expressions

mfriedenhagen:
* Order is important in _create_schedule_with_variables.
* Cron expression in `entity_config` must be replaced *before* it is decomposed into scheduled data.

Refs: #455


mfriedenhagen:
@gdubicki, sorry about the new PR 🤕 .  I just tried main on our internal instance and it choked because replacing happened after the final POST data was created.

PR #805, opened 2024-07-17
ci: Switch to Trusted Publishing to PyPI

gdubicki:
see https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing

Also use the latest v1.x verison of the pypa/gh-action-pypi-publish action


gdubicki:
I already configured it in PyPI as documented at https://docs.pypi.org/trusted-publishers/adding-a-publisher/

TimKnight-DWP (approved):

Issue #627, opened 2023-10-29
Migrate project's `members` configuration feature to use python-gitlab library


amimas:
Goal of this issue is to migrate the `members` configuration feature to use [python-gitlab library](https://python-gitlab.readthedocs.io/en/stable/index.html) instead of using homegrown API in this project.

Todo:

- Update `gitlabform/processors/project/members_processor.py` to use python-gitlab library

PR #822, opened 2024-09-10
refactor #627 moving members to python-gitlab

Shackelford-Arden:
Moving `members` over for #627

edit: Fixed issue link.


gitlabform/processors/project/members_processor.py
```diff
@@ -101,11 +102,25 @@ def _process_users(
         enforce_members: bool,
         keep_bots: bool,
     ):
+
+        project: Project = self.gl.projects.get(project_and_group)
```
TimKnight-DWP, line 106: is there a reason not using the cached method here?

TimKnight-DWP (changes requested):
One small query - otherwise looks good, especially that none of the service tests needed to be change to support this 👍

gitlabform/processors/project/members_processor.py
Shackelford-Arden: Mostly that I wasn't fully aware of it 😁 I've stepped away for a bit, but will pull in the cached method instead, as that would definitely be better!

Shackelford-Arden: Updated :)

TimKnight-DWP:
I might get chance to merge today - but I won't get chance to make a new release til next week

TimKnight-DWP:
Unlikely to merge today - stuff come up at the day job 👍

TimKnight-DWP:
@Shackelford-Arden had a few incidents to handle last week but I _should_ have capacity today 👍

Shackelford-Arden:
> @Shackelford-Arden had a few incidents to handle last week but I _should_ have capacity today 👍 

No worries 🙂 Thank you!

PR #828, opened 2024-09-23
run gitlabform for a single group and all its subgroups

exaV:
solves https://github.com/gitlabform/gitlabform/issues/613

This feature is disabled by default and can be enabled with the cli flag `--recurse-subgroups` or `-r`.

### DIFF
--- a/gitlabform/__init__.py
+++ b/gitlabform/__init__.py
@@ -55,6 +55,7 @@ class GitLabForm:
         config_string=None,
         noop=False,
         output_file=None,
+        recurse_subgroups=True,
     ):
         if target and config_string:
             # this mode is basically only for testing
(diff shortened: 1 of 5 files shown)


gitlabform/__init__.py
```diff
@@ -55,6 +55,7 @@ def __init__(
         config_string=None,
         noop=False,
         output_file=None,
+        recurse_subgroups=True,
```
TimKnight-DWP, line 58: should default to False?

TimKnight-DWP (changes requested):

gitlabform/__init__.py
exaV: depends on whether you want to make this the default (for cli) in the future. The tests currently pass with either True or False. If you want to go with True in the future (IMO you should) then i would leave it like that.

TimKnight-DWP (approved):

exaV:
Are we waiting for another review? Please let me know if there's anything I can do me to get this merged

TimKnight-DWP:
@exaV nope just waiting for me to remember 😓

PR #830, opened 2024-09-24
feat: added pre protection checks

chrishill-1:
This code allows us to check whether a branch needs unprotecting to make changes or keep protected if no changes are to be made.


gitlabform/processors/project/files_processor.py
```diff
@@ -17,6 +17,25 @@
 
 
 class FilesProcessor(AbstractProcessor):
+    def pre_protection_check(self, project_and_group: str, configuration: dict):
```
TimKnight-DWP, line 20: This makes sense as a first step, try plugging it into the rest of the logic 👍

gitlabform/processors/project/files_processor.py
```diff
@@ -17,6 +17,25 @@
...
+        all_branches = self.gitlab.get_branches(project_and_group)
+        sorted_branches = sorted(all_branches)
+        current_protected_state = sorted_branches["protected"]
+
+        desired_protected_state = True
+        desired_push_access_level = "no one"
+        desired_merge_access_level = "maintainers"
+
+        needs_unprotect = current_protected_state and (
```
TimKnight-DWP, line 29: if this bit of logic could be extracted you might be able to add a new unit test file for files_processor and unit test it. I think this is the chunkiest bit of logic here, so would be good to add some unit tests to help document it and test logic flow makes sense

gitlabform/processors/project/files_processor.py
```diff
@@ -17,6 +17,25 @@
...
+        sorted_branches["push_access_levels"][0]["access_level_description"] != desired_push_access_level or 
+        sorted_branches["merge_access_levels"][0]["access_level_description"] != desired_merge_access_level
+        )
+
+        if needs_unprotect:
```
TimKnight-DWP, line 34: If branch is currently unprotected, then it should also never need unprotecting 🤔

tests/unit/processors/test_files_processor.py
```diff
@@ -0,0 +1,115 @@
...
+                "feature": {
+                    "protected": True,
+                    "push_access_levels": [{"access_level_description": "developers"}],
+                    "merge_access_levels": [{"access_level_description": "developers"}]
+                },
+            }
+        )
+        self.processor = FilesProcessor(self.mock_gitlab, None, True) # Config and strict mode don't affect pre_protection_check
+    
+    def run_tests(self):
```
TimKnight-DWP, line 42: probably not required, pytest _should_ execute everything in this file already right?

tests/unit/processors/test_files_processor.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        Run all tests and print results.
+        """
+        self.test_no_changes_needed_on_protected_branch()
+        self.test_changes_needed_on_protected_branch()
+        self.test_unprotected_branch_should_remain_unprotected()
+        self.test_branch_not_found()
+
+    def assert_equal(self, result, expected, test_name):
```
TimKnight-DWP, line 51: Use native pytest assertions - see `test_branch_protector.py` as an example - tests should actually fail not just print to console

tests/unit/processors/test_files_processor.py
```diff
@@ -0,0 +1,115 @@
...
+        result = self.processor.pre_protection_check(
+            "project/group", 
+            "dev", 
+            desired_protected_state=False,
+            desired_push_access_level="no one",
+            desired_merge_access_level="maintainers"
+        )
+        self.assert_equal(result, False, "test_unprotected_branch_should_remain_unprotected")
+
+    def test_branch_not_found(self):
```
TimKnight-DWP, line 99: Generally these are a good suite of tests for pre_protection_checks, easy to see what is happening and logic tested

TimKnight-DWP (changes requested):
@chrishill-1 few more comments

tests/unit/processors/test_files_processor.py
TimKnight-DWP: Updates look good for Tests @chrishill-1 I'll give it another e2e pass this morning

TimKnight-DWP: Linters (Black and Types) are failing in the checks below, best to get them fixed this morning whilst I do a review

gitlabform/processors/project/files_processor.py
```diff
@@ -17,27 +17,57 @@
...
+
+
     def __init__(self, gitlab: GitLab, config: Configuration, strict: bool):
         super().__init__("files", gitlab)
         self.config = config
         self.strict = strict
         self.branch_protector = BranchProtector(gitlab, strict)
 
-    def _process_configuration(self, project_and_group: str, configuration: dict):
+    def process_configuration(self, project_and_group: str, configuration: dict):
```
TimKnight-DWP, line 57: why has the name of function changed? the initial _ in `_x_y_` denotes a "private" method

gitlabform/processors/project/files_processor.py
```diff
@@ -17,27 +17,57 @@
...
+
+            if needs_unprotect:
+                print(f"Branch '{branch}' needs changes in protection settings.")
+                return True
+            else:
+                print(f"No changes needed on branch '{branch}'.")
+                return False
+
+        except NotFoundException:
+            print(f"Branch '{branch}' not found.")
```
TimKnight-DWP, line 47: Use `debug`, `info` etc from "logging"
`from logging import debug` see projects.py as an example

gitlabform/processors/project/files_processor.py
```diff
@@ -58,6 +97,18 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
                 verbose(f"Processing file '{file}' in branch '{branch}'")
 
+                # Get desired protection state from configuration
+                desired_protected_state = configuration.get("branches|" + branch + "|protected", False)
+                desired_push_access_level = configuration.get("branches|" + branch + "|push_access_level", "no one")
+                desired_merge_access_level = configuration.get("branches|" + branch + "|merge_access_level", "maintainers")
+
+                # Call pre_protection_check to see if branch protection needs changes
+                if not self.pre_protection_check(project_and_group, branch, desired_protected_state, desired_push_access_level, desired_merge_access_level):
+                    debug(f"Branch '{branch}' is already unprotected and should remain so. Skipping changes.")
```
TimKnight-DWP, line 107: @chrishill-1 next thing is to now actually do the logic to not unprotect branch if it doesn't need it 😄

gitlabform/processors/project/files_processor.py
```diff
@@ -17,6 +17,72 @@
...
+            debug(f"Branch '{branch}' requires updates. Unprotecting and applying changes.")
+            modify_protection = True
+        
+        if modify_protection:
+            # Unprotect the branch if necessary
+            self.branch_protector.unprotect_branch(project_and_group, branch)
+
+        try:
+            # Apply the changes to the branch
+            self._process_configuration(project_and_group, branch)
```
TimKnight-DWP, line 79: _process_configuration is called by the abstract interface somewhere upstream so this implementation in process_branch won't end up being called

TimKnight-DWP (changes requested):

TimKnight-DWP:
Paired with @chrishill-1 to look at the code from top to bottom and determine if extra logic was required to check whether a branch needed unprotecting before it got reprotected.

In Files_Processor and Branch_Protector (the onyl scenarios where branches get unprotected in codebase) - there is existing logic which checks if changes have been made and if not, then the branch does not get unprotected / reprotected

TimKnight-DWP (approved):

TimKnight-DWP (changes requested):
request for additional review since we paired

gitlabform/processors/util/branch_protector.py
```diff
@@ -152,9 +149,17 @@ def do_protect_branch(self, requested_configuration, project_and_group, branch):
             if key != "protected"
         }
 
+        # Could be updated to python-gitlab project.protectedbranches.delete(branch_name) and create
```
TimKnight-DWP, line 152: Comment

jackhindleydwp (approved):
nice solution to some weird api behaviour with branch protection :D

Issue #836, opened 2024-09-30
BUG: Adding members who already exist in project


TimKnight-DWP:
**Describe the bug**
GitlabForm attempts to add members who already exist in project, resulting in a 409 error

<img width="967" alt="image" src="https://github.com/user-attachments/assets/a6d7ee99-70e8-4bfc-8426-d6f9eecdc6ec">

```
Adding user 'jeff.turl' who previously was not a member.
[https://gitlab.com:443](https://gitlab.com/) "POST /api/v4/projects/41450720/members HTTP/11" 409 35
Warning: Error occurred while processing project dwp/secure-development/compliance-framework, exception:
409: Member already exists
```

**GitLabForm version**
3.15.0

**GitLab version**

latest-ee

TimKnight-DWP:
`members_processor` is the likely culprit

`get_members_from_project` still uses non python-gitlab method, later we get the project again using python-gitlab helper methods. It probably needs a bit more of a refactor to ensure we get project once, check members using python-gitlab apis, and don't try add ones who already exist.

We only need to change members in a project if they
a) Aren't directly added already
b) Have had their access level changed

amimas:
I'm getting the same error. I thought this might have been fixed in v3.16.0 that included #838, but doesn't look like that fixed it. In case it helps, here's my config:

```yaml
projects_and_groups:
  gitlab-instance-7ee2208b/*:
    group_settings:
      # Some group settings
    group_members:
      enforce: true
      keep_bots: true
      users:
        gitlab-service-bot:
          access_level: owner
    members:
      enforce: true
      keep_bots: true
      groups:
        users-abc:
          group_access: developer
        users-xyz:
          group_access: developer
        # more groups added/shared-with each projects
      users:
        some-bot:
          access_level: developer
    project_settings:
      # config for project settings....
```

I run gitlabform as `gitlabform --config <path-to-above-config> ALL_DEFINED`. Now every project under `gitlab-instance-7ee2208b` group is failing to be processed with following error related to user/member config.

```
Warning: Error occurred while processing project gitlab-instance-7ee2208b/<name-of-project>, exception:
409: Member already exists
```

I suspect implementation was incomplete in #627. Not sure if it's an easy fix. For now, can we revert that and #838? I'm currently on version 3.14.1 and would love to upgrade to get some of the other features/fixes but can't do that because of this issue.

TimKnight-DWP:
[...]

PR #840, opened 2024-10-07
fix: Adding members who already exist in project

TimKnight-DWP:
- Checks for members who already existed were case insensitive
- Closes https://github.com/gitlabform/gitlabform/issues/836


jackhindleydwp (approved):
looks good; addresses a small but damaging bug, and introduces further test coverage the adding of members to a project

Issue #765, opened 2024-05-17
Incorrect order of gitlab API call for`members`, `merge_requests_approval_rules`, `branches` provided for a project


hiaik:
**Describe the bug**

For the following config
```
  "project_group/*":
    merge_requests_approval_rules:
      any:
        name: "Any approver in project"
        approvals_required: 0
        rule_type: any_approver
      enforce: true
    project_settings:
      default_branch: main
    branches:
      main:
        protected: true
        push_access_level: 0
        merge_access_level: 0
        code_owner_approval_required: true
      develop: *default_branch_settings
  "project_group/project":
    members:
      groups:
        group_label_1:
          group_access: 30
        group_label_2:
          group_access: 40
    merge_requests_approval_rules:
      develop:
        approvals_required: 1
        name: "develop branch"
        groups:
          - group_label_1
        protected_branches:
          - develop
      main:
        approvals_required: 1
        name: "main branch"
        groups:
          - group_label_2
        protected_branches:
          - main
    branches:
      main:
        allowed_to_merge:
          - group: group_label_1
      develop:
        allowed_to_merge:
          - group: group_label_2
```

I am getting error:
```
Request url='.../protected_branches?name=develop', method=POST, json='{"allowed_to_merge": [{"group_id": 21305}], "allowed_to_push": [{"user_id": 2633}, {"user_id": 3598}], "code_owner_approval_required": true, "merge_access_level": 30, "push_access_level": 0, "unprotect_access_level": 40}' failed - expected code(s) [200, 201], got code 422 & body: '{"message":["Merge access levels group does not have access to the project"]}'
```

If I leave only 
```
  "project_group/project":
    members:
      groups:
        group_label_1:
          group_access: 30
        group_label_2:
          group_access: 40
```

It runs successfully, and after that I can run entire config successfully.

I assume there is a broken order for gitlab API calls.

**GitLabForm version**
3.10.0

**GitLab version**

16.11.1-ee

amimas:
I wonder if it's a matter of processing the `members` config first (after `project` config). Is it possible some other workflow might require a different order of processing? Not sure.

I have been running gitlabform in 2 stages like you mentioned. First stage in the pipeline only handles members of project/group and other permission related settings (i.e. project visibility, or other access settings) for projects. The second stage in the pipeline does all other settings.

Not sure if this should be classified as bug or feature request :)

hiaik:
Looks like a bug currently from my POV.

[...]

Issue #644, opened 2023-11-20
Change to protected_environment deploy_access_levels is not detected


hertzsprung:
**Describe the bug**

Changing gitlabform protected_environment deploy_access_levels from `user` to `access_level` does nothing (debug logs erroneously state: "myenv of protected_environments in mygroup/myproject doesn't need an update"

Config before:
```
mygroup/myproject:
  protected_environments:
    enforce: true
    myenv:
      deploy_access_levels:
        - user: my-user
```

Config after:
```
mygroup/myproject:
  protected_environments:
    enforce: true
    myenv:
      deploy_access_levels:
        - access_level: developer
```

**GitLabForm version**

3.8.0

**GitLab version**

GitLab Enterprise Edition 16.7.0-pre [65594f41831](https://gitlab.com/gitlab-org/gitlab/-/commits/65594f41831)

amimas:
I assune the gitlab version 16.7 is a typo. As of right now, I think latest version is 16.6. If this is not a typo, is it possible you're working with an unreleased gitlab version that has a regression bug?

amimas:
@hertzsprung - any update available from you on this?

I was able to reproduce this in PR #857 when added new acceptance tests. Please see the description of that PR. It seems to be a **GitLab Issue**. The current work-around I can see is to simply re-run gitlabform again a 2nd time. Or, it might be addressed if the above PR is merged and released with a performance hit.

amimas:
@hertzsprung - The PR mentioned above has been merged and released in v4.0.1. It should be resolved now. Closing this as fixed.

PR #521, opened 2023-03-26
Try to reproduce #510

gdubicki:
#510

PR #857, opened 2024-10-17
fix: project settings failing to apply because membership config is not processed

### LINKED ISSUE #764: GitLab 17.0 Issues

amimas:
There are certain features in GitLab that are related or dependent. For example:

- branch protection can be configured with granular option such as specific user or group is allowed to push or allowed merge.
- environment protection can be configured same way so that selected user or group is allowed to deploy.
- environment protection approval rules can also have similar setup as well
- merge request approval rules can be configured to select specific user or group

All of the above are dependent on membership of the project. The selected user must be a member with appropriate access or the project must be shared with the selected group.

Currently, GitLabForm's processing of various config happens in the order where membership config happens after processing branch protection setting and before processing other settings mentioned above. So, depending on the config used by the user, branch protection will fail because project's membership hasn't been processed yet and selected user/group is not a member of the project to be set in branch protection.

This MR fixes the above issue by moving the processing of project membership to happen earlier. It fixes the issue #765 

While the fix was very easy, the acceptance tests in gitlabform revealed a GitLab issue. When GitLabForm adds a user to a project and then configures other settings (i.e. branch protection, environment protection, merge request approvals), the same user cannot be configured immediately. For branch protection config, GitLab returns an error message about user not being a member of the project. This is same error mentioned in the issue #765. For other settings, there aren't any error but the settings do not configure the user as specified. This seems to match the bug reported in issue #644. @gdubicki had also encountered similar issue in the past in PR #521 (https://github.com/gitlabform/gitlabform/pull/521#discussion_r1155365504)

Because of above mentioned **GitLab issue**, I added a 2 second delay. This is a random number that allowed me to get the tests to pass on my local machine, although occasionally that wasn't enough. This is a performance hit to the users. I'm not sure if we should include this. In reality, this issue is likely going to happen rarely because the easy fix for users is to rerun gitlabform a 2nd time when the membership should already be set by GitLab and those settings will be configured successfully. But, without this I'm not sure how we can get our acceptance tests to pass and validate that gitlabform is working correctly.

resolves #764
related to #644

### DIFF
--- a/gitlabform/processors/multiple_entities_processor.py
+++ b/gitlabform/processors/multiple_entities_processor.py
@@ -70,7 +70,7 @@ class MultipleEntitiesProcessor(AbstractProcessor, metaclass=abc.ABCMeta):
         else:
             enforce = False
 
-        if "members" in configuration and "protected_environments" in configuration:
+        if "members" in configuration and ("protected_environments" or "merge_requests_approval_rules") in configuration:
             # When gitlabform needs to update project membership and also
             # configure environment protection, there seems to be a race condition
             # or delay in GitLab. Automated acceptance tests in gitlabform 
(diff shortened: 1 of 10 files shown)


TimKnight-DWP (approved):
Looks good @amimas

Issue #877, opened 2024-11-12
Introduce new STOP_AT parameter (in addition to START_FROM)


hiaik:
**What:**
We would like to have `STOP_AT` parameter to specify which project should be processed as last (all the rest are skipped).

**Current behaviour:**
Only start-from is supported which is unclear on how to use (what was the purpose of that feature)
```
  -sf START_FROM, --start-from START_FROM
                        start processing projects from the given one (as numbered by "x/y Processing group/project" messages) (default: 1)
```

**Why:**
We need to run gitlabform project for ~600 projects hourly. On gitlabform 4.0.1 version time jumped up (from 40 minutes to 1.5 hour) in pipeline as there was a logic change in how API are executed. It will be more and more in the future.

Idea is to be able to run the same `config.yml` in parallel (in batches):
* gitlabform -c config.yml ${PROJECT_PATH} --start-from 1 --stop-at 100
* gitlabform -c config.yml ${PROJECT_PATH} --start-from 101 --stop-at 200
...
* gitlabform -c config.yml ${PROJECT_PATH} --start-from N --stop-at M

That way we can process all projects in batches via single config file.

Are you ok if we contribute to that feature?

amimas:
Hi @hiaik - Contribution is always welcome. I do have some thoughts before that is continued though. Would love to get feedbacks.

First, let's address the issue with a jump in execution time. This came from the fix that was added via PR #857. There's a 2 second wait time added in branch protection config for each branch when the project also contains member config. Reason for that is due to a GitLab Issue, which is explained in that PR. So in your case, 2 seconds for 600 projects is about 1200 seconds (20 minutes) waiting time. If there were multiple branches being configured, it will take longer. So, that explains the execution time increase for you from 40 minutes to 90 minutes. I just opened another PR #878 that should reduce the time a little bit, but I imagine you'll still see total execution time at around 60 minutes.

I think the GitLab issue mentioned/found in that PR should be addressed. If someone can work with GitLab support and lead the effort on that, it would be best solution in general. That way we don't need to add that wait time unnecessarily. There was a related issue in GitLab, but it was simply closed. Probably didn't come into the radar of appropriate folks.

https://gitlab.com/gitlab-org/gitlab/-/issues/383666#note_1338023609

[...]

PR #878, opened 2024-11-12
perf: reduce execution time of branch processor

amimas:
Previously a fix was added so that members configuration is processed before processing other configurations that are dependent on it. See PR #857 for details.

This revealed an issue in GitLab. After a user is added to a project, if that user is referenced in other processor like branch protection, the processing doesn't take place successfully. In case of branch protection an error is received saying the user is not a member of the project. In other processors, there are no errors but the config is not applied. The work-around for this GitLab issue was to wait a little.

The work-around mentioned above for waiting was added to branch protection inside a loop so that it will wait for every branch config. But that is not necessary. Waiting once at the beginning of branch config processing should be sufficient. This will save a lot of time when processing many projects with multiple branches for those projects.

related to #877

### DIFF
--- a/gitlabform/processors/project/branches_processor.py
+++ b/gitlabform/processors/project/branches_processor.py
@@ -19,22 +19,21 @@ class BranchesProcessor(AbstractProcessor):
         self.strict = strict
 
     def _process_configuration(self, project_and_group: str, configuration: dict):
+        if "members" in configuration:
+            # When gitlabform needs to update project membership and also
+            # configure branch protection, there seems to be a race condition
+            # or delay in GitLab. Automated acceptance tests in gitlabform
+            # creates new user and adds to the project followed by configuring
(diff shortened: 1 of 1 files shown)


TimKnight-DWP (approved):

hiaik:
Thanks guys, do you know when this change makes it into a new version?

amimas:
@hiaik - v4.0.2 is now available.

hiaik:
thanks @amimas 

for 377 projects:
* 3.16.5 - 39 mins
* 4.0.2 - 51 mins (which is +12 mins = 377 * 2 (sleep(2)) = 754 s = ~12.5 mins)

All is how we expected that - thanks for that further improvement from 4.0.1.

Next step for us will be contribute with `STOP_AT` https://github.com/gitlabform/gitlabform/issues/877 to have a proper approach for parallelism.

PR #887, opened 2024-11-21
fix: wildcard based branch protection config fails to apply

### LINKED ISSUE #886: Unexpected behavior when trying to add a protected branch that uses a wildcard.
**Describe the bug**

When attempting to use gitlab-form to configure wildcard-based branch protection (e.g., r-*), the tool fails to apply the protection, even if branches matching the wildcard (e.g., r-feature1, r-test) already exist in the repository. However the GitLab UI allows the creation of wildcard branch protection rules and successfully applies the protection to matching branches. (Even if no matching branches exist.) See the below screenshot as an example.

gitlabissue

The following warning is shown when running gitlab-form:

```
Processing section 'branches'
Warning: Branch '*-some-name-suffix' not found when trying to processing it to be protected/unprotected!
Warning: Branch 'r-*' not found when trying to processing it to be protected/unprotected!
```

This can be reproduced using the following configuration (some of which was pulled directly from the documentation [here](https://gitlabform.github.io/gitlabform/reference/protected_branches/): 
```
config_version: 3
gitlab:
  ssl_verify: true
  timeout: 20
  url: https://gitlab.com

projects_and_groups:
  "group/project1":
    branches:
      'r-*':
        protected: true
        create: true
        allowed_to_merge: 
          - access_level: maintainer
        allowed_to_unprotect:
          - access_level: maintainer
        allowed_to_push:
          - access_level: no access
      '*-some-name-suffix':
        protected: true
        push_access_level: no access
        merge_access_level: developer
```

**GitLabForm version**

`🏗 GitLabForm version:  4.0.3 = the latest stable 😊`

**GitLab version**

`17.6.0-pre`

Please could anyone advise on this? Am I missing anything here? Thanks!

KieranSt97:
This pull request refactors some of the logic to do with protected branches to fix a bug introduced in `4.0.2` of gitlab form. This bug is specifically to do with trying to protect a branch in gitlab using a wildcard. 

It adds a new acceptance test case to cover the new code. 

Closes #886

### DIFF
--- a/gitlabform/processors/project/branches_processor.py
+++ b/gitlabform/processors/project/branches_processor.py
@@ -73,9 +73,7 @@ class BranchesProcessor(AbstractProcessor):
         elif protected_branch and not branch_config.get("protected"):
             self.unprotect_branch(protected_branch)
 
-    def protect_branch(
-        self, project: Project, branch: str, branch_config: dict
-    ):
+    def protect_branch(self, project: Project, branch: str, branch_config: dict):
         """
         Protect a branch using given config.
         Raise exception if running in strict mode.
@@ -86,9 +84,7 @@ class BranchesProcessor(AbstractProcessor):
         try:
             project.protectedbranches.create({"name": branch, **branch_config})
         except GitlabCreateError as e:
-            message = (
-                f"Protecting branch '{branch}' failed! Error '{e.error_message}"
-            )
+            message = f"Protecting branch '{branch}' failed! Error '{e.error_message}"
 
             if self.strict:
                 fatal(
--- a/gitlabform/processors/project/branches_processor.py
+++ b/gitlabform/processors/project/branches_processor.py
@@ -1,4 +1,3 @@
-import re
 from typing import Optional
 from logging import debug, info
 from cli_ui import warning, fatal
@@ -138,4 +137,4 @@ class BranchesProcessor(AbstractProcessor):
 
     @staticmethod
     def is_branch_name_wildcard(branch):
-        return re.fullmatch(".*\\*.*", branch)
+        return "*" in branch
--- a/gitlabform/processors/project/branches_processor.py
+++ b/gitlabform/processors/project/branches_processor.py
@@ -73,18 +73,20 @@ class BranchesProcessor(AbstractProcessor):
         elif protected_branch and not branch_config.get("protected"):
             self.unprotect_branch(protected_branch)
 
-    def protect_branch(self, project: Project, branch: str, branch_config: dict):
+    def protect_branch(self, project: Project, branch_name: str, branch_config: dict):
         """
         Protect a branch using given config.
         Raise exception if running in strict mode.
         """
 
-        debug("Setting branch '%s' as protected", branch)
+        debug("Setting branch '%s' as protected", branch_name)
 
         try:
-            project.protectedbranches.create({"name": branch, **branch_config})
+            project.protectedbranches.create({"name": branch_name, **branch_config})
         except GitlabCreateError as e:
-            message = f"Protecting branch '{branch}' failed! Error '{e.error_message}"
+            message = (
+                f"Protecting branch '{branch_name}' failed! Error '{e.error_message}"
+            )
 
             if self.strict:
                 fatal(
--- a/gitlabform/processors/project/branches_processor.py
+++ b/gitlabform/processors/project/branches_processor.py
@@ -36,11 +36,11 @@ class BranchesProcessor(AbstractProcessor):
         """
         protected_branch: Optional[ProjectProtectedBranch] = None
 
-        if not self.check_for_wildcard(branch_name):
+        if not self.is_branch_name_wildcard(branch_name):
             try:
                 branch = project.branches.get(branch_name)
             except GitlabGetError:
-                message = f"Branch '{branch_name}' not found when trying to processing it to be protected/unprotected!"
+                message = f"Branch '{branch_name}' not found when processing it to be protected/unprotected!"
                 if self.strict:
                     fatal(
                         message,
@@ -135,5 +135,5 @@ class BranchesProcessor(AbstractProcessor):
         return branch_config
 
     @staticmethod
-    def check_for_wildcard(branch):
+    def is_branch_name_wildcard(branch):
         return re.fullmatch(".*\\*.*", branch)
--- a/gitlabform/processors/project/branches_processor.py
+++ b/gitlabform/processors/project/branches_processor.py
@@ -1,7 +1,8 @@
+import re
 from typing import Optional
 from logging import debug, info
 from cli_ui import warning, fatal
-from gitlab.v4.objects import Project, ProjectBranch, ProjectProtectedBranch
+from gitlab.v4.objects import Project, ProjectProtectedBranch
 from gitlab import (
     GitlabGetError,
     GitlabDeleteError,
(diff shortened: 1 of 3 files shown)


gitlabform/processors/project/branches_processor.py
```diff
@@ -33,28 +34,28 @@ def process_branch_protection(
...
-                    message,
-                    exit_code=EXIT_INVALID_INPUT,
-                )
-            else:
-                warning(message)
+        if not self.check_for_wildcard(branch_name):
+            try:
+                branch = project.branches.get(branch_name)
+            except GitlabGetError:
+                message = f"Branch '{branch_name}' not found when trying to processing it to be protected/unprotected!"
```
amimas, line 43:
```suggestion
                message = f"Branch '{branch_name}' not found when processing it to be protected/unprotected!"
```

gitlabform/processors/project/branches_processor.py
```diff
@@ -66,28 +67,24 @@ def process_branch_protection(
...
             elif not protected_branch:
-                self.protect_branch(project, branch, branch_config)
+                self.protect_branch(project, branch_name, branch_config)
         elif protected_branch and not branch_config.get("protected"):
             self.unprotect_branch(protected_branch)
 
-    def protect_branch(
-        self, project: Project, branch: ProjectBranch, branch_config: dict
-    ):
+    def protect_branch(self, project: Project, branch: str, branch_config: dict):
```
amimas, line 76: Since you've changed the method signature, (i.e. `branch` is now a string instead of type `ProjectBranch`), this is causing the current test failure for `files` proessor.

https://github.com/gitlabform/gitlabform/actions/runs/11958905461/job/33339482178?pr=887

If this should be a branch name, you'll need to update the `files_process.py` as well and pass the branch name instead of the branch object. I think that should fix test failure.

May I suggest renaming `branch` to `branch_name`? I feel that's a bit clearer what is expected.

gitlabform/processors/project/branches_processor.py
```diff
@@ -136,3 +133,7 @@ def transform_branch_config(self, branch_config: dict):
                             item["group_id"] = self.gl.get_group_id(item.pop("group"))
 
         return branch_config
+
+    @staticmethod
+    def check_for_wildcard(branch):
```
amimas, line 138: Minor comment. The method name seems (to me) a little odd. It doesn't indicate what should happen in this method or if it will return something. Based on the method logic, since it may return a match or `None` maybe it can be something like `get_wildcard_branch_name`. Don't think that make sense either. Could call it `is_branch_name_wildcard` or something better and and in the logic, return boolean only.

gitlabform/processors/project/branches_processor.py
```diff
@@ -136,3 +133,7 @@ def transform_branch_config(self, branch_config: dict):
...
+        return re.fullmatch(".*\\*.*", branch)
```
amimas, line 139: I'm not too familiar with this. Could you help me understand how it helps to validate if the string contains wildcard or not? Is there a sample or explanation somewhere that I could read?

tests/acceptance/standard/test_branches.py
```diff
@@ -60,6 +60,36 @@ def test__protect_and_unprotect(self, project, branch):
...
+            merge_access_user_ids,
+            _,
+            _,
+            unprotect_access_level,
+        ) = get_only_branch_access_levels(project, branch_wildcard)
+        assert push_access_levels == [AccessLevel.NO_ACCESS.value]
+        assert merge_access_levels == [AccessLevel.MAINTAINER.value]
+        assert push_access_user_ids == []
+        assert merge_access_user_ids == []
+        assert unprotect_access_level is AccessLevel.MAINTAINER.value
```
amimas, line 91: What do you think of this. After `run_gitlabform`, create a branch that matches the wildcard. Then get the access config of that branch to see if it's been protected or not. Or, do you think that shouldn't be necessary?

amimas (changes requested):
Thanks a lot for this PR.

gitlabform/processors/project/branches_processor.py
KieranSt97: This is just a simple regex match to check if there is a `*` in the branch name. 

I used the examples in the gitlab docs [here](https://docs.gitlab.com/ee/user/project/repository/branches/protected.html#protect-multiple-branches-with-wildcard-rules) to test this 

```
*-stable
production/*
*gitlab*
```

When trying to cut a branch in gitlab containing a `*` you get the error `Branch name cannot contain '*'`

Do you think I should add some comments around this?

tests/acceptance/standard/test_branches.py
KieranSt97: Yes I think thats probably worth doing, I'll try and give that a go

gitlabform/processors/project/branches_processor.py
KieranSt97: I've made the above changes

gitlabform/processors/project/branches_processor.py
KieranSt97: I believe this may now be sorted, could you take another look?

tests/acceptance/standard/test_branches.py
KieranSt97: This has now been added

gitlabform/processors/project/branches_processor.py
amimas: Nice!

gitlabform/processors/project/branches_processor.py
amimas: Sounds good. Thanks for the update.

gitlabform/processors/project/branches_processor.py
amimas: > When trying to cut a branch in gitlab containing a * you get the error Branch name cannot contain '*'

I believe that's a different scenario. A branch **name** cannot be `*`. But, you can created branch protection **rule** using `*`, which I believe is going to target every branch in the repo.

image

amimas: > This is just a simple regex match to check if there is a * in the branch name.

If we just want to check for the presence of wildcard characters in the name, then we can do this: `return '*' in branch_name or '?' in branch_name`. Your logic currently doesn't check for the `?` character, which seems to be accepted by GitLab when creating a branch protection rule.

image

Maybe we don't need to use regular expression? I think that might be helpful if we want to validate the wildcard pattern itself, but that logic should match how gitlab validates a wildcard pattern. Maybe we don't need to go that far? What do you think?

tests/acceptance/standard/test_branches.py
amimas: Thanks. Let's close this thread.

gitlabform/processors/project/branches_processor.py
KieranSt97: I believe the code caters for this scenario as the branch name still contains a `*` so the regex returns true and still falls down the correct code path, does that make sense? I'm struggling to think of a valid test scenario

KieranSt97: In this scenario what does the '?' do? I can't find reference to this anywhere in the gitlab documentation. Its also not possible to create a branch with a '?' in the **name** so I'm unsure how this fits in.

I see your point about not having to use a regex match though. I've pushed a change for this. 

I think the code currently caters for a protected branch that is just called '*' but I can't think of a valid test scenario to add for this.

amimas: > Do you have a link to any documentation about using '?'

https://www.geeksforgeeks.org/wildcard-pattern-matching/

Does that help?

KieranSt97: That does help, I think I understand how the '?' works in wildcard pattern matching but I can't get it to work inside gitlab at all. I definitely could be doing something wrong here. I have a repo here that has a currently open branch called `r-1-3-0` should the below protection not match that branch? 

image

amimas: I think `?-*` should match `r-1-3-0`. At least it does when I try it here...

https://www.digitalocean.com/community/tools/glob?comments=true&glob=%3F-%2A&matches=false&tests=r-1-3-0

But I'm not sure what gitlab does under the hood. Their doc shows simple example like you mentioned earlier and there's no reference to their wildcard pattern matching. At least I didn't find it. Is it possible that they don't actually support `?` in the pattern? 🤷 . Maybe we just don't add the validation of `?` in the test. Clearly GitLab allows creating protection rule using `?` and so we keep the logic to make sure gitlabform allows it.

KieranSt97: Yeah I'm really not sure to be honest, I've added the logic into the code to be able to cater for it and all the tests are still passing.

amimas: Sounds good. Thanks for all the work and tests on this PR.

amimas (approved):

Issue #891, opened 2024-11-25
Base64 encoded contents returned by file API not raw contents leading to unnecessary commits


alex-lmco:
The files API, see here: https://python-gitlab.readthedocs.io/en/stable/gl_objects/projects.html#project-files

Returns the contents from the gitlab api which is Base64 encoded, this line:

https://github.com/gitlabform/gitlabform/blob/f96efbe89ad4ae707a5d55ff476be825b8bb16d2/gitlabform/processors/project/files_processor.py#L141

Need to first decode() it and then convert it to a string for the not equals comparison here:

https://github.com/gitlabform/gitlabform/blob/f96efbe89ad4ae707a5d55ff476be825b8bb16d2/gitlabform/processors/project/files_processor.py#L143

For more visual results, you end up comparing base64 to non-base64 (which will always fail) and therefore make empty commits, like this:

Changing file '.gitlab/renovate.json' in branch 'alpha', delta:
**- ewogICAgIiRzY2hlbWEiOiAiaHR0cHM6L...**
+ {
+     "$schema": "https://docs.renovatebot.com/renovate-...

amimas:
Thanks for reporting @alex-lmco . Looks like you've done a lot of troubleshooting work to identify the root cause. Could you help contribute the fix with a test case? The repo must be missing a test case resulting in this bug being added when the files processor was recently refactored.

alex-lmco:
Unfortunately I can't contribute the fix. It is probably though just: `repo_file.decode().decode(sys.stdout.encoding)`

TimKnight-DWP:
I'll pick this up today - affecting us wide spread

PR #892, opened 2024-11-26
fix: reduce unnecessary commits when modifying files

TimKnight-DWP:
- resolves: https://github.com/gitlabform/gitlabform/issues/891
- decodes files from base64 prior to comparing


adam-moss (approved):
lgtm

Issue #142, opened 2020-11-21
Add group-level hooks and push rules (and default branch protection example)


gdubicki:
It's better to provide group-level configs whenever you want some settings to apply to all projects within a group or groups.

They:
* apply to all the new projects by default,
* are faster to process with this app as you do it in a single place,
* generate less internal clutter in your GitLab instance,

For hooks and push rules we would need to actually implement the API but it's most probably copy-paste from project-level versions.

For default branch protection it's a matter of providing an example in the config, see https://docs.gitlab.com/ee/api/groups.html#update-group , `default_branch_protection` attribute.

lfvjimisola:
Additional info from @gdubicki :

you can probably use the SingleEntityProcessor and add it with just a few lines of code, like in 

https://github.com/gdubicki/gitlabform/blob/fc7a3de0f8d16152257553c85b95d3077e9d2ee1/gitlabform/processors/project/project_push_rules_processor.py#L5-L13

```
class ProjectPushRulesProcessor(SingleEntityProcessor):
    def __init__(self, gitlab: GitLab):
        super().__init__(
            "project_push_rules",
            gitlab,
```

lfvjimisola:
1. Will `enforce` be available in order to remove group level webhooks?
2. Is it just to set an empty string to empty a group push rule?

amimas:
For group level webhooks, `enforce` should be supported along with `delete` key for individual webhook. They are supported for project level webhooks.

For "push rule", `enforce` most likely doesn't make sense. I think it's more appropriate to add support for `delete` key. It's not available in project level push rule config right now either.

> Is it just to set an empty string to empty a group push rule?

That would be the workaround right now.

jimisola:
In the feature we should limits issues so that they are atomic in sense of features. This issue handles two features which I believe that one (group push rules) has now been implemented.

amimas:
Agree. Implementations should be done as atomic as possible. It does require some administrative work from the maintainers too (i.e. grooming issues, labeling them, creating separate issues, enforcing it in incoming contributions without demotivating them, etc.). It can easily overwhelm you when you're spending all your personal time and effort into the project 😄 . But, I completely agree with you.

PR #911, opened 2025-01-06
feat: add support for group push rules

szEvEz:
Partially implements https://github.com/gitlabform/gitlabform/issues/142


gitlabform/gitlab/group_push_rules.py
```diff
@@ -0,0 +1,25 @@
+from gitlabform.gitlab.groups import GitLabGroups
+
+class GitLabGroupPushRules(GitLabGroups):
+    def get_group_push_rules(self, project_and_group_name: str):
+        gid: str = self._get_group_id(project_and_group_name)
+        return self._make_requests_to_api(
```
TimKnight-DWP, line 6: By preference we now like to use `python-gitlab` directly rather than "make_requests_to_api" as it provides a cleaner interfac

tests/acceptance/premium/test_group_push_rules.py
```diff
@@ -0,0 +1,46 @@
...
+
+        run_gitlabform(config, group)
+
+        push_rules = group.pushrules.get()
+        assert push_rules.commit_message_regex == ""
+        assert push_rules.member_check is False
+        assert push_rules.commit_committer_check is True
+        assert push_rules.commit_committer_name_check is True
+
+    def test__edit_group_push_rules(self, group):
```
TimKnight-DWP, line 27: Tests look good 👍

TimKnight-DWP (changes requested):

gitlabform/gitlab/group_push_rules.py
szEvEz: Thanks, I've updated accordingly.

TimKnight-DWP (approved):

szEvEz:
Not sure why the tests are failing tho, as I have working tests locally

```
$ py.test tests/acceptance/ -k "TestGroupPushRules"
============================================================================================= test session starts ==============================================================================================
platform darwin -- Python 3.13.1, pytest-8.3.4, pluggy-1.5.0
configfile: pytest.ini
plugins: cov-6.0.0, rerunfailures-15.0, typeguard-4.3.0
collected 222 items / 220 deselected / 2 selected                                                                                                                                                              

tests/acceptance/premium/test_group_push_rules.py ..
```

amimas:
Would be nice to include support for `enforce` key.

szEvEz:
> Would be nice to include support for `enforce` key.

Could you elaborate on that a bit please? 🙂

amimas:
> > Would be nice to include support for `enforce` key.
> 
> Could you elaborate on that a bit please? 🙂

Hi @szEvEz - I mis-understood the feature. After going through feature in GitLab's docs, I realized we don't actually add multiple rules, even though they call it push rule**s**. In the case of multiple push rules, the `enforce` key would've made sense.

I think in this case, adding support for `delete` key might make sense. Imagine you want to delete the rule that was previously configured. I haven't had a chance to review the code yet, but didn't see it being mentioned in the docs. So, I'm guessing it's not implemented. I suppose it could be a separate feature later since existing feature at the project level push rule also doesn't seem to support it.

docs/reference/push_rules.md
```diff
@@ -4,6 +4,24 @@
...
+
+This section purpose is to manage the [group push rules](https://docs.gitlab.com/ee/user/project/repository/push_rules.html).
+
+The settings should be as documented at GitLab's [Push rules section of the Groups API docs](https://docs.gitlab.com/ee/api/group_push_rules.html#add-push-rules-to-a-group), **except the id**.
+
+Example:
+
+```yaml
+projects_and_groups:
+  group_1/project_1:
```
amimas, line 17: It should be 

```suggestion
  group_1/*:
```

tests/acceptance/premium/test_group_push_rules.py
```diff
@@ -0,0 +1,46 @@
...
+        self.test__create_group_push_rules(group)
```
amimas, line 28: Don't need to do this. You can delete this line.

The `group` parameter is a pytest fixture. If you check this fixture in `conftest.py`, you'll notice that this fixture is set to `class` scope. That means, this group is available for the entire duration of the `TestGroupPushRules` class. As a result, when the first/previous test case runs, it already created/configured the group push rule for you. That's why you don't need to call the previous test case again.

Hope that makes sense.

amimas (changes requested):
Left my findings of the test failure. That should fix the issue. Not sure why it's working for you though.

tests/acceptance/premium/test_group_push_rules.py
```diff
@@ -0,0 +1,46 @@
...
+              commit_committer_check: false
+              commit_committer_name_check: false
+        """
+
+        run_gitlabform(config, group)
+
+        push_rules = group.pushrules.get()
+        assert push_rules.commit_message_regex == ""
+        assert push_rules.member_check is True
+        assert push_rules.commit_committer_check is False
```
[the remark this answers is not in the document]
szEvEz: Thanks for digging deeper into this. Funny enough, I can not reproduce this. 

```sh
curl -X GET --header "PRIVATE-TOKEN: ${GITLAB_API_TOKEN}" ${GITLAB_API_URL}/groups/2511/push_rule" | jq .

{
  "id": 10027,
  "created_at": "2025-01-07T09:33:39.829Z",
  "commit_message_regex": "(.|\\s)*\\S(.|\\s)*",
  "commit_message_negative_regex": "",
  "branch_name_regex": "",
  "author_email_regex": "",
  "file_name_regex": "",
  "deny_delete_tag": false,
  "member_check": false,
  "prevent_secrets": false,
  "max_file_size": 0,
  "commit_committer_check": false,
  "commit_committer_name_check": false,
  "reject_unsigned_commits": null,
  "reject_non_dco_commits": null
}
```

`commit_committer_check` is explicitly set to false for me. 

But maybe that's currently just a 'me' issue. I will update the test accordingly.

amimas (approved):
Looks good.

TimKnight-DWP (approved):

Issue #701, opened 2024-03-06
Being able to set merge request approvals on group level


rogierlommers:
Hey! First of all, I like your project. We're playing a bit with our gitlab configurations and how to apply it within the organization. The documentation is quite clear for us, but there is one thing I don't really understand.

<img width="517" alt="Screenshot 2024-03-06 at 14 50 11" src="https://github.com/gitlabform/gitlabform/assets/3907131/b6540580-8647-47d2-a871-92f03d0a205e">

We want to set the request-approval-rules on a group basis. This setting should be inherited to all the groups/projects that will be created inside this group. is this possible?

With the code snippet below i can "walk through" all repositories and change the settings after creation. However, I want to avoid that new projects/repos have the setting `prevent approval by author` to value `false`.

```
---
config_version: 3
gitlab:

projects_and_groups:
  '*':
    merge_requests_approvals:
      reset_approvals_on_push: false
      merge_requests_author_approval: true
      disable_overriding_approvers_per_merge_request: true
```

rogierlommers:
A bit more information: it turns out that the gitlab API does support these operations:

[EDIT]

```
curl --header "PRIVATE-TOKEN: xxx" --header "Content-Type: application/json" -X PUT  --data  <<SEE BELOW>>'' https://gitlab/api/v4/groups/25459/merge_request_approval_setting
```

REQUEST BODY
{
   "require_password_to_approve":false,
   "allow_author_approval":false,
   "allow_overrides_to_approver_list_per_merge_request":false,
   "require_password_to_approve":false,
   "retain_approvals_on_push":false,
   "selective_code_owner_removals":false,
   "allow_committer_approval":false
}

amimas:
Hi 👋 . Based on your latest comment, I think you want to manage "merge request approval" setting in a group. Please confirm. If that's the case, I believe this is not supported by gitlabform at the moment. You'll find the [gitlabform docs on this setting](https://gitlabform.github.io/gitlabform/reference/merge_requests/) specifying that the available config is for a project level setting.

> With the code snippet below i can "walk through" all repositories and change the settings after creation. However, I want to avoid that new projects/repos have the setting `prevent approval by author` to value `false`.

To me, the above is a bit confusing. Even if you configure the setting in a group manually (without gitlabform), it will be inherited by all projects under that group. I'm not aware of any option to control the setting to not be applied on newly created projects after this setting is configured.

[...]

PR #914, opened 2025-01-08
feat: add support for group level merge request approval rules and approval settings

szEvEz:
Implements https://github.com/gitlabform/gitlabform/issues/701

I definitely need to add more tests for the group approval rules.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ lint = [
 # Documentation generation stack.
 docs = [
     "markdown==3.10.3",
-    "zensical==0.0.57",
+    "zensical==0.0.58",
     "pygments==2.21.0",
     "pymdown-extensions==11.0.2",
 ]
--- a/uv.lock
+++ b/uv.lock
@@ -662,13 +662,13 @@ dev = [
     { name = "twine", specifier = "==7.0.0" },
     { name = "types-requests", specifier = "==2.33.0.20260712" },
     { name = "xkcdpass", specifier = "==1.30.0" },
-    { name = "zensical", specifier = "==0.0.57" },
+    { name = "zensical", specifier = "==0.0.58" },
 ]
 docs = [
     { name = "markdown", specifier = "==3.10.3" },
     { name = "pygments", specifier = "==2.21.0" },
     { name = "pymdown-extensions", specifier = "==11.0.2" },
-    { name = "zensical", specifier = "==0.0.57" },
+    { name = "zensical", specifier = "==0.0.58" },
 ]
 lint = [
     { name = "bandit", specifier = "==1.9.4" },
@@ -2156,7 +2156,7 @@ wheels = [
 
 [[package]]
 name = "zensical"
-version = "0.0.57"
+version = "0.0.58"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "click" },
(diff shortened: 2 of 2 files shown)


amimas:
@szEvEz - Thanks a lot for looking to add the new feature.

Last year we have done a lot of refactoring and moving towards using the python-gitlab library. Personally, I'd prefer that we don't grow the homegrown API code in gitlabform.

Might be a coincidence... I kinda started to think about it recently. Unfortunately the python-gitlab library doesn't have support for managing group level merge request approval rules/settings. That's why I started looking at adding the support in that library.

https://github.com/python-gitlab/python-gitlab/pull/3078

Can we wait till the above PR is merged and released? I see you've already put a lot of work in this PR but if we can get this implemented using the python-gitlab library, it'll save us from having to rework/refactor this feature again.

szEvEz:
Hey @amimas,

ah that's actually funny. I've not seen your PR that adds this feature to python-gitlab. For me it's fine if we wait till that gets merged and then I can reimplement this using the library 🙂 

Can you give me a ping once the upstream PR is merged and released?

gitlabform/gitlab/group_merge_requests_approvals.py
```diff
@@ -0,0 +1,83 @@
...
+    def edit_group_approval_rule(
+        self,
+        project_and_group_name,
+        rule_in_gitlab,
+        rule_in_config,
+    ):
+        gid = self._get_group_id(project_and_group_name)
+        approval_rule_id = rule_in_gitlab["id"]
+
+        self._make_requests_to_api(
```
TimKnight-DWP, line 66: Can this be done using python-gitlab please?

TimKnight-DWP (changes requested):

gitlabform/gitlab/group_merge_requests_approvals.py
szEvEz: Yes, definitely. Waiting for https://github.com/python-gitlab/python-gitlab/pull/3078 and will reimplement.

amimas: @szEvEz - The feature in python-gitlab has been merged and they have released it as well. Latest change in `main` branch has updated to the python-gitlab version that includes that feature. Please have a look at this PR and refactor it when you get a chance. I think it'll simplify the implementation.

amimas:
I've updated the PR title. I don't think there's a setting or API for group level "MR approval setting", is there? I think it's only "MR approval rule".

szEvEz:
> I've updated the PR title. I don't think there's a setting or API for group level "MR approval setting", is there? I think it's only "MR approval rule".

I thought that would be https://docs.gitlab.com/17.8/ee/api/merge_request_approval_settings.html, or am I mistaken?

amimas:
Huh... Interesting. You're not mistaken. I just wasn't aware of that. I don't think the `python-gitlab` has support for that "group level MR approval setting" though. Please double check. The feature added was for "group level MR approval rules" only.

So, this PR probably can just implement the approval rules related work. The approval settings will need to be available via python-gitlab.

szEvEz:
> Huh... Interesting. You're not mistaken. I just wasn't aware of that. I don't think the `python-gitlab` has support for that "group level MR approval setting" though. Please double check. The feature added was for "group level MR approval rules" only.
> 
> So, this PR probably can just implement the approval rules related work. The approval settings will need to be available via python-gitlab.

Yeah, there is no support from python-gitlab side yet, that's why I went with the internal implementation there.

setup.py
```diff
@@ -48,8 +48,8 @@ def get_version_file_path():
         "MarkupSafe==3.0.2",
         "mergedeep==1.3.4",
         "packaging==24.2",
-        "python-gitlab==5.3.1",
-        "python-gitlab[graphql]==5.3.1",
+        "python-gitlab==5.6.0",
+        "python-gitlab[graphql]==5.6.0",
```
amimas, line 52: I believe these updates are already done in `main` branch. Let's undo this change.

tests/acceptance/premium/test_group_merge_request_approval_rules.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+
+        run_gitlabform(config, group)
+        rules = group.approval_rules.list()
+        assert len(rules) == 1
+
+        for rule in rules:
+            assert rule.name == "Any Approver"
+            assert rule.approvals_required == 5
+            assert rule.rule_type == "regular"
```
amimas, line 28: Just a minor comment or question. I'm not sure the for loop is giving any advantage. Even if we add more than 1 rules in the config, the assertions have hard-coded value. So, it won't be useful, right? Maybe let's just go with `rule[0].... = ...`?

amimas (commented):
@szEvEz - Overall it looks pretty good to me. That's why don't want to block this PR. I left couple of comments. Please have a look when you get a chance.

After this is merged, will you have time to help migrating the group level approval settings to python-gitlab? It will require a bit extra work since it involves adding the feature into that library first.

setup.py
szEvEz: Done, merged main into my branch

tests/acceptance/premium/test_group_merge_request_approval_rules.py
szEvEz: That's a good hint, I think the code there was actually meant for the multiple rules part. Removed the loop and added some additonal asserts.

szEvEz:
> @szEvEz - Overall it looks pretty good to me. That's why don't want to block this PR. I left couple of comments. Please have a look when you get a chance.
> 
> After this is merged, will you have time to help migrating the group level approval settings to python-gitlab? It will require a bit extra work since it involves adding the feature into that library first.

I'd love to help with the migration to python-gitlab. I am a bit limited on time in the next 3 weeks, but after that I am happy to help :)

amimas (approved):
Thanks again for working on this.

amimas:
@szEvEz - Could you please take a look at the test failure?

szEvEz:
> @szEvEz - Could you please take a look at the test failure?

When I run the tests locally against my own instance, the tests work as expected.

I think this could be related to a feature flag, which needs to be enabled on Gitlab Instance level. (https://archives.docs.gitlab.com/17.8/ee/api/merge_request_approvals.html#group-level-mr-approvals)

The gitlab api documentation also states the following

> On GitLab Self-Managed, by default this feature is not available. To make it available, an administrator can [enable the feature flag](https://archives.docs.gitlab.com/17.8/ee/administration/feature_flags.html) named approval_group_rules. On GitLab.com and GitLab Dedicated, this feature is not available. This feature is not ready for production use.

PR #915, opened 2025-01-09
refactor: use graphql to get member_roles to reduce permission level required

TimKnight-DWP:
- graph ql GET only needs Guest+ permissions
- REST requires significantly higher permission sets which not all users maybe comfortable running with
- better error handling for missing Users

### DIFF
--- a/gitlabform/gitlab/__init__.py
+++ b/gitlabform/gitlab/__init__.py
@@ -2,6 +2,8 @@ import enum
 
 from typing import List
 
+from gitlab import GraphQL
+
 from gitlabform.gitlab.commits import GitLabCommits
 from gitlabform.gitlab.group_badges import GitLabGroupBadges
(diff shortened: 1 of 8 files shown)


TimKnight-DWP:
Not sure what's going on with Coverage Patch 🤔

amimas:
@TimKnight-DWP - This is great! The graphql option will also enable adding features that are not available via REST API. I must've missed it in the code but how would we access the graphql instance from python-gitlab? For example: in any of the processor, currently we do `self.gl.<python-gitlab-object>` for REST API based actions. What would we do that in a processor? Should there be something like `self.gl_graphql`?

amimas (approved):
Thanks for the update @TimKnight-DWP . I wonder if the PR title/commit type should be `refactor` instead of `feat`. Maybe additional scope with the `refactor` type for more clarity but can't think of a good suggestion right now.

TimKnight-DWP:
@amimas happy to do that on squash

PR #918, opened 2025-01-13
fix: remove caching on get_user_id_cached

TimKnight-DWP:
- the get user by username is already cached
- adds extra logging to help users debug effectively

### DIFF
--- a/gitlabform/processors/group/group_members_processor.py
+++ b/gitlabform/processors/group/group_members_processor.py
@@ -289,6 +289,7 @@ class GroupMembersProcessor(AbstractProcessor):
                     # be completely deleted from gitlab.com
                     message = f"Could not find User '{user}' on the Instance so can not remove User"
                     error(message)
+                    continue
 
                 try:
                     group.members.delete(user_id)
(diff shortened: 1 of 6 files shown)


gitlabform/processors/group/group_members_processor.py
```diff
@@ -289,8 +295,8 @@ def _process_users(
                 try:
                     group.members.delete(user_id)
                 except GitlabDeleteError as delete_error:
-                    warning(f"Member '{user}' could not be deleted: {delete_error}")
-                    pass
+                    error(f"Member '{user}' could not be deleted: {delete_error}")
+                    raise delete_error
```
SachinKSingh28, line 299: just double checking if this will have any side effects with processing as we are changing the behaviour from passing to raising exception.

SachinKSingh28: Looks good to me otherwise.

TimKnight-DWP: @SachinKSingh28 yes it will but I think rather than hiding the error it should be raised - that is consistent with the Project Members processor

SachinKSingh28:
> The other thought I had whilst walking my dog, was to remove throwing an exception down here and change the function to return either User or None (and percolate that change up), so the callers have to handle None case in whatever way is appropriate for their functionality.
> 
> Is probably easier to trace and debug too

Yep makes sense.

PR #922, opened 2025-01-15
feat: add project security settings support

szEvEz:


gitlabform/gitlab/project_security_settings.py
```diff
@@ -0,0 +1,22 @@
+from gitlabform.gitlab.core import GitLabCore
+
+
+class GitlabProjectSecuritySettings(GitLabCore):
+    def get_project_security_settings(self, project_and_group_name):
+        security_settings = self._make_requests_to_api(
```
TimKnight-DWP, line 6: Does python-gitlab support Project Security settings? It would be preferable to use that API where possible thanks

TimKnight-DWP (changes requested):

gitlabform/gitlab/project_security_settings.py
adam-moss: @TimKnight-DWP from looking at the docs I don't believe python-gitlab supports this currently so my recommendation would be we accept this as it offers an immediate user benefit (for those on the Ultimate tier, anyway).

TimKnight-DWP (approved):

adam-moss:
@szEvEz can you address the test failures please so we can progress this?  Thanks!

TimKnight-DWP:
`test_group_members_users.py::TestGroupMembersUsers::test__not_remove_users_with_enforce_false` is consistently failing on all branches - I'm investigating locally on https://github.com/gitlabform/gitlabform/pull/969 - will try get it fixed so we can unblock this and others

PR #923, opened 2025-01-16
refactor: migrate project push rules processor to python-gitlab

amimas:
This PR refactors the processor used for managing project level push rules. The methods for managing this resource is removed and switched to use python-gitlab.


TimKnight-DWP:
Are you telepathic @amimas 😂 - I was going to do this today as I'd spotted a bug where our Email REGEX was not being updated so figured step 1 was make it python-gitlab

TimKnight-DWP (approved):

PR #926, opened 2025-01-16
refactor: remove deprecated config syntax for project merge request approval settings

amimas:
TBD


TimKnight-DWP (approved):

TimKnight-DWP:
@amimas - agreed this **should not** be a breaking change as discussed in slack, we planned this for V4

However "refactor" in semantic-version parlance would normally be associated with a Patch version. I feel like for clarity it probably deserves to be a Minor version 🤔

gdubicki (changes requested):
Frankly, I think that we lost the opportunity to apply this change in v4 and should postpone this removal to v5. [It has always been very important to this project to not break anyone's work in an unexpected way](https://gitlabform.github.io/gitlabform/#stability) and I think it would be doing that if we would release it sooner.

TimKnight-DWP:
I guess a question is @gdubicki - do we want to plan for V5 release in say Feb, get some of the current in flight Feature MRs through and then release v5?

I'm personally of the opinion that people shouldn't shy away from frequent major versions if there's a good reason for it and the migration path (especially since v3-4 was dropping python support) is simply laid out?

amimas:
Good points from both of you. I also don't see any issues with frequent major version release if an incoming change requires it. Otherwise, we'll need a separate branching and release strategy that will collect/accumulate all the breaking changes together for release - which invites issues related to clean commit history or merge conflicts. Don't think we need all that complexity for this project.

gdubicki:
I think we should also uptick the version of the config in this release and document how to upgrade like we did for v2->v3 [here](https://gitlabform.github.io/gitlabform/upgrade/#from-2-to-3).

amimas:
> I think we should also uptick [the version of the expected config](https://github.com/gitlabform/gitlabform/blob/8f5174dc4232a4664de949fb3f3265e0fa82b71b/gitlabform/configuration/core.py#L55-L55) in this release and document how to upgrade like we did for v2->v3 [here](https://gitlabform.github.io/gitlabform/upgrade/#from-2-to-3).

Good point @gdubicki . Will try to come back to it later.

TimKnight-DWP:
@amimas did we want to progress this one again?

amimas:
> @amimas did we want to progress this one again?

Hey @TimKnight-DWP - Thanks for the ping! Yeah, I'd love go move this along. As @gdubicki suggested, we should probably move config version from v3 to v4 to drop support for older config syntax. That depends on when we plan on doing the next major release.

Let me know what your thoughts are. If there are other things that are good candidate for a major release, maybe we can move this along.

TimKnight-DWP:
Yeah - to be honest I'm a proponent of "be happy to major release regularly", maybe we get some of the relatively big PRs going on right now in, have a couple of week "stability window" and then be in a place to consider a major release? @amimas

amimas:
> Shall we target v5 for this one?

Yes, let's do that. I'll try and get this PR up-to-date and complete the outstanding tasks. The config version will change - FYI.

PR #930, opened 2025-01-21
chore: update base image to Alpine 3.21

rroesch1:
* Python images based on Alpine 3.18 are unmaintained (last update: 2024-04-09)
* Thus we should switch to images based on Alpine 3.21

### DIFF
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -1,6 +1,4 @@
-FROM python:3.12-alpine3.18
-COPY --from=pandoc/core:2 /usr/bin/pandoc* /usr/bin/
-COPY --from=pandoc/core:2 /usr/lib/libffi* /usr/lib/
+FROM python:3.12-alpine3.21
 RUN apk add --no-cache \
     gmp \
     lua5.3 \
--- a/dev/tests.Dockerfile
+++ b/dev/tests.Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.12-alpine3.18
+FROM python:3.12-alpine3.21
 
 COPY ../setup.py /gitlabform/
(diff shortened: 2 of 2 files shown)


dev/Dockerfile
```diff
@@ -1,6 +1,4 @@
-FROM python:3.12-alpine3.18
-COPY --from=pandoc/core:2 /usr/bin/pandoc* /usr/bin/
```
rroesch1, line 2: fyi: i removed this line because there is no file matching `/usr/bin/pandoc*` in `pandoc/core:2`

amimas: Thanks for looking into this. I'm actually curious about the original intention. Looks like the goal was to include pandoc binaries into gitlabform's image. But I wonder why.

@gdubicki - Do you happen to remember about this setup? I tried to browse the history of this file and went back all the way to when it was introduced. But, unfortunately I didn't find the answer as to why pandoc was being included.

gdubicki: I think that without pandoc the app would not install as we have run it in `setup.py`? If the image builds **and** the app runs from it, then we don’t need it!

rroesch1: @amimas i tested it locally and it works for me

gdubicki: Sorry, now I noticed that it’s the image for running local dev in Docker, so the main question is if https://github.com/gitlabform/gitlabform/blob/main/docs/contrib/local_development.md#running-unit-tests-in-a-docker-container works. Could you please check that, @rroesch1 ?

gdubicki (approved):

gdubicki:
Thank you for your contribution, @rroesch1! 😊

PR #931, opened 2025-01-23
feat: Add custom project settings topics and its test

### LINKED ISSUE #925: Implement add/update/delete project topics
Implement so that projects topics can be added, updated and deleted for projects.

Project Topics are set using [Edit Project](https://docs.gitlab.com/ee/api/projects.html#edit-a-project) and its `topics` attribute.

jimisola:
Initial work on adding support for add/delete/update of project topics.

Closes #925


jimisola:
@amimas Need help with a few things:

- [  ] self._needs_update takes a Dict (used code from project_push_rules) but topics is a List.
- [  ] do I have to add something somewhere for project_topics to be available as a key in the YAML file or is it parsed automagically?
- [  ]  project_topics should support enforce and delete, I looked at BadgesProcessor but it 

a) does not use python-gitlab it seems but MultipleEntitiesProcessor
b) topics are part of projects.settings and does not have their own API endpoint which badges has

jimisola:
The YAML syntax is still to be decided on. Does gitlabform have any other project setting the we have update/delete/enforce for? I couldn't find it. Others such as [project badges](https://gitlabform.github.io/gitlabform/reference/badges/#project-badges) are slightly different and have their own API endpoint.

Shall one be able to specify deletion of topic(s) while also specifying topics that should exist?

```yaml
projects_and_groups:
  group_1/project_1:
    project_topics:
      topics:
        - topicA
        - topicB
      enforce: true # optional
```

What if you also want to delete topics?

```yaml
projects_and_groups:
  group_1/project_1:
    project_topics:
      topics:
        - topicA
        - topicB
            delete: true
```

jimisola:
Implementation, tests and docs. Tests pass and I think that I covered all the use-cases.

amimas:
I'm actually on the fence about this. Managing topics for a project is already supported via `project_settings` config. Introducing a custom processor/config section (i.e. `project_topics`) just for configuring topics seems to complicate things in my opinion. Having 2 different ways of managing topics can be confusing and also can lead to issues such as configuration conflicts, maintaining, fixing bugs. For example: which one should we use: `project_settings.topics` or `project_topics`? What happens if both are used? In case of issues, fix bugs in multiple locations, maybe.

The existing feature allows topics to be added to a project. Because of the way GitLab's edit project API works, this means overwriting existing set to topics with whatever is sent to the API endpoint. For example, the below config sample means only the configured topics will exist for the project.

```yaml
projects_and_groups:
  "test-group/test-project":
    project_settings:
      visibility: internal
      default_branch: main
      auto_devops_enabled: false
      topics:
        - topic1
        - topic2
```

If we want to delete topics, the config would be this:

```yaml
projects_and_groups:
  "test-group/test-project":
    project_settings:
      visibility: internal
      default_branch: main
      auto_devops_enabled: false
      topics: []
```

I recall from our slack chat that the proposed config/processor in this PR came up because sometimes topics are being managed outside of gitlabform. In that case, couldn't this external processor (i.e. script) update the gitlabform config file accordingly to reflect the state of project topics? Then when gitlabform runs, existing setup will work fine without requiring this extra "feature"?

If we still need "extra" ability for managing topics, I think that should be done as part of the existing processor/config section. So, that would mean adding the logic under `project_settings.topics` instead of introducing a new config section named `project_topics`.

Having said that, I'm still not sure if this feature is necessary. It feels like an unique situation and I imagine gitlabform is not a tool for every case. But, it can be used for any use case by utilizing other processes (i.e. script that generates a config as needed and runs gitlabform). Would be good to get others to weigh in. What do you think @gdubicki @TimKnight-DWP ?

jimisola:
@amimas I'll answer your comment in full later but you are forgetting one important thing: topics it the _ONLY_ project setting that is [an array](https://docs.gitlab.com/ee/api/projects.html#edit-a-project) and, therefore probably, the only setting that has this issue (where you risk of overwritten).

amimas:
Hmm... That might be debatable. I mean, just because an attribute is complex type (i.e. list, dict), doesn't mean it has to be treated differently (i.e. separate API endpoint).

Actually there are other similar cases already. Take the permission for a branch protection as an example. The attributes such as: who can push, who can merge, etc. takes an array specifying those users/groups. But they are part of the same branch protection API endpoint.

amimas:
The branch protection is actually pretty good example for custom logic. GitLab's API expects user/group **ID* (integer). But, gitlabform allows specifying username or group path for ease of use.

Similarly, if we do need any specific need in project topics, I'm leaning towards adding logic to the existing proecssor. But, that's after we agree on what that need/feature is.

jimisola:
I try to avoid having a bunch of scripts outside of gitlabform. Having scripts outside of gitlabform does not solve our problem because we wouldn't be able to use topics in gitlabform at all since it would overwrite existing once. Not being able to set them in gitlabform would mean that 1) it's not a single source of truth and 2) that these scripts would have to sync with gitlabform.

E.g. in gitlabform configuration is made for groups/* and a topic (A) is set as well. 

You are saying set that in a script instead. So, we set topic A for that path in a script instead. Now the path is changed in the gitlabform but forgotten in the script. 

That a mistake waiting to happen.

jimisola:
> Hmm... That might be debatable. I mean, just because an attribute is complex type (i.e. list, dict), doesn't mean it has to be treated differently (i.e. separate API endpoint).

I think you are missing my point here. My point is that with non-complex types you don't have this issue. You can't update/set parts of an integer or boolean. 

Where as wanting to add a topic to existing ones is a basic and likely use-case. 

> Actually there are other similar cases already. Take the permission for a branch protection as an example. The attributes such as: who can push, who can merge, etc. takes an array specifying those users/groups. But they are part of the same branch protection API endpoint.
> 

I'll look at that.

jimisola:
> If we still need "extra" ability for managing topics, I think that should be done as part of the existing processor/config section. So, that would mean adding the logic under `project_settings.topics` instead of introducing a new config section named `project_topics`.

Agreed. That would be ideal.

What are you proposing that the syntax would be for this to be a non-breaking change?

jimisola:
that would mean adding the logic under project_settings.topics instead of introducing a new config section named project_topics.

amimas:
> What are you proposing that the syntax would be for this to be a non-breaking change?

Implementation-wise I think the update can happen in 2 phases. First, update the project settings without the `topic` attribute. Second, update the project but only apply the `topic` attribute. And in this 2nd phase there has to be extra logic because GitLab's API only expects list of string for `topic` attribute but in our config syntax, it'll be a bit complex.

For config syntax, thinking out loud here. We have to deal with it based on how GitLab's API works. Which means, we have to prepare a list of string (i.e. topics) and send it to the API. GitLab will overwrite everything with whatever is sent as part of the API data.

With that in mind, following snippet introduces a **special item** named `keep_existing` that is a boolean type. It should be removed from the data since it's only for gitlabform config syntax. By default it should be `false` or does not exist, for backward compatibility with current config. If this key enabled, that means remaining items should be an "addition". So, we need to retrieve the existing topics, add these new topics, and then send it to GitLab's API.

```yaml
projects_and_groups:
  "test-group/test-project-1":
    project_settings:
      visibility: internal
      default_branch: main
      auto_devops_enabled: false
      topics:
        - keep_existing: true
        - topic1
        - topic2:
```

I'm not sure if `enforce` key makes sense here, just because of how GitLab's API works as explained above. Besides, `enforce` is basically the existing behaviour - only use the topics that are in the config and remove everything else.

In terms of "delete" functionality, I guess it could follow the earlier example? The following snippet says, keep existing topics, but add a new one (i.e. `topic1`) and delete `topic2` if it exists.

```yaml
projects_and_groups:
  "test-group/test-project-1":
    project_settings:
      visibility: internal
      default_branch: main
      auto_devops_enabled: false
      topics:
        - keep_existing: true
        - topic1
        - topic2:
          delete: true
```

The key here is to make sure the processing is being done correctly because the data under `topics` is not a list of string anymore, but that's what it has to be when calling GitLab's API.

What do you think?

jimisola:
In general, it looks good to me. Thinking whether keep_existing is best or if there are alternatives.

That said, I'm a bit afraid of making the existing project settings "complicated" since it's now no longer just a 1-to-1 match with the attributes in GitLab API.

Will this be difficult to [document](https://gitlabform.github.io/gitlabform/reference/settings/) in a way that is understandable? 

I'm thinking that we add a section explaining the additional syntax for topics. What do you think?

I'm not 100% sure on how to migrate my code for project_topics to projects.settings,
but I'll have a look now.

jimisola:
How likely is it that someone would like to use `keep_existing` as a topic?


```
          project_settings:
              topics:
                - keep_existing # a topic name
                - topicA
                - topicB
                - topicC
```

vs

```
          project_settings:
              topics:
                - keep_existing: true
                - topicA
                - topicB
                - topicC
```

I can look to see if it's a string or not handle it accordingly.

jimisola:
I've updated the code but I have a circular dependency which I don't get (not familiar with the inner workings of gitlabform well enough yet).


```
(venv) jimisola@turing:~/cloud/dev/clones/jimisola/github/gitlabform (feature/add_cud_project_topics)$pytest tests/acceptance/standard/test_project_settings.py 
ImportError while loading conftest '/home/jimisola/cloud/dev/clones/jimisola/github/gitlabform/tests/acceptance/conftest.py'.
tests/acceptance/__init__.py:12: in <module>
    from gitlabform import GitLabForm
gitlabform/__init__.py:36: in <module>
    from gitlabform.configuration.transform import ConfigurationTransformers
gitlabform/configuration/transform.py:15: in <module>
    from gitlabform.gitlab import AccessLevel
gitlabform/gitlab/__init__.py:13: in <module>
    from gitlabform.gitlab.project_badges import GitLabProjectBadges
gitlabform/gitlab/project_badges.py:1: in <module>
    from gitlabform.gitlab.projects import GitLabProjects
gitlabform/gitlab/projects.py:7: in <module>
    from gitlabform.gitlab import GitLab
E   ImportError: cannot import name 'GitLab' from partially initialized module 'gitlabform.gitlab' (most likely due to a circular import) (/home/jimisola/cloud/dev/clones/jimisola/github/gitlabform/gitlabform/gitlab/__init__.py)
```

amimas:
> Implementation-wise I think the update can happen in 2 phases. First, update the project settings without the topic attribute. Second, update the project but only apply the topic attribute. And in this 2nd phase there has to be extra logic because GitLab's API only expects list of string for topic attribute but in our config syntax, it'll be a bit complex.

Thinking about it again, it doesn't/shouldn't need to be done in 2 phase. The data preparations for topic should happen anyways. And then we should send the whole data as request to gitlab. My earlier suggestions will add a performance hit since every project update will end up with 2 API calls. That's not necessary.

> I don't have the know how on how to migrate projects_settings to python-gitlab. Is that something you think we should do before?

I do think that would be ideal. I imagine that will make the implementation of this feature easier. In fact, I already have an open PR about it (#929 ). There's just one outstanding question for that. I can try look at that again when I get a chance and possibly we can come to a workaround for the outstanding item.

> How likely is it that someone would like to use `keep_existing` as a topic?

Good point. I can't imagine a use case where that would be a valid topic. Still, if we want to be extra cautious, we can add a prefix. Maybe something like `gitlabform_keep_existing` and/or `glf_keep_existing`. Should this be extended to the delete functionality too? For example: `gitlabform_delete_topic1` would mean delete topic1.

jimisola:
> > Implementation-wise I think the update can happen in 2 phases. First, update the project settings without the topic attribute. Second, update the project but only apply the topic attribute. And in this 2nd phase there has to be extra logic because GitLab's API only expects list of string for topic attribute but in our config syntax, it'll be a bit complex.
> 
> Thinking about it again, it doesn't/shouldn't need to be done in 2 phase. The data preparations for topic should happen anyways. And then we should send the whole data as request to gitlab. My earlier suggestions will add a performance hit since every project update will end up with 2 API calls. That's not necessary.

Have a look at my implementation in the PR. I frist fetch the _existing topics_ always as it now but I can change this so that those are only fetched if keep_existing is True.

Then it's one PUT api call to set/update/delete topics.
 
> > I don't have the know how on how to migrate projects_settings to python-gitlab. Is that something you think we should do before?
> 
> I do think that would be ideal. I imagine that will make the implementation of this feature easier. In fact, I already have an open PR about it (#929 ). There's just one outstanding question for that. I can try look at that again when I get a chance and possibly we can come to a workaround for the outstanding item.

It's still out of my comfort zone. Especially, since I'm having issues with my computer freezing when I run gitlab docker.
 
> > How likely is it that someone would like to use `keep_existing` as a topic?
> 
> Good point. I can't imagine a use case where that would be a valid topic. Still, if we want to be extra cautious, we can add a prefix. Maybe something like `gitlabform_keep_existing` and/or `glf_keep_existing`. Should this be extended to the delete functionality too? For example: `gitlabform_delete_topic1` would mean delete topic1.

I don't like the prefix idea. I can easily just check if keep_existing is a boolean or not. That way. it's transparent.

With `gitlabform_delete_topic1` did you mean:

```yaml
          project_settings:
              topics:
                - keep_existing: true
                - topicC
                - topicD
                - gitlabform_delete_topicA
```

That is just a confusing syntax and you don't the the clarity as you do with:

```
            project_settings:
              topics:
                - keep_existing: true
                - topicA:
                    delete: true
                - topicC
```

amimas:
When you get a chance, let's update this PR based on latest updates. The PR #929 has been merged. So, adding custom logic to existing processor for `project_settings` config should be simpler. Just need to prepare the topics related data according to the proposed config syntax discussed above.

jimisola:
@amimas Looks like the pipeline passed now. Marking as ready. I will 'squash and merge' upon approval.

docs/features.md
```diff
@@ -30,6 +30,7 @@ GitLabForm enables you to manage the [(GitLab's) Application Settings](reference
     * [Push Rules](reference/push_rules.md) (**GitLab Premium (paid) only**)
     * [Integrations](reference/integrations.md)
     * [Settings](reference/settings.md#project-settings)
+
```
amimas, line 33: Let's undo this change.

gitlabform/processors/project/project_settings_processor.py
```diff
@@ -48,3 +52,46 @@ def _print_diff(
...
+        if keep_existing:
+            api_adjusted_topics.extend(project_settings_in_gitlab.get("topics", []))
+
+        # List of topics not having delete = true or no delete attribute at all
+        topics_to_add: List[str] = [
+            list(t.keys())[0] if isinstance(t, dict) else t
+            for t in project_settings_topics
+            if isinstance(t, str)
+            or (isinstance(t, dict) and not list(t.values())[0].get("delete", False))
+        ]
```
amimas, line 82: I'm not an expert. So, for me it's not easy to follow. Could something like this work?

```python
topics_to_add: List[str] = []

for t in project_settings_topics:
    if isinstance(t, str):
        topics_to_add.append(t)
```

Do we need to check if a topic does not have delete attribute? Since we're looking for topics to be "added", isn't it enough to just take the ones that are just string type? If they are not, then they are not valid config/data type, as far as "topic to add" is concerned, right?

I prefer readability and it might even be helpful for beginners to be able to contribute. Unless the performance benefit is significant.

amimas (commented):
Haven't finished going through the entire change. Ran out of time and will try to get back to it later. But, thought I'd post some feedbacks.

gitlabform/processors/project/project_settings_processor.py
jimisola: Your change is not the same as mine (although I prefer the readability).

Mine will handle

```yaml
        - topicC
            delete: false

```
as opposed to your proposal. That would mean that `delete: false` is not valid. If so, then we can change it. But, since it's a boolean I wanted to have that covered.

Also, your proposal is good in a way because it would also handle (not in the best way but):

```
        - topicC
            nonsense: false
```

and with handle I mean that GitLabForm will not crash but topicC will be delete with perhaps is not what the user want.

I think that I need to make a pass first:

1. check for Dict
2. if Dict, check that it is `delete: true|false` (or even just `true`) and otherwise output an error about syntax error.

jimisola:
> Haven't finished going through the entire change. Ran out of time and will try to get back to it later. But, thought I'd post some feedbacks.

Thanks for your detailed review. Much appreciate (sat a little too late last night). Commit your changes to the docs and left is the handling of `delete: true`.

docs/features.md
amimas: Reopening this one. The latest diff as of now still shows this line.

I think it might split the feature list with a blank line. Since our CI pipeline doesn't build a review version for the website, I can't tell unless we manually verify it locally. So, might be safer to undo the change in this file.

gitlabform/processors/project/project_settings_processor.py
amimas: Ahh... I see. I hadn't reviewed after this point. I saw the next lines of code is dealing with `delete` and thought it might be handled there. But... I was just thinking of something else.

Is there ever a case where `delete: false` is needed? That's equivalent of not adding that config, isn't it? I mean users can add `delete: false` for a topic or just list that topic as a string. Only time we need to do something is when `delete: true` is set for a topic. What do you think? Not sure if I'm forgetting something.

Similarly, is there ever a need for `keep_existing: false`? Users can add that config or omit it. It just means the existing behaviour. We only need to act when `keep_existing: true` is in the config, right?

Just wondering if the above would simplify the logic anymore and maybe make it readable too. I haven't really thought it through it.

gitlabform/processors/project/project_settings_processor.py
```diff
@@ -48,3 +52,46 @@ def _print_diff(
             entity_config,
             only_changed=diff_only_changed,
         )
+
+    def _process_project_topics(
```
amimas, line 56: After this method is completed, the updated data will be sent to the API. I think it might be good to have some debug logs since there's a good amount of data transformation happening here.

gitlabform/processors/project/project_settings_processor.py
```diff
@@ -48,3 +52,46 @@ def _print_diff(
...
+
+        for i, topic in enumerate(project_settings_topics):
+            if isinstance(topic, dict) and "keep_existing" in topic:
+                value = topic["keep_existing"]
+                if isinstance(value, bool):
+                    keep_existing = value
+                    del project_settings_topics[i]
+                    break
+
+        api_adjusted_topics: List[str] = []
```
amimas, line 71: Do you think this variable could be renamed? It's not easy to tell what this is for, without knowing the inner workings of this method. Does something like `generated_project_topics` make sense?

tests/acceptance/standard/test_project_settings.py
```diff
@@ -16,3 +19,78 @@ def test__builds_for_private_projects(self, gl, project):
...
+        run_gitlabform(config, project)
+
+        updated_project = gl.projects.get(project.id)
+        project_topics: List[str] = updated_project.topics
+
+        assert len(project_topics) == 2
+        assert "topicC" in project_topics
+        assert "topicD" in project_topics
+
+    def test__edit_project_settings_topics_keep_existing(
```
amimas, line 47: Let's add a test case where `keep_existing: false` is also tested.

tests/acceptance/standard/test_project_settings.py
```diff
@@ -16,3 +19,78 @@ def test__builds_for_private_projects(self, gl, project):
...
+        updated_project = gl.projects.get(project.id)
+        project_topics: List[str] = updated_project.topics
+
+        assert len(project_topics) == 4
+        assert "topicA" in project_topics
+        assert "topicB" in project_topics
+        assert "topicC" in project_topics
+        assert "topicD" in project_topics
+
+    def test__delete_project_topics(self, gl: Gitlab, project: Project) -> None:
```
amimas, line 74: Let's add a test case where `delete: false` is also tested.

gitlabform/processors/project/project_settings_processor.py
jimisola: We don't really generate any topics per se. We do adjust the topics that should be sent in the API call.

The current name may not be ideal but I'm not certain that your suggestion is better.

amimas: Yeah, I'm not saying use my suggestion. I couldn't think of anything else at the time. The reason I thought of `generated_project_topics` is because the list of topics are actually being prepared (for lack of better name) here to be sent to the API. Maybe `project_topics_to_be_set` for absolute clarity? Again, just something off the top of my head.

tests/acceptance/standard/test_project_settings.py
jimisola: How should we handle these if the key is not delete/keep_existing? That is,

```
            project_settings:
              topics:
                - keep_existing: true
                - topicA:
                    delete: true
                - topicC
                - topicZ:
                    not_a_valid_key: true
```

amimas: What do you suggest?

I'm thinking if there's a config for a topic that is not supported (i.e. `not_a_valid_key: true`), we print a warning about this not being a supported/recognized config and then move on.

 ```yaml
            project_settings:
              topics:
                - keep_existing: true
                - topicA:
                    delete: true
                - topicC
                - topicZ:
                    not_a_valid_key: true
```

If the `--strict` mode is enabled, we can exit with a non-zero code for invalid config. I'm not too sure if this is the sort of scenario that is intended for the strict mode, but I can't think of any reason why it wouldn't be.

If there's a config like this where `not_a_valid_key: true` itself is configured as a topic, then we should just take that as-is (i.e. a topic for the project).

```yaml
            project_settings:
              topics:
                - keep_existing: true
                - topicA:
                    delete: true
                - topicC
                - not_a_valid_key: true
 ```

What do you think?

File: gitlabform/processors/project/project_settings_processor.py (hunk 4 above)
Comment (jimisola) [reply gitlabform/processors/project/project_settings_processor.py:71]: I used adjusted (prepared) because the ones in the config file are adjust based on the config file and the current topics. api_adjusted because they are adjusted due to how the api works, i.e. no update method - just set.

Let's remove API.

Suggestion:

adjusted_project_topics_to_set

This name clearly indicates that the topics have been adjusted and are intended to be set as the project’s topics.

File: tests/acceptance/standard/test_project_settings.py (hunk 5 above)
Comment (jimisola) [reply tests/acceptance/standard/test_project_settings.py]: Appreciate the extensive response.

I think your suggestion is good. The only thing I'm wondering about is consistency.

Do we handle the use of `enforce` and `inherit` in the same way?

I glanced at the code and it seems as it's not consistent as it is now, e.g.


* we default to false: https://github.com/gitlabform/gitlabform/blob/main/gitlabform/processors/project/job_token_scope_processor.py#L35
* from what I can see we don't check for valid datatypes (key names)


I think that we should go with your suggestion (it's what I would have suggested). Being consistent and also handling incorrect name of keys is a bigger task and something that I suggest we tackle in https://github.com/gitlabform/gitlabform/issues/331 and https://github.com/gitlabform/gitlabform/issues/499.

Comment (amimas) [reply tests/acceptance/standard/test_project_settings.py]: > from what I can see we don't check for valid datatypes (key names)

I believe that's because of the pattern of following [raw parameter passing](https://gitlabform.github.io/gitlabform/reference/#raw-parameters-passing).

> Do we handle the use of `enforce` and `inherit` in the same way?

They are slightly different, in my opinion. Here, we are actually allowing different kinds of datatype into an attribute, transforming that input into what is expected by GitLab (i.e. a list of strings). What happens if you tried the following in current latest version of gitlabform where this feature is not available yet:

```yaml
            project_settings:
              topics:
                - keep_existing: true
                - topicA:
                    delete: true
                - topicC
```

I assume it might fail because maybe gitlab API doesn't like it (because of the object for `topicA`). Otherwise, the list would be set as the project topics including `keep_existing: true`. We could probably go that way too. Maybe we don't need to do extra data type validation?

gitlabform/processors/project/project_settings_processor.py
jimisola: added as 

`debug(f"topics after adjustment: {adjusted_project_topics_to_set}")`

tests/acceptance/standard/test_project_settings.py
jimisola: I assume the same as you. GitLab API would fail/warn since topics is not a list of strings.

So, 

1. `keep_existing: true|false`: handled correctly as-is
2. `keep_existing: true|false`: handled accordingly, child `delete` is removed and topic converted to str
3. for all other cases value the incorrect datatypes and/or topic has a child that is not `delette`: error left to GitLab API

Correct?

setup.py
```diff
@@ -53,6 +54,7 @@ def get_version_file_path():
         "requests==2.32.3",
         "ruamel.yaml==0.17.21",
         "types-requests==2.32.0.20241016",
+        "types-setuptools==75.8.0.20250110",
```
amimas, line 57: just for my knowledge, why is it needed now?

tests/acceptance/__init__.py
```diff
@@ -64,72 +66,78 @@ def allowed_codes(codes: Union[int, List[int]]):
...
+        group_obj = gl.groups.create(
             {
                 "name": group_name,
                 "path": group_name,
                 "parent_id": parent_id,
                 "visibility": "internal",
             }
         )
+    # Cast the returned RESTObject to Group
+    group: Group = cast(Group, group_obj)
```
amimas, line 103: Why is this casting needed now?

tests/acceptance/__init__.py
amimas, line 1: There's a ton of changes in this file. It's good, but I think it could've been a separate chore PR instead. Keeps the review and contexts of changes clearer.

tests/acceptance/premium/test_push_rules.py
amimas, line 1: Same comment as above with updates to this file. Personally I prefer changes in PRs to be isolated to the original purpose of the PR.

amimas (approved):
Thanks for working on this feature @jimisola. Left couple of comments/feedbacks but I think overall it looks good. Can you think of anything that is outstanding or is this ready to go?

setup.py
jimisola: typing for setuptools. All my code is using typing and some dependencies need typing dependencies to be added.
I'm quite skeptic that it needs to be a runtime dependency but I went with the same scope as types/typing for requests library.

When we redo the build and project management we should have runtime, lint and test (for simplicity dev = lint + test + types).

tests/acceptance/__init__.py
jimisola: typing (mypi). as per comment, a RESTObject is returned to a Group in python-gitlab instead of a Group.

tests/acceptance/premium/test_push_rules.py
jimisola: Me too. Since, the GitHub Actions failed I needed to fix the mypy errors.
The only one change I felt was not really related were changing the two regexps (by @TimKnight-DWP) to use raw strings (not to cause mypi errors).

tests/acceptance/__init__.py
jimisola: I needed this code to use typing otherwise it would cause mypy errors in my code (something like "Call to untyped function in typed context").

We discussed this in #maintainers. Either we add typing everywhere at the same time or we do the changes like I did here. I really don't want to add new code that is not using typing.

jimisola:
> Thanks for working on this feature @jimisola. Left couple of comments/feedbacks but I think overall it looks good. Can you think of anything that is outstanding or is this ready to go?

Thanks. Appreciate the feedback loops. Nothing outstanding to my knowledge.

jimisola:
@amimas I've resolved all threads (as I answered your question and you already approved). I'll wait with the merging until later tonight, so that you can read the comments in case there is something there.

setup.py
amimas: Sounds good. Yeah, I was also thinking about why these were runtime/required dependency. But, like you said I think the project could possibly benefit from a refined setup.

PR #935, opened 2025-01-29
chore(ci): configure dependabot to group python-gitlab packages

amimas:
There are currently 2 packages required from python -gitlab. Dependabot opens separate PR for updating their versions but they fail to install due to conflicts. See issue #934.

This change updates the config so that python-gitlab packages are updated together in a single PR. That should help us easily stay up-to-date with newer release of that package, without needing to manually perform update.

closes #934

Issue #934, opened 2025-01-29
Dependabot PR for updating python-gitlab fails in CI


amimas:
Dependabot opened 2 PRs for updating python-gitlab version but they are failing to pass the CI because they need to be updated together.

```
The conflict is caused by:
    gitlabform 4.1.1 depends on python-gitlab==5.6.0
    gitlabform 4.1.1 depends on python-gitlab==5.3.1
```

The conflict comes from `setup.py` having 2 separate dependencies listed:

```
        "python-gitlab==5.3.1",
        "python-gitlab[graphql]==5.3.1",
```

Ref: https://github.com/gitlabform/gitlabform/actions/runs/13019532488/job/36316801617

Example PRs: #932, #933 

Need to configure dependabot so that these updates are done together.

PR #936, opened 2025-01-29
chore(ci): fix dependabot config group package pattern

amimas:
I think the previous attempt in #935 for grouping python-gitlab packages together by dependabot PR was incomplete. I re-triggered dependabot but it didn't seem to resolve the issue mentioned in #934. I suspect the pattern for selecting which packages to be groped was not correct.

This PR uses a glob pattern instead of a static string to specify that python-gitlab packages should be grouped together. Not sure how to test this though without going through reproducing the issue elsewhere and testing the fix, which is kind of a lot of work.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,4 +20,4 @@ updates:
     groups:
       python-gitlab:
         patterns:
-          - "python-gitlab"
+          - "python-gitlab*"


TimKnight-DWP (approved):

amimas:
Looks like this change worked! #937

PR #941, opened 2025-02-03
chore(dev): add options gitlab-ce etc  to run_gitlab_docker

jimisola:
Make the dev script a bit more configurable by providing cmd line options:

      --help                  Display this help message.
      --gitlab-version <version> Specify the GitLab version to deploy (default: latest).
      --gitlab-flavor <flavor> Specify the GitLab flavor to deploy ("ce" or\"ee"). Default is "ee".
      --cpus <value>          Limit CPU usage (e.g., 0.5, 2, etc.). Default: no limit.
      --memory <value>        Limit memory usage (e.g., 512m, 2g, etc.). Default: no limit.

### DIFF
--- a/dev/run_gitlab_in_docker.sh
+++ b/dev/run_gitlab_in_docker.sh
@@ -41,22 +41,101 @@ cecho() {
 }
 
 # ============================================================================
+#  Defaults and Argument Parsing
+# ============================================================================
+
+# Default values
+gitlab_version="latest"
+gitlab_flavor="ee"
+cpu_limit=""
+memory_limit=""
+
+# Process command-line arguments
+while [[ $# -gt 0 ]]; do
+  key="$1"
+  case $key in
+    --help)
+      echo "Usage: $0 [options]"
+      echo ""
+      echo "  --help                  Display this help message."
+      echo "  --gitlab-version <version> Specify the GitLab version to deploy (default: latest)."
+      echo "  --gitlab-flavor <flavor> Specify the GitLab flavor to deploy (\"ce\" or \"ee\"). Default is \"ee\"."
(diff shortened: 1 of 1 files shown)


amimas (approved):
I quickly went over the changes and didn't actually try it out, but it looks alright to me. Just couple of comments/suggestions from my observation.

- Before we merge this I suggest PR title is changed to `build` or `chore` to indicate commit type. I understand it can be seen as a feature for the dev script, but from a consumer perspective of this project, there is no feature they are getting out of gitlabform as a result of this update.

- This is probably for future. We can maybe just use a python script. It'll be a lot less and simple, especially when adding a CLI type interface. It's reasonable to expect that the end-user of this script will have python installed since that's required for developing gitlabform anyways.

jimisola:
> I quickly went over the changes and didn't actually try it out, but it looks alright to me. Just couple of comments/suggestions from my observation.
> 
> * Before we merge this I suggest PR title is changed to `build` or `chore` to indicate commit type. I understand it can be seen as a feature for the dev script, but from a consumer perspective of this project, there is no feature they are getting out of gitlabform as a result of this update.

Ok. I was not sure if we had some sort of standard. From my experience/POV:

* build: things related to build of software such as setupstools, hatch etc and I do see a dev script as part of the build 
* chore: reserved for renovate (bots), but not written in stone (hence, why I used feat(dev)). I understand you're view, perhaps chore(dev) is better then.

> * This is probably for future. We can maybe just use a python script. It'll be a lot less and simple, especially when adding a CLI type interface. It's reasonable to expect that the end-user of this script will have python installed since that's required for developing gitlabform anyways.

Yeah. I was on my way to changing it to a python script, but I was not sure if I would break it for someone else then.

PR #943, opened 2025-02-05
Migrate to pyproject.toml

gdubicki:
as [while `setup.py` is not deprecated, the use of `pyproject.toml` is recommended](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/).

A visible benefit from this PR will be that in https://pypi.org/project/gitlabform/ there will be more links for the project shown.

~Also there will be less files in the root directory of the project, which means less scrolling on GitHub.~ (Nvm this, the UI has changed so it doesn’t matter for this purpose.)


pyproject.toml
jimisola, line 1: Please have a look at this [example](https://github.com/luftfartsverket/reqstool-client/blob/main/pyproject.toml).

* [tool.pytest.ini_options], if there is something there that is relevant now otherwise I'll look into it when we move on to hatch
* [tool.black]: line-length (120 :pray:)  and target-version 
* [tool.flake8]: cmd line options and max-line-length (120 :pray:)

gdubicki: > Please have a look at this [example](https://github.com/luftfartsverket/reqstool-client/blob/main/pyproject.toml).
> 
> * [tool.pytest.ini_options], if there is something there that is relevant now otherwise I'll look into it when we move on to hatch

Sorry, what? I didn’t get that. :/

> * [tool.black]: line-length (120 🙏)  and target-version
> * [tool.flake8]: cmd line options and max-line-length (120 🙏)

I am for 120 too but it will require reformatting the project, which is fine but we should do it in a separate PR. Any objections, @amimas @TimKnight-DWP ?

jimisola: Yes. Separate PR it is.

Is there anything here that would be of use now or shall I do it as I do the hatch and Python 3.13 thing?

```
[tool.pytest.ini_options]
python_classes = ["!TestDataModelGenerator", "!TestStatisticsItem"]
addopts = [
    "-rsxX",
    "-s",
    "--import-mode=importlib",
    "--log-cli-level=DEBUG",
    '-m not slow and not integration',
]
pythonpath = [".", "src", "tests"]
testpaths = ["tests"]
markers = [
    "flaky: tests that can randomly fail through no change to the code",
    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
    "integration: tests that require external resources",
]
```

gdubicki: Frankly I would prefer this PR to be mostly about the format change for bigger additions I would suggest separate ones.

jimisola (approved):

gdubicki:
Thanks for the review @jimisola and @TimKnight-DWP!

Please note that while I tested if the package builds after this change, I obviously did not make a new release, so please keep an eye for possible issues on the next one. Sorry in advance for them, if the occur! cc @amimas

PR #954, opened 2025-02-13
fix: project topics should not be removed if topics are not in the config

### LINKED ISSUE #946: Project topics removed by default
With changes due to #931, projects that had manually added topics have them removed if gitlabform config doesn't specify the topics attribute.  The default behavior should be to keep any existing topics unless an empty topics array is specified.

Steps to recreate:
* Set a topic on a project
* Ensure the relevant gitlabform config that applies to the project has `project_settings` without a `topics` property specified
* Run gitlabform, ensuring the project is evaluated
* Verify that gitlabform removed the topic

Expected behavior:
* Set a topic on a project
* Ensure the relevant gitlabform config that applies to the project has `project_settings` without a `topics` property specified
* Run gitlabform, ensuring the project is evaluated
* Verify project retains original topic

amimas:
If the `project_settings` config does not specify `topics` attribute, any topics added manually are being removed. This is bug introduced in v4.1.2. This change reproduces the bug and fixes it so that manually managed topics configs are not removed.

fixes #946

### DIFF
--- a/gitlabform/processors/project/project_settings_processor.py
+++ b/gitlabform/processors/project/project_settings_processor.py
@@ -59,6 +59,9 @@ class ProjectSettingsProcessor(AbstractProcessor):
     ) -> None:
         project_settings_topics: Dict = project_settings_in_config.get("topics", [])
 
+        if not project_settings_topics:
+            return
+
         keep_existing: bool = False
 
         for i, topic in enumerate(project_settings_topics):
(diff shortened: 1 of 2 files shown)


jimisola (approved):
Looks. I noticed that our topics were gone  a few days ago and then something else came up. Created an internal issue to troubleshot and missed that a user found.

I thought we covered all the causes but we apparently let this one slip through. 

PR looks good.

PR #955, opened 2025-02-13
refactor: migrate 'merge_requests_approvals' processor to python-gitlab

### LINKED ISSUE #928: Migrate `merge_requests_approvals` config to use python-gitlab library
Goal of this issue is to migrate the `merge_requests_approvals` configuration section to use [python-gitlab library](https://python-gitlab.readthedocs.io/en/stable/) instead of using homegrown API in this project.

High level Tasks:

- Update `gitlabform/gitlabform/processors/project/merge_requests_approvals.py` to use the python-gitlab library
    - To use python-gitlab use `self.gl` instead of `self.gitlab`. For reference, see some of the other processors that are already using python-gitlab library (i.e. `branches_processor.py`)
- Any cleanup necessary from switching to python-gitlab related to this processor

amimas:
This PR refactors the logic that processes `merge_requests_approvals` section in a config for managing a project's MR approval settings. The code has been refactored to use python-gitlab as part of the effort of moving the project to use that library instead of relying on homegrown logic.

closes #928

### DIFF
--- a/gitlabform/processors/project/merge_requests_approvals.py
+++ b/gitlabform/processors/project/merge_requests_approvals.py
@@ -1,17 +1,44 @@
-from cli_ui import fatal
+from cli_ui import fatal, debug
+from typing import Callable
+from gitlab.v4.objects import Project
 
 from gitlabform.gitlab import GitLab
-from gitlabform.processors.single_entity_processor import SingleEntityProcessor
+from gitlabform.processors.abstract_processor import AbstractProcessor
+from gitlabform.processors.util.difference_logger import DifferenceLogger
 
 
-class MergeRequestsApprovals(SingleEntityProcessor):
+class MergeRequestsApprovals(AbstractProcessor):
     def __init__(self, gitlab: GitLab):
-        super().__init__(
-            "merge_requests_approvals",
-            gitlab,
-            get_method_name="get_approvals_settings",
-            edit_method_name="post_approvals_settings",
+        super().__init__("merge_requests_approvals", gitlab)
+        self.get_entity_in_gitlab: Callable = getattr(
+            self, "get_project_mr_approvals_settings"
         )
+
+    def _process_configuration(self, project_path: str, configuration: dict) -> None:
+        debug("Processing project merge requests approvals settings...")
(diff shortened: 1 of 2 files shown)


gitlabform/processors/project/merge_requests_approvals.py
```diff
@@ -1,18 +1,45 @@
...
         )
 
+    def _process_configuration(self, project_path: str, configuration: dict) -> None:
+        debug("Processing project merge requests approvals settings...")
+        project: Project = self.gl.get_project_by_path_cached(project_path)
+
+        mr_approval_settings_in_config: dict = configuration.get(
+            "merge_requests_approvals", {}
+        )
+        mr_approval_settings_in_gitlab: dict = project.approvals.get().asdict()
```
jimisola, line 24: Since you use typing you might as well specify what the dict's key and value have for data types.

amimas: Thanks for the review and feedback @jimisola . Do you mean something like this?

```suggestion
        mr_approval_settings_in_gitlab: dict[str, bool] = project.approvals.get().asdict()
```

If yes, I'm not sure I can do that reliably. Looking at GitLab's API for [merge request approval settings](https://docs.gitlab.com/ee/api/merge_request_approval_settings.html#update-project-mr-approval-settings), the above might be applicable. However when they update that API endpoint with different types of attributes, above typing won't make sense.

Maybe I misunderstood what you meant. If so, please clarify. Otherwise, I think it should be as-is?

jimisola: Ok, I see. That's what I meant. But, you can combine

Dict[str, Union [str, Any]]

Gives a hint that it's mostly str, str at least.

jimisola (approved):

gitlabform/processors/project/merge_requests_approvals.py
amimas: Ahh... okay. If the type hint will indicate "anything", I feel let's just keep it as `dict`. It's simpler for readability. Since you've approved the current way, I assume you're okay with it. Going to resolve this thread.

PR #957, opened 2025-02-15
docs: recommend uv instead of pipx for running/installing

gdubicki:
as it handles Python upgrades automatically, while pipx may keep you stuck with the existing, old version of Python (unless I am missing something)

It's also the ✨shiny new thing✨

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -10,9 +10,9 @@ You can just run the latest stable version with:
 ```shell
 docker run -it -v $(pwd):/config ghcr.io/gitlabform/gitlabform:latest gitlabform
 ```
-* [pipx](https://github.com/pypa/pipx):
-```shell
-pipx run gitlabform
+* [uv](https://github.com/astral-sh/uv):
+```uv
+uvx gitlabform
 ```
 
 
@@ -22,9 +22,9 @@ Similarly, you can run the latest 3.* version with:
 ```shell
 docker run -it -v $(pwd):/config ghcr.io/gitlabform/gitlabform:3 gitlabform
 ```
-* [pipx](https://github.com/pypa/pipx):
+* [uv](https://github.com/astral-sh/uv):
 ```shell
-pipx run --spec 'gitlabform>=3,<4' gitlabform
+uvx --from 'gitlabform>=3,<4' gitlabform
 ```
 
 See [this](https://github.com/gitlabform/gitlabform/pkgs/container/gitlabform) for all the available Docker tags.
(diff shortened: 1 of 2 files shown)


jimisola (commented):
I was waiting with this changes until we moved to gitlab.com as I think I mentioned in the TODO/priority list in our maintainer chat.

I only see changes in the docs not in the CI/CD. I think that it's essential to use the same build and package tools in CI/CD as we recommend in our docs. How many of us have tested with uv? Just so that we don't recommend something that there might be issues with.

Also see a value as not freeze these type of changes, CI/CD etc, until the move to gitlab.com so that we don't have to continue to make changes to the GitLab CI/CD that @TimKnight-DWP has worked with.

Once moved to gitlab.com, I'm all for uv and a move to hatch (as per previous chat).

docs/installation.md
```diff
@@ -22,9 +22,9 @@ Similarly, you can run the latest 3.* version with:
 ```shell
 docker run -it -v $(pwd):/config ghcr.io/gitlabform/gitlabform:3 gitlabform
 ```
-* [pipx](https://github.com/pypa/pipx):
+* [uv](https://github.com/astral-sh/uv):
 ```shell
-pipx run --spec 'gitlabform>=3,<4' gitlabform
+uvx --from 'gitlabform>=3,<4' gitlabform
```
amimas, line 27: This isn't really about `uv`. Looks like we already had constraint about gitlabform version being lower than 4. Not really familiar with either commands or what it's for. Should this be changed, now that we're already on version 4? But... what's the point of having a constraint? If we don't specify it, will it automatically install the latest one?

amimas (approved):
Sounds interesting, although I'm not really familiar with `uv` yet. Will try to learn it sometime. If this looks good to you, please feel free to merge.

One question: if we're instructing `uv` for local dev, why don't we use that in the CI too?

gdubicki:
Hey @jimisola and @amimas!

I created this as I noticed that `pipx upgrade gitlabform` doesn't correctly upgrade my install of the app because it was keeping using Python 3.11 for it and obviously gitlabform v4 is not available for it.

As I was reading about the fixes to update the docs I could really find a good one :/, that's why I suggested this change.

I didn't want to change to dev/CI/CD solution yet as I didn't use `uv` enough yet to do that.

Anyway, we don't need to move forward with this yet, we can wait for the move to gitlab.com first.

jimisola:
Ok. I think that we shall wait until the move to gitlab.com. There are a couple of things that I think that we should do at the same time.

amimas:
PR #1247 is introducing `uv` and other modern tools along with a development toolkit for gitlabform. Closing this PR in favor of that one.

PR #958, opened 2025-02-15
test: ensure that the app keeps running on MacOS and Windows

gdubicki:


gdubicki:
There's nothing to review here until we merge it and see how (if?) it works, so I will do that. cc @jimisola @amimas @TimKnight-DWP

PR #960, opened 2025-02-16
Optimize Docker image

gdubicki:
But I have _tiny_ doubts about whether the created image works. 😅 I mean: the app definitely _runs_, but I wonder if something will not fail when actually using some of its features.


pyproject.toml
```diff
@@ -62,6 +60,8 @@ test = [
     "pytest-cov==6.0.0",
     "pytest-rerunfailures==15.0",
     "xkcdpass==1.19.9",
+    "types-requests==2.32.0.20241016",
+    "types-setuptools==75.8.0.20250210",
```
amimas, line 64: Thanks for moving them out of the runtime dependencies.

May I request to move these 1 line up so that they are ordered alphabetically? :)

.dockerignore
amimas, line 1: What do you think of something like the following so that we don't need to manually maintain every directories or files.

```dockerignore
# Ignore everything by default
*
# Allow following files as only these are needed for building the container
! gitlabform/
! pyproject.toml
```

Dockerfile
```diff
@@ -1,14 +1,10 @@
 FROM python:3.12-alpine3.21
-RUN apk add --no-cache \
-    gmp \
-    lua5.3 \
-    lua5.3-lpeg
-RUN mkdir gitlabform
-COPY . /gitlabform
```
amimas, line 7: Don't we need this command to copy the sources into the container so that the next instruction can build?

Dockerfile
```diff
@@ -1,14 +1,10 @@
...
 
+# delete the pip cache because even with `--no-cache-dir` pip still creates ~1.4MB of cache...
 RUN cd gitlabform \
```
amimas, line 4: Minor commant. I wonder if it'd be good to use a separate instruction like this:

```
# Set the working directory for further steps
WORKDIR /app
```

so that we don't need to `cd` into the directory inside the `RUN` command. Thinking it makes it a bit more readable maybe? What do you think?

amimas (commented):
Haven't tried this out myself yet. Left a few comments to consider.

.dockerignore
gdubicki: Oh, sure, why I haven’t thought about it myself? 😅🙏

shoddyknight:
Shall we target v5 for this @amimas @gdubicki - since we're doing Release-Candidates for that right now as we have other large Breaking Changes to test

.dockerignore
amimas: Actually, there's a flaw with that approach. That could technically allow certain unwanted files inside the `gitlabform/` directory. It's probably better to be explicit similar to `.gitignore` file.

pyproject.toml
amimas: Pushed an update with this change.

Dockerfile
amimas: Confirmed this command is needed. Restored it. Pushed an update with this change.

Dockerfile
amimas: Pushed an update with this change.

amimas:
@gdubicki @TimKnight01 - Picked this up and rebased and pushed an update to it. My update mainly introduces multi-stage builds. Here's what image size looks like with various versions.

```
REPOSITORY                      TAG                                            IMAGE ID       CREATED          SIZE
localhost/gitlabform            optimized-multi-stage                          7ebf4d5297b3   16 hours ago     68.5MB
localhost/gitlabform            optimized                                      9bbf10d15f97   19 hours ago     77.1MB
ghcr.io/gitlabform/gitlabform   4.8.0                                          2c707b9ff8b0   3 days ago       88.5MB
```

Here you can see our current released version (v4.8.0) image size is 88.5MB. Original update from this PR reduced it down to 77.1MB. With the multi-stage docker build, it can come down to 68.5MB.

Let me know if you see any issues or concerns. Otherwise, let's get this in for v5 release.

TimKnight01:
@amimas looks good - I'll merge into V5 and do a test of v5 against a Gitlab repo somewhere

Issue #944, opened 2025-02-06
test_files.py::TestFiles::test__set_file_strongly_protected_branch consistently fails


TimKnight-DWP:
**Describe the bug**

tests/acceptance/standard/test_files.py::TestFiles::test__set_file_strongly_protected_branch fails in Pipeline at least 1 out of 4 times, my suspicion is that there is now some asynchonicity in Gitlab processing. 

This failing test needs investigating and seeing what we can do to make it more stable

amimas:
Looks like the issue still persists :(

PR #961, opened 2025-02-16
Workaround for #944

gdubicki:


jimisola (approved):
If you can run the test successfully, say 10 times, then I think we should consider it solved and merge.

amimas (approved):
The logic looks fine to me. I say let's merge if the CI passes. We'll find out in a week or so if it actually did address the annoying flaky issue.

Only thing I'd request is to use conventional commits for the PR title and use that as the commit message when merging :)

gdubicki:
> @jimisola: If you can run the test successfully, say 10 times, then I think we should consider it solved and merge.

I did rerun it about 5 times in GitHub Actions and a few times locally and all tries succeeded...

> @amimas: (...) I say let's merge if the CI passes. We'll find out in a week or so if it actually did address the annoying flaky issue.

...so I will merge it. :)

Issue #533, opened 2023-04-17
Variable deletion is unsuccessful due to environment scope


amimas:
**Describe the bug**

I have a config for a project whose CI CD variables are configured using the same key but scoped to different "environment" with different values. The config looks something like this:

```yaml
my-group/*:
    variables:
      # Remove unconfigured variables from **all projects**.
      enforce: true

my-group/sub-group/project1:
   variables:
      aws_access_key_id_review:
         key: AWS_ACCESS_KEY_ID
         value: ******
         protected: false
         masked: false
         environment_scope: review/*
         filter[environment_scope]: review/*
      aws_access_key_id_review:
         key: AWS_ACCESS_KEY_ID
         value: ******
         protected: false
         masked: false
         environment_scope: staging/*
         filter[environment_scope]: staging/*
```

In the above config, I have set `enforce: true` for all projects. This ensures any unconfigured projects in that group will have it's variables deleted. It works, but I just came across an edge case.

I removed the `my-group/sub-group/project1` config. After running `gitlabform` I get following error message:

```
Warning: Error occurred while processing project my-group/sub-group/project1, exception:
Request url='https://gitlab.example.com/api/v4/projects/my-group/sub-group/project1/variables/AWS_ACCESS_KEY_ID', method=DELETE, failed - expected code(s) [204, 404], got code 409 & body: '{"message":"There are multiple variables with provided parameters. Please use 'filter[environment_scope]'"}'
```

I suspect `gitlabform` does not does not have the context or set `filter[environment_scope]` parameter when calling the api for variable deletion. Also not sure if this is due to the project being in sub-group. I did come across edge case related to sub-group and variable enforcement before. See #457 for details.

**GitLabForm version**

Output of `gitlabform -V`:  `🏗 GitLabForm version: 3.4.3 = outdated 😔 , please update! (the latest stable is 3.5.0)`

**GitLab version**

`15.9.3-ee`

PR #962, opened 2025-02-17
fix: project variables with same name but different scope should be deletable

amimas:
This PR fixes the bug reported in #533 but in the scope of project. I haven't checked but I suspect "group variables processor" might have similar bug too. But, let's do that in a separate PR.

I think this is an important bug that unfortunately hadn't been fixed yet. As per the referenced issue, if there are multiple CI CD variables with same key/name but scoped to different "environment", gitlabform does not make the API call correctly. As a result, the variable cannot be deleted and an error is returned.

There's a lot of changes in this PR, but here's an overview of the changes for better understanding, which is somewhat grouped in the commits:

- Refactors the test cases so that we use parameterized data: This reduces code duplications among different test cases and we can simply manage the test data.
- Adds more test cases that previously weren't available: This helped to reproduce the bug and also added more test scenarios.
- Migrate the project variables processor to use python-gitlab: This would be required later anyways and by doing it now we avoid modifying shared code (`MultipleEntitiesProcessor`), which can unintentionally affect other processors and introduce other bugs in case of lack of tests.
- Cleanup codes as a result of migration to python-gitlab

related-to: #533


amimas:
I took a look at the group variable processor. As I originally suspected, I was able to reproduce the same bug. Also, the processing logic is almost exactly same. So, decided to refactor the PR a bit more. Now, have a shared logic in `variables_processor`, which is used by both group and project level variable processor. This allowed avoiding a lot of code duplication. Only duplication exists now is in the corresponding acceptance tests.

amimas:
Not sure why the tests for group variables are failing. Can't seem to reproduce the failure when I run them on my local machine. :(

amimas:
The failed tests for "group variables" are basically because of certain variables already existing - preventing from setting up the necessary test scenario.

https://github.com/gitlabform/gitlabform/actions/runs/13415602358/job/37475593296?pr=962

Added some debug logging. For example - make sure existing variables have been deleted before starting test and double check if those exists:

```
Checking previously deleted variable <class 'gitlab.v4.objects.variables.GroupVariable'> => {'variable_type': 'env_var', 'key': 'FOO', 'value': '123456', 'hidden': False, 'protected': False, 'masked': False, 'raw': False, 'environment_scope': '*', 'description': None}
...
...
DEBUG    urllib3.connectionpool:connectionpool.py:544 http://localhost:80 "GET /api/v4/groups/99/variables/FOO?filter%5Benvironment_scope%5D=%2A HTTP/1.1" 404 41
DEBUG    root:test_group_variables.py:51 The variable <class 'gitlab.v4.objects.variables.GroupVariable'> => {'variable_type': 'env_var', 'key': 'FOO', 'value': '123456', 'hidden': False, 'protected': False, 'masked': False, 'raw': False, 'environment_scope': '*', 'description': None} was deleted successfully
```

Clearly the cleanup is working, but the tests are still failing complaining about the variables with same key/name already exists:

```
FAILED tests/acceptance/standard/test_group_variables.py::TestGroupVariables::test__variables[test_case_5_single_var_multiple_scopes] - gitlab.exceptions.GitlabCreateError: 400: {'key': ['(FOO) has already been taken']}
```

I ran the tests about many times locally, but can't seem to reproduce the issue. Made sure I'm using same gitlab version as in CI right now (v17.8.2).

🤷

amimas:
I managed to reproduce the group variable related test failures on my local machine after I did a fresh setup of the python virtual environment. It seems to be a bug in python-gitlab library. Opened the following issue, which is currently a blocker for this PR.

https://github.com/python-gitlab/python-gitlab/issues/3138

gitlabform/processors/util/variables_processor.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def __init__(self, needs_update: Callable):
+        self.needs_update: Callable = needs_update
+
+    # Overload *variants* for 'get_variables_from_gitlab'.
+    # These variants give extra information to the type checker.
+    # They are ignored at runtime.
+    # See: https://mypy.readthedocs.io/en/stable/cheat_sheet_py_typed.html#overloading
+    @overload
+    def get_variables_from_gitlab(self, group_or_project: Group) -> List[GroupVariable]: ...
```
TimKnight-DWP, line 13:
```suggestion
    def get_variables_from_gitlab(self, group: Group) -> List[GroupVariable]: ...
```

Unless it genuinely can take a project in which case the typing needs correcting

gitlabform/processors/util/variables_processor.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @overload
+    def get_variables_from_gitlab(self, group_or_project: Project) -> List[ProjectVariable]: ...
```
TimKnight-DWP, line 16:
```suggestion
    def get_variables_from_gitlab(self, project: Project) -> List[ProjectVariable]: ...
```

As above, I see now @overload is doing interesting work based on the var type

New thing in Python for me 👍

amimas: Yeah, same here. I'm also learning through this. I think I needed to do this so that typing works correctly.

amimas:
@TimKnight-DWP  - Could you please review this?

Finally managed to get the CI issue resolved. The problem was the environment scope related tests for group variables require premium license. I had those tests in the standard test suite. I forgot that the standard test suite doesn't enable the premium license when they are executed in GitHub actions. That's why the tests kept failing in CI but was running fine in my local machine.

Let me know if you have any questions or would like to discuss the changes further.

TimKnight-DWP (approved):

Issue #942, opened 2025-02-04
Breaking inheritance validation error in 4.1.2


gosoly:
We have the higher level config: 
```
  "viator/infrastructure/*":
    project_settings:
      jobs_enabled: true
    project_push_rules:
      max_file_size: 20 # MB
    branches:
      master: *default_branch_settings
      develop: *default_branch_settings
      production: *default_branch_settings
      main: *default_branch_settings
    merge_requests_approval_rules:
      inherit: false
      infrastructure:
        name: "Any approver in the Infrastructure team"
        approvals_required: 1
        groups:
          - viator/org/infrastructure
      enforce: true
```
and it applied the merge_requests_approval_rules on below fine without `merge_requests_approval_rules`
```
"viator/infrastructure/terraform/database-users"
    .....
```
4.1.2 show error 
```
🏗 GitLabForm version: 4.1.2 = the latest stable 😊
Warning: Connected as non-admin. You may encounter permission issues.
Running in dry-run mode...
>>> Getting requested groups or projects...
Error: The inheritance-break flag set in "viator/infrastructure/aws/aws-workspace-scripts" under key "merge_requests_approval_rules" is invalid
because it has no higher level setting to inherit from.
```
project config:
```
  "viator/infrastructure/aws/aws-workspace-scripts":
    merge_requests_approval_rules:
      inherit: false
      infrastructure_or_devx: *infrastructure_or_devx_merge_requests_approval_rule
      enforce: true
```

These same configuration show no above error when run using 4.1.1

amimas:
Hi @gosoly - Could you please clarify or post the full config? It was kind of hard to follow the issue you described. I'll try to rephrase it below and let me know if I got it wrong.

Your config looks like this (removed a few things that aren't relevant):

```yaml
  "viator/infrastructure/*":
    merge_requests_approval_rules:
      inherit: false
      infrastructure:
        name: "Any approver in the Infrastructure team"
        approvals_required: 1
        groups:
          - viator/org/infrastructure
      enforce: true
  "viator/infrastructure/aws/aws-workspace-scripts":
    merge_requests_approval_rules:
      inherit: false
      infrastructure_or_devx: *infrastructure_or_devx_merge_requests_approval_rule
      enforce: true
```

What command did you use when you ran gitlabform? I understand you're getting the following error message right now:

```
Error: The inheritance-break flag set in "viator/infrastructure/aws/aws-workspace-scripts" under key "merge_requests_approval_rules" is invalid
[...]

PR #969, opened 2025-03-03
test: fix issues with eventual consistency in tests and cover an mr_approval_rules use case

TimKnight-DWP:
- #942


TimKnight-DWP:
hmmm tests passed locally when I ran them all, will investigate

adam-moss (approved):
LGTM

PR #982, opened 2025-03-23
docs: fix changelog for v4.2.0

amimas:
This PR fixes the changelog for v4.2.0. Original changelog had some duplicate information from prior version and also missed some other entries. Updates from this PR should make it more accurate.

After this PR is merged, we should manually copy/paste the updated changelog to existing github release details: https://github.com/gitlabform/gitlabform/releases/tag/v4.2.0


gdubicki (approved):
Thank you, @amimas! But I don't think this required a PR, you could have just pushed this change to main, IMHO. 😊

Issue #541, opened 2023-05-05
How to add users ONLY to top-level group?


felsgaertner:
Hi,  as GitLab inherits permissions downwards to descendant groups, it is sufficient to only add members to one group.

GitLabForm adds users to all descendant groups, though it is not required. How do I prevent this?

So if have these groups:
~~~text
A
A/1
A/B
A/C
A/C/2
~~~

and add members to the configuration:

~~~yaml
projects_and_groups:
  '*':
    group_members:
      users:
        vip:
          access_level: owner
~~~

I end up with user "vip" assigned to all groups. This example has just A as root, but there could by more children next to B and C.

I just want the user assigned to group A.

amimas:
> GitLab inherits permissions downwards to descendant groups, it is sufficient to only add members to one group.

Maybe I'm just mis-interpreting this. But just to clarify, subgroups will always inherit members configured in the parent group. That's how Gitlab works.

> I just want the user assigned to group A.

In your example all the subgroups under A will inherit it.

amimas:
You could configure members on the 2nd level. For example:

A/1
A/B
A/C

You can configure the "vip" user to only A/1.

There's also a `inherit: false` config in gitlabform that can be used for breaking config inheritance from higher level. Checkout the gitlabform docs.

felsgaertner:
> > GitLab inherits permissions downwards to descendant groups, it is sufficient to only add members to one group.
> 
> Maybe I'm just mis-interpreting this. But just to clarify, subgroups will always inherit members configured in the parent group. That's how Gitlab works.
> 
> > I just want the user assigned to group A.
> 
> In your example all the subgroups under A will inherit it.

Yes, that's how GitLab works. But gitlabform assigned the members to the subgroups as well - even though it is not needed and ignored by gitlab.

felsgaertner:
> You could configure members on the 2nd level. For example:
> 
> A/1 A/B A/C
> 
> You can configure the "vip" user to only A/1.
> 
> There's also a `inherit: false` config in gitlabform that can be used for breaking config inheritance from higher level. Checkout the gitlabform docs.

Yes, I know. But as I might have more than only these groups, there could be more subgroups, I do not want to configure each one repeatedly.

adam-moss:
Since you can't break inheritance on users we assign them at the project level not on groups.  It has taken a while but we're almost free of group-level users other than GATs.

[...]

Issue #842, opened 2024-10-07
Option to disable inheritance


kriko:
GitLabForm has a feature to apply settings from a parent group to sub-groups, however this can lead to unexpected configuration that is not desired.

We are using GitLabForm to manage configuration of a hierarchical GitLab setup, with many levels of sub-groups. However, when a parent group as a set of say, badges, these will also be configured for sub-groups - this in turn ends up with duplicate or triple amount of badges for GitLab projects (since badges are additive in GitLab by default - parent group badges are displayed on child levels).

One way to work around this issue is by adding `inherit: false` option to the sub-group level, however this quickly becomes a nuisance, because it would require adding this configuration parameter to all sub-groups and if we add it for a configuration element that is not present in the parent, we quickly get the following error:
```
Error: The inheritance-break flag set in "parent/sub1/sub2/sub3" under key "group_members" is invalid
because it has no higher level setting to inherit from.
```

Thus, to work around this issue, I propose a global level configuration for GitLabForm, to disable inheritance for all properties.

eg.
```
config_version: 3
inherit: false # Disable config inheritance on a global level

gitlab:
  url: https://gitlab.yourcompany.com
  # alternatively use the GITLAB_TOKEN environment variable
  token: "<private token of an admin user>"

```

amimas:
Hi @kriko - Thanks for opening the issue.

> GitLabForm has a feature to apply settings from a parent group to sub-groups, however this can lead to unexpected configuration that is not desired.

I want to make a correction here because the terminologies being used can be confusing or misleading to someone who's new (or not) to gitlab or gitlabform.

GitLabForm does not have a feature to apply settings from "parent group" to "sub-group". That's a feature of GitLab itself. You configure a particular setting at the group level (i.e. branch protection setting) and that setting gets applied to (or inherited by) all the projects under that group, including those in the subgroup.

GitLabForm applies settings to a group or project as per the config file used for its execution. To minimize/simply the config, it has a feature of writing configs using a hierarchy. When it runs, it basically generates the effective config from that hierarchy for every entity and then applies the settings accordingly. For more details, see the [config reference here](https://gitlabform.github.io/gitlabform/reference/#projects-and-groups-configuration).

As for your suggestion of adding a feature to config syntax, let's revisit the solution later. You gave an example of an issue when configuring badges. Could you please share a sample config to clarify the issue? I'm wondering whether it's a feature for `badges` config as opposed to a global option, which can introduce other unintended consequences.

[...]

Issue #398, opened 2022-08-15
Filter `projects_and_groups` configuration using project topics


armingerten:
## Problem description

The current [configuration syntax](https://gitlabform.github.io/gitlabform/reference/#projects-and-groups-configuration) allows the usage of `*` wildcards to apply a configuration to all projects or groups (or group/subgroup). However, sometimes multiple projects that share a common characteristic are scattered across multiple (sub-)groups. This makes it hard to apply the same configuration to multiple projects with said common characteristic without moving them into the same (sub-)group.

Imagine the following structure:
```
my_group
└─ my-product-a (group)
   └─ cool-python-app (python project)
   └─ really-cool-java-app (java project)
└─ my-product-b (group)
   └─ awesome-python-app (python project)
   └─ cool-application (python project)
```
In this scenario, I would like to apply a configuration to all python projects... (and only to the python projects!)

## Implementation proposal

GitLab allows adding "topics" to a project. Topics are single strings that annotate a project and allow searching / filtering for a topic. As described in the example above, all python projects could be annotated with the a topic identified as `python`.

This proposal would introduce a new configuration syntax that filters based on topics. The topic filter is indicated by the `?` character that follows appended to the `*` wildcard. The topic filter then allows specifying one or multiple topics (separated by `,`) that are required for the specified configuration:

```yaml
projects_and_groups:
  "*?<topic>":
    # common-level config filtered by <topic>
  group_1/*?<topic>:
    # group-level config filtered by <topic>
  group_2/*?<topicA>,<topicB>:
    # group-level config filtered by <topicA> and <topicB>
```

### Example configuration

```yaml
projects_and_groups:
  "my_group/*?python":
    # common-level config for python projects
```

amimas:
One work-around could be yaml anchors. You can move the shared configs into one or more yaml anchor and then apply them wherever needed.

armingerten:
> One work-around could be yaml anchors. You can move the shared configs into one or more yaml anchor and then apply them wherever needed.

That is true in terms of re-usability of configuration. I was thinking more in the direction of not having to explicitly specify my projects in the `config.yml` but rather apply the configuration based on a predicate (e.g. "a python project").

gdubicki:
Hi @armingerten!

Thank you for the detailed feature proposal.

We will take it into consideration when designing v4 of the app, but I cannot commit to any deadlines for it, sorry (see #343 for more info).

gdubicki:
[...]

Issue #325, opened 2022-01-13
Expand wildcard support


syphernl:
Take for instance this repositories on GitLab:

```
- group/otherproject
- group/something/else
- group/project/app
- group/project/services/backend
- group/project/services/frontend
- group/project/services/database
- group/subgroup/project/app
- group/subgroup/project/services/backend
- group/subgroup/project/services/frontend
- group/subgroup/project/services/database 
```

I'd like to have Gitlabform process _all_ repo's except for the `app` ones

 I have tried this:

```
skip_projects:
  - "app"
```

and

```
skip_projects:
  - "*/app"
```

Unfortunately this did not work.

I have also tried to set other settings for this particular type of project like this:

```
projects_and_groups:
"group/*/app":
    project_settings:
      default_branch: develop
```

This results in getting an error that the project does not exist.

Is this something that is already possible but I'm simply doing it wrong or is this something that can be implemented?

sy-be:
We are having an issue where GLF would "reactivate" projects marked for deletion. When projects are marked for deletion - the project name changes to something like: `foobar-deleted-1234`. Upon subsequent GLF run and if targeted as part of the group - it causes deleted projects to get reactivated.

I'd love to have the following syntax supported under `skip_projects`:
```
skip_projects:
  - *-deleted-*
```

or maybe add a regex support.

PR #986, opened 2025-03-27
feat: Allow "group/subgroup/" as keys to provide configuration that is not inherited recursively

Anedaar:
This change allows groups and subgroups to be referenced as 

    projects_and_groups:
      group/:
        <config_group>
      group/subgroup:
        <config_subgroup>

<config_group> will only affect group and projects directly in group, but not any subgroups or projects in subgroups.

Similarly, <config_subgroup> will only affect group/subgroup and projects in it, but not any subsubgroups.

This allows e.g. for a group to be public, but for all subgroups to be private:

    projects_and_groups:
      group/:
        group_settings:
          visibility: public
      group/subgroup:
        group_settings:
          visibility: internal
      group/*:
        group_settings:
          visibility: private

- group will be public
- group/subgroup will be internal
- all other subgroups of group will be private

The feature as implemented recognizes the "inherit: false" flag. Keys marked "inherit: false" in "group/" will not be inherited from "group/*":

    projects_and_groups:
      group_1/*:
        deploy_keys:
          key_a:
            key: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDB2QKx6BPzL...
            title: global_key # this name is show in GitLab
            can_push: false
      group_1/:
        deploy_keys:
          inherit: false
          key_b:
            key: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDtbyEK66RXg...
            title: group_key # this name is show in GitLab
            can_push: false

Subgroups of group_1 will have "key_a", group_1 itself will only have "key_b".

This feature should also solve issue #541.
This feature will probably become irrelevant once regexes are implemented.


gdubicki:
Thanks a very interesting contribution, @Anedaar, thank you! Have you perhaps compared it to our analysis of the alternatives here https://github.com/gitlabform/gitlabform/wiki/Current-and-proposed-way-of-selecting-targets ?

amimas:
I feel this PR is introducing breaking change. But more importantly, I don't understand the problem being solved based on the description. It just jumped into solutioning. For example in the proposed config, now it's not clear at all whether a given section is a subgroup or a project. It adds extra cognitive load to interpret the config and it can lead to lots of issues/confusions.

I suggest opening a new issue or engage in an existing issue. Make the problem statement as clear as possible. We already have feature for breaking inheritance.

jimisola:
We also cover this in https://github.com/gitlabform/gitlabform/wiki/Configuration-Syntax-v5-Draft#alternatives-for-includeexclude

I think we shall handle an improved syntax for project/group selection in one go.

amimas:
Hi @Anedaar . Thanks for all of that details. There's a lot to unpack in there. It's possible I might have missed a few things from your reply.

Regarding whether this is a breaking change or not - I'm still leaning towards this being a breaking change. The updated config syntax changes how we would read the config and what we'd expect to happen or interpret the config.

As for your first example: I think it might be a matter of expectation. For example: if all subgroups under `example` group should have `private` visibility, then the config should be following (as you already noted), to avoid config duplication:

```yaml
    "external/*":
      group_settings:
        visibility: private

    "external/CompanyA/*":
      group_settings:
        description: Projects with CompanyA.

    "external/CompanyB/*":
      group_settings:
        description: Projects with CompanyB.
```

To be fair, I do understand some of the limitations you've mentioned. GitLabForm is not perfect. For me personally, config readability is extremely important. A common proposal on this topic I've seen makes it impossible to infer what the config is doing from readability perspective. For example, the following maybe fine for automation but as I said, when reading the config, how do I know what's a project and what's a group?

```
<A>, when A:
.... is a project: Refers only to this project (no changes).
.... is a group: Invalid reference (no changes, see below to not break running GitLabForm).
<A>/, when A:
.... is a project: Invalid reference (no changes—well, sort of, currently it’s not just an "invalid reference").
.... is a group: Refers to this group and its projects, but NOT subgroups.
<A>/*, when A:
.... is a project: Invalid reference (no changes).
.... is a group: Refers to this group, its projects, its subgroups, and their projects (no changes).
```

Currently the purpose of `<group>/*` is to be used for common settings for all resources under this group including the host group itself. If something is not common across the board (e.g. group/project description), it shouldn't be there or if there's an exception `inherit: false` is to be used for that specific resource. It's clear and concise (in my opinion) and I'm not sure if this is clearly understood by gitlabform users.

Also, there are other proposal about expanding on the wildcard usage for selecting a resource (e.g. group/project) to be updated. See #398 and #325 . That's why we probably should think about overall changes to config syntax, in terms of resource selection, so that all of these use cases are properly evaluated and then come up with a proposal/design, and then implementation. I think it's a huge undertaking that will require lots of time and effort. Will you be able to help lead this effort?

What do you think @jimisola @gdubicki @TimKnight-DWP ?

TimKnight-DWP:
@amimas I think we have definetly fallen afoul of specifying some "group" settings under the `group/*` heading, perhaps we need to be clearer in the documentation and feedback, some kind of linter?

TimKnight-DWP:
@amimas @Anedaar 
One example I can think of which would be fixed by this is `integrations`

```
group/*:
  integrations:
      gitlab-slack-application:
        use_inherited_settings: false
        merge_requests_events: true
        merge_request_channel: maintainers
```

In this scenario _all_ projects get updated with the specific merge_request_channel etc, however what I want, is to set that integration at the group level, and to have the projects set `use_inherited_settings: true` -> which there is currently? no way to achieve in GLF

itewk:
this would solve:
* https://github.com/gitlabform/gitlabform/issues/1427
* https://github.com/gitlabform/gitlabform/issues/842

i am not really understanding the case against this. 

we really need a way to set settings on a speicfic group only and not all the nested groups, and that is currently impossible. having the three syntaxes seems clear to me:

"foo/bar/hello": -> refers to the project "hello" under the groups "foo/bar"
"foo/": -> refers to the group "foo" and only the group "foo"
"foo/*" -> refers to the group "foo" and all nested groups and projects

PR #987, opened 2025-03-28
feat: add support for group and project avatar

### LINKED ISSUE #698: Uploading Project Avatar
I'm having trouble uploading avatars for my projects. I always get `{"error":"avatar is invalid"}` error.

I can't change using curl to change one by one.

I think that the problem is that gitlabform can't find my image file but can't figure out the correct url. Tried relative, absolute, with starting '@'...

rickbrouwer:
Add support for group and project avatar.

### Checklist

- [X] Tests have been added

Fixes #698


docs/reference/avatar.md
```diff
@@ -0,0 +1,49 @@
...
+This section purpose is to manage the project avatar - the image that is displayed next to your project name in GitLab.
+
+
+The value is a path to an image file relative to the directory where GitLabForm is run. If the value is explicitly set to "", the existing project avatar will be removed.
+
+Example:
+```
+projects_and_groups:
+  group_1/project_1:
+    project:
```
gdubicki, line 14: You cannot use `project` as the key here as it will collide with the `ProjectProcessor` that uses it. You should use a different key then, like `project_avatar`.

docs/reference/avatar.md
```diff
@@ -0,0 +1,49 @@
+# Avatars
+
+## Project Avatars
+
+This section purpose is to manage the project avatar - the image that is displayed next to your project name in GitLab.
+
+
+The value is a path to an image file relative to the directory where GitLabForm is run. If the value is explicitly set to "", the existing project avatar will be removed.
```
gdubicki, line 8: To be consistent with how we operate on files in the `files` section, we should allow both absolute and relative paths, but the relative ones should refer to the config file location, not the current working directory, see:

https://github.com/gitlabform/gitlabform/blob/873524007730b883580f7c976577e71ef930ce33/gitlabform/processors/project/files_processor.py#L103-L107

docs/reference/avatar.md
```diff
@@ -0,0 +1,49 @@
...
+      avatar: "images/project-logo.png"
+```
+
+To remove an avatar, set the value to "":
+
+```projects_and_groups:
+  group_1/project_1:
+    project:
+      avatar: ""
+```
```
gdubicki, line 24: Again, for consistency with how we delete config elsewhere, I think we should rather allow this syntax:

```
projects_and_groups:
  group_1/project_1:
    project_avatar:
      delete: true
```

gitlabform/gitlab/group_avatar.py
```diff
@@ -0,0 +1,51 @@
...
+
+        Args:
+            group_name: The group identifier (group or subgroup path)
+        """
+
+        group_id = self._get_group_id(group_name)
+        url = f"{self.url}/api/v4/groups/{group_id}"
+
+        # Put without avatar file
+        return self.session.put(url, data={"avatar": ""}, timeout=self.timeout)
```
gdubicki, line 51: This works, of course, but we are in a process of moving away from homegrown GitLab library into the standard python-gitlab, so instead of all this the GroupAvatarProcess should rather use the code from it.

See https://python-gitlab.readthedocs.io/en/stable/gl_objects/groups.html, "Set the avatar image for a group:".

The same applies to editing project avatar.

Sorry if this was not clear enough in our contribution docs 😞, we will improve those!

docs/reference/avatar.md
```diff
@@ -0,0 +1,49 @@
...
+This section purpose is to manage the group avatar - the image that is displayed next to your group name in GitLab.
+
+
+The value is a path to an image file relative to the directory where GitLabForm is run. If the value is explicitly set to "", the  existing group avatar will be removed.
+
+Example:
+```
+groups:
+  group_1:
+    group:
```
gdubicki, line 36: Although there is no group processor that would use `group`, I think we prefer to not use it only for avatars, so I suggest using a name like `group_avatar` here.

gdubicki (changes requested):
Thank you for your contribution, @rickbrouwer!

We would love to get it merged, but for the good of the project as a whole and long-term maintainability we would need to make some changes first to make it work more consistently with other parts of the app and more predictably for existing users.

Sorry if our contribution docs didn't make some of those things clear, we will do our best to improve them.

In case you don't have the time or will to improve the things we pointed out, we can help but I encourage you to try to do it first!

docs/reference/avatar.md
amimas: It'd be nice if we discuss the overall config syntax. I noticed the referenced issue didn't conclude what the config syntax should look like. Going through that issue (#698), I think the `ProjectSettingsProcessor` can be refactored so that it can handle uploading avatar file. In that case the config might look something like:

```yaml
projects_and_groups:
  group_1/project_1:
    project_settings:
      visibility: internal
      avatar:
         file: <path to image file>
         delete: <true|false>
```

I haven't thought it through yet though. Of course, we should also consider what the group related avatar config would look like.

That's why I suggest we discuss in the above issue about the config syntax and come to a conclusion before making code changes.

jimisola: I agree. I learned my lesson with topics. Let's decide on the config syntax before making approving an implementation.

docs/reference/avatar.md
rickbrouwer: Yeah, agree! I'd love to hear what you guys want. The choice doesn't really matter to me, for example under the `project_settings` or is own `project_avatar`.

gdubicki: Thanks for being open to changes, @rickbrouwer! See the thread above for more talk about that.

docs/reference/avatar.md
gdubicki: Good point, @amimas! I think you proposal to move it under project_settings is perfect, but let’s in fact discuss the whole thing in the issue.

TimKnight-DWP (approved):

rickbrouwer:
FYI.

Support for removing avatars has been implemented in python-gitlab with [this PR](https://github.com/python-gitlab/python-gitlab/pull/3175).

I am now waiting for a release of the new python-gitlab before this PR can continue. The previous release of python-gitlab was a while ago. I expected a release on the 28th due to their [release process](https://github.com/python-gitlab/python-gitlab/blob/main/CONTRIBUTING.rst?plain=1#L175-L177), but I see that their release process has not been up to par with what is described for a while now.

rickbrouwer:
@amimas 

It took a while, but python-gitlab 6.0.0 has been released and is merged here, so I think this PR can continue.
Could you run the final tests?

Thanks 🙏🏼

docs/features.md
```diff
@@ -4,6 +4,7 @@ GitLabForm enables you to manage the [(GitLab's) Application Settings](reference
 
 * Group:
     * [Badges](reference/badges.md#group-badges)
+    * [Avatar](reference/avatar.md#group-avatar)
```
amimas, line 7: Is the anchor link correct? The heading in `avatar.md` is `Group Avatars`. I suspect you need:

```suggestion
    * [Avatar](reference/avatar.md#group-avatars)
```

docs/features.md
```diff
@@ -14,6 +15,7 @@ GitLabForm enables you to manage the [(GitLab's) Application Settings](reference
 * Project:
     * [Archive/unarchive](reference/archive_unarchive.md)
     * [Badges](reference/badges.md#project-badges)
+    * [Avatar](reference/avatar.md#project-avatar)
```
amimas, line 17: Is the anchor link correct? The heading in `avatar.md` is `Group Avatars`. I suspect you need:

```suggestion
    * [Avatar](reference/avatar.md#project-avatars)
```

docs/reference/avatar.md
amimas, line 1: Let's add this new docs to the site navigation pane. You need to update the `mkdocs.yml` file. I suggest adding it above this line or propose another line that you think is ideal.

https://github.com/gitlabform/gitlabform/blob/30a957659fa62ffb94d7cc8111dacb4c68a8f25b/mkdocs.yml#L13

gitlabform/processors/group/group_settings_processor.py
```diff
@@ -15,6 +15,14 @@ def _process_configuration(self, group: str, configuration: Dict):
 
         gitlab_group: Group = self.gl.get_group_by_path_cached(group)
 
+        # Process avatar separately before other settings
+        self._process_group_avatar(gitlab_group, configured_group_settings)
+
+        # Remove avatar from config to prevent it from being processed in the standard way
+        if "avatar" in configured_group_settings:
+            configured_group_settings = configured_group_settings.copy()
+            del configured_group_settings["avatar"]
```
amimas: For my learning, is there any benefit to copying the settings before the `avatar` key/object is removed? The copied dict is being reassigned to the same variable.  Could you help me understand why it's needed?

gitlabform/processors/group/group_settings_processor.py
```diff
@@ -28,3 +36,29 @@ def update_group_settings(gitlab_group: Group, group_settings_config: dict):
...
             gitlab_group.__setattr__(key, value)
             gitlab_group.save()
+
+    def _process_group_avatar(self, gitlab_group: Group, group_settings_config: dict) -> None:
+        """Process group avatar settings from configuration."""
+        avatar_path = group_settings_config.get("avatar")
+        if avatar_path is None:
+            return
+
+        debug(f"Processing group avatar configuration: {avatar_path}")
```
amimas: What do you think of moving the debug line at the beginning of the function? Also, maybe add another debug line before returning so that it's clear that no further processing is done because `avatar` config is not provided.

Thoughts?

gitlabform/processors/group/group_settings_processor.py
```diff
@@ -28,3 +36,29 @@ def update_group_settings(gitlab_group: Group, group_settings_config: dict):
...
+        debug(f"Setting group avatar from file: {avatar_path}")
+        try:
+            with open(avatar_path, "rb") as avatar_file:
+                gitlab_group.avatar = avatar_file
+                gitlab_group.save()
+            debug("Group avatar uploaded successfully")
+        except FileNotFoundError:
+            debug(f"Group avatar file not found: {avatar_path}")
+        except Exception as e:
+            debug(f"Error uploading group avatar: {str(e)}")
```
amimas: Is it possible to incorporate the `self._needs_update` method so that avatar processing is only done if necessary? Otherwise right now setting/removing of avatar will be done on every execution. Let's see if we can avoid that or let me know if I missed it.

The logic doesn't necessarily have to be in the lines I selected for this comment.

tests/acceptance/standard/test_avatar.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def test__project_avatar_delete(self, project):
+        # First set an avatar
+        config = f"""
+        projects_and_groups:
+          {project.path_with_namespace}:
+            project_settings:
+              avatar: "{self.test_image_path}"
+        """
+        run_gitlabform(config, project)
```
amimas: Since we're using the `project` fixture, we shouldn't need to set the avatar first. The `project` fixture is scoped to the test class. So, earlier test would have set the avatar already if that passed, because the project remains in the state that the last test left it. That's why here, we should be able to just go ahead and attempt deleting the avatar, but still we should verify that the avatar exists though.

tests/acceptance/standard/test_avatar.py
amimas, line 1: I think current pattern is to have separate test file for group vs project features. To keep things consistent, can we move the group avatar to a separate test? Please let me know if i'm mistaken. Thanks.

docs/features.md
amimas: Or, maybe fix the heading in the doc since there can only be one avatar.

amimas: Resolved in latest update.

docs/features.md
amimas: Resolved in latest update.

docs/reference/avatar.md
amimas: Resolved in latest update.

gitlabform/processors/group/group_settings_processor.py
amimas: Resolved in latest update.

gitlabform/processors/group/group_settings_processor.py
rickbrouwer: Currently working on the updates. Before deleting I can do a check to see if it is set. For placing it is more difficult. After all, the url that you provide in the config is different from the final avatar_url. Now in theory you can retrieve the avatar_url and compare it, but that may not say anything yet. After all, if the company_logo.png changes to company_logo.png but the logo is actually different then it skips it. That should not be. Then you can still download the logo and do a hash comparison, but that seems to me so much overhead that you might as well always place the avatar.

gitlabform/processors/group/group_settings_processor.py
amimas: Thanks for the update on this.

rickbrouwer:
@amimas 

Thanks for the good and correct feedback! I think I have implemented almost everything. If there are any comments or remarks, please let me know or code suggestions are also welcome.

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,75 @@
...
+
+        config = f"""
+        projects_and_groups:
+          {group.full_path}/*:
+            group_settings:
+              avatar: "{nonexistent_path}"
+        """
+
+        # This should run without crashing, even though the file doesn't exist
+        run_gitlabform(config, group)
```
amimas: I haven't actually thought about this before. Do you recall if this was already discussed? Why do we not want to fail when invalid avatar file path is provided in the config?

GitLabForm can continue running for processing other configs, but exit with non-zero to indicate there were failures. Otherwise, this will be a silent failure, which is not good. I think this is supposed to be the default behaviour, but not sure where/how this is configured. To validate this, additional config is needed though for the test (e.g. group setting or a project setting and validate they have been processed).

Ideally this can also be caught by the dry-run/noop feature, but we can leave that for a separate PR, if you like.

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,75 @@
...
+from tests.acceptance import (
+    run_gitlabform,
+)
+
+
+class TestGroupAvatar:
+    def setup_method(self):
+        # Use gitlabform logo
+        self.project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../.."))
+        self.test_image_path = os.path.join(self.project_root, "docs/images/gitlabform-logo.png")
```
amimas: The documentation for this feature suggests both absolute and relative path in the config is supported. Looks like the test cases are only validating absolute path. Can we add a validation for relative path too?

gitlabform/processors/group/group_settings_processor.py
```diff
@@ -28,3 +35,40 @@ def update_group_settings(gitlab_group: Group, group_settings_config: dict):
...
```
amimas: Added another comment about handling failures. Is that as simple as re-raising the exceptions.

tests/acceptance/standard/test_avatar.py
amimas: Thanks for the update on this.

tests/acceptance/standard/test_avatar.py
amimas: Thanks for the update on this.

gitlabform/processors/group/group_settings_processor.py
amimas: That's a good point @rickbrouwer . Thanks for the explanation. May I suggest adding this as a brief comment in the code? This is to help all of us in the future :)

gitlabform/processors/group/group_settings_processor.py
rickbrouwer: Done :)

tests/acceptance/standard/test_group_avatar.py
rickbrouwer: Done :)

tests/acceptance/standard/test_group_avatar.py
rickbrouwer: You are absolutely right, changed it :)

docs/reference/avatar.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+## Group Avatar
+This section purpose is to manage the group avatar - the image that is displayed next to your group name in GitLab.
+The value is a path to an image file. You can use both absolute paths and relative paths.
+
+Example:
+```
+groups:
+  group_1:
```
amimas: Correct config for targeting a given group setting.

```suggestion
  group_1/*:
```

docs/reference/avatar.md
```diff
@@ -0,0 +1,43 @@
...
+    group_settings:
+      avatar: "images/group-logo.png"
+```
+
+To remove an avatar:
+```
+groups:
+  group_1:
```
amimas: Correct config for targeting a given group setting.

```suggestion
  group_1/*:
```

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,292 @@
...
+            group_settings:
+              avatar: "{self.test_image_path}"
+        """
+        run_gitlabform(config_set, group)
+
+        # Refresh group data to get current state
+        group = group.manager.get(group.id)
+
+        # Verify that avatar exists
+        assert group.avatar_url is not None, "Avatar should exist after setting it"
```
amimas: I think we'd expect the group avatar to be set by the previous test, right? Do really need to run through gitlabform again before we begin the test? Wouldn't the last assert be sufficient before the test begins?

tests/acceptance/standard/test_project_avatar.py
```diff
@@ -0,0 +1,292 @@
...
+            project_settings:
+              avatar: "{self.test_image_path}"
+        """
+        run_gitlabform(config_set, project)
+
+        # Refresh project data to get current state
+        project = project.manager.get(project.id)
+
+        # Verify that avatar exists
+        assert project.avatar_url is not None, "Avatar should exist after setting it"
```
amimas: I think we'd expect the project avatar to be set by the previous test, right? Do really need to run through gitlabform again before we begin the test? Wouldn't the last assert be sufficient before the test begins?

tests/acceptance/standard/test_group_avatar.py
amimas: Thanks for the updates.

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,292 @@
...
+from tests.acceptance import (
+    run_gitlabform,
+)
+
+
+class TestGroupAvatar:
+    def setup_method(self):
+        # Use gitlabform logo
+        self.project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../.."))
+        self.test_image_path = os.path.join(self.project_root, "docs/images/gitlabform-logo.png")
```
amimas: Minor comment. For clear readability of the test codes, can we rename the variable? Maybe something like `test_image_path_absolute`?

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,292 @@
...
+        config_absolute = f"""
+        projects_and_groups:
+          {group.full_path}/*:
+            group_settings:
+              avatar: "{self.test_image_path}"
+        """
+        run_gitlabform(config_absolute, group)
+
+        group = group.manager.get(group.id)
+        assert group.avatar_url is not None
```
amimas: I suggest removing test 1. As you said, it's already tested in previous test case/function. Don't need to repeat it. We can break the remaining tests into separate functions.

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,292 @@
...
+        # Test 2: Simple relative path
+        original_cwd = os.getcwd()
+        temp_dir = tempfile.mkdtemp()
+
+        try:
+            os.chdir(temp_dir)
+
+            # Copy image to temp directory
+            temp_image_name = "relative_test_avatar.png"
+            shutil.copy2(self.test_image_path, temp_image_name)
```
amimas: Double-checking the logic here. Are we actually testing "relative path" usage? Based on the documentation on `tempfile.mkdtemp()`, it will return you the absolute path of the temp directory.

https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp

Please let me know if I'm looking at the wrong documentation.

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,292 @@
...
+
+            # Use non-existent relative path
+            nonexistent_relative_path = "images/nonexistent_avatar.png"
+
+            config = f"""
+            projects_and_groups:
+              {group.full_path}/*:
+                group_settings:
+                  avatar: "{nonexistent_relative_path}"
+            """
```
amimas: In this test case, do we need those logic around `os.getcwd`, `tempfile.mkdtemp`? Not sure how it's being used in the tests. Could you please clarify what I'm missing?

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,292 @@
...
+        # Verify it's the expected exit code for processing errors
+        assert exc_info.value.code == 2
+
+        # Refresh group data and verify description wasn't changed
+        updated_group = group.manager.get(group.id)
+        current_description = getattr(updated_group, "description", "")
+
+        # Description should remain unchanged since the process failed
+        assert current_description == original_description
+        assert current_description != test_description
```
amimas: My understanding was that gitlabform will continue to process other settings even if avatar setup fails. So, in this case I expected gitlabform exiting with error code `2` due to avatar setup failure but the description is still configured.

Maybe I'm mistaken. Are you able to test what happens with currently released gitlabform version? Maybe an invalid config for `group_setting` but another setting (i.e. `group_hooks`)? Will group hooks still be processed if group setting fails?

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,292 @@
...
+            projects_and_groups:
+              {group.full_path}/*:
+                group_settings:
+                  avatar: ""
+            """
+            run_gitlabform(config_remove, group)
+            group = group.manager.get(group.id)
+
+        # Store the current avatar state (should be None or gravatar)
+        original_avatar_url = group.avatar_url
```
amimas: I think we discussed in previous round of reviews. There are fixtures that are scoped to the test function only. So, if we use the `group_for_function` fixture instead of `group`, we can avoid this entire logic, because `group_for_function` will be a new group created with fresh state. We can just have a single line assert to make sure group avatar is not set.

tests/acceptance/standard/test_group_avatar.py
```diff
@@ -0,0 +1,292 @@
...
+        """
+        run_gitlabform(config_delete_empty, group)
+
+        # Refresh and verify nothing changed
+        group = group.manager.get(group.id)
+        assert group.avatar_url == original_avatar_url
+
+    def test__group_avatar_generic_exception_handling(self, group):
+        """Test generic exception handling during avatar upload"""
+
```
amimas: Is this redundant?

amimas (commented):
Thanks @rickbrouwer . Added some comments based on new update. Please let me know if they make sense.

Regarding the logic that creates temp directory - what do you think of moving that to a separate function as a "fixture". The fixture can be configured to accept input to decide whether it'll include sub-directory or not.  Or the subdirectory could be a separate fixture if that's simpler. You can add the fixtures to `conftest.py` file.

If it's done as a fixture, it should make the tests more readable, in my opinion. It'll reduce code duplications in lot of places and also makes it available for other similar tests in future.

Please let me know what you think.

gitlabform/processors/group/group_settings_processor.py
amimas: Thanks!

tests/acceptance/standard/test_group_avatar.py
amimas: Looks like the upates to the tests doesn't align what I had expected. Opened a separate comment. Please let me know if my understanding is wrong.

https://github.com/gitlabform/gitlabform/pull/987#discussion_r2160370318

tests/acceptance/standard/test_group_avatar.py
amimas: @gdubicki @TimKnight-DWP - I'm not sure if I have the correct understanding. Appreciate your feedback on this. Take the following config as an example:

```yaml
projects_and_groups:
  group_1/project_1:
    project_settings:
      avatar: '/path/to/image_file.png'
      description: "Some description about this project"
```

Here, `avatar` and `description` are set by separate APIs. The `avatar` gets processed first, if it exists in the config. In this scenario if there's an exception/failure with setting the avatar, should gitlabform continue with remaining processes and finally exit with error code 2 to indicate there was a failure? Or, should gitlabform terminate right after the failure of processing the avatar?

amimas: I think I had misunderstood gitlabform behaviour. Let's look at following config:

```yaml
projects_and_groups:
  group_1/project_1:
    project_settings:
      avatar: '/path/to/image_file.png'
      description: "Some description about this project"
  group_1/project_2:
    project_settings:
      avatar: '/path/to/image_file.png'
      description: "Some description about this project"
```

When gitlabform is executed, the default value of `--terminate` is set to `False`. This means even if `group_1/project_1` fails for any reason in `project_setting`, gitlabform will still continue and process `group_1/project_2` but exit with an error at the end.

But, if we enable the terminate option (e.g. `gitlabform --terminate`) - this means immediate termination. For example: if there's an error processing `group_1/project_1` executing `project_setting`, gitlabform will not continue to process `group_1/project_2` and exit immediately.

The current architecture of GitLabForm doesn't support the above behaviour in lower level processing. Consider the following example where separate API calls are involved (i.e. 1 for avatar and 1 for rest of the setting):

```yaml
projects_and_groups:
  group_1/project_1:
    project_settings:
      avatar: '/path/to/image_file.png'
      description: "Some description about this project"
```

Here, regardless of `--terminate` parameter, if processing avatar fails, it will not process rest of the setting. Here's the summary of the current limitation.

The current GitLabForm architecture has this hierarchy:

- GitLabForm level: Controls whether to continue processing other groups/projects
- Processor level: Each processor processes all its settings or fails entirely
- Setting level: Individual settings within a processor are processed sequentially

The `--terminate` parameter only affects level 1, not levels 2 or 3.

@gdubicki  - The above is not documented anywhere. Could you please confirm if my understanding is correct?

amimas:
@rickbrouwer  - Trying to review and test things out on my end whenever I have a little time. I think this PR is very close. 

Would you be able to look into the outstanding comments above? Only comment I think you don't need to address at the moment is this: https://github.com/gitlabform/gitlabform/pull/987#discussion_r2160370318, because I think it's already working as per gitlabform's design. I'm trying to validate this myself.

rickbrouwer:
@amimas 

I've done some more updates. I think and hope we're there now :)

tests/acceptance/standard/test_group_avatar.py
TimKnight-DWP: @amimas @rickbrouwer - yes I would prefer it be consistent, if --terminate is not set, it should show a warning for failing to process the avatar and then crack on with the rest of it. Avatars feel like a nice to have compared to must have functionality of applying access control.

**That** and have Avatars be processed last in a Project or Group level?

TimKnight-DWP (changes requested):
See discussion around exiting

rickbrouwer:
> @rickbrouwer LGTM, that was the last bit I was unsure on resolution, will rebase and set to auto merge

Thanks. I did one more update, the code coverage wasn't quite right.

TimKnight-DWP:
@rickbrouwer can you revert that change to the changelog please?

rickbrouwer:
> @rickbrouwer can you revert that change to the changelog please?

Hmm, I'm just looking at it, but I don't see any changes on the changelog.md. Or am I looking at this completely wrong?

TimKnight-DWP:
Ah maybe it needs a rebase, I looked at the comparison on the last commit and showed changelog diff

amimas:
Thanks @rickbrouwer . I did look at your latest updates. Moving the avatar processing logic to be after settings processing is smart!

My main concern was whether exception from avatar processing should be raised or not. This is a continuation from this discussion thread: https://github.com/gitlabform/gitlabform/pull/987#discussion_r2187359461

I believe the exception should be raised. Otherwise it doesn't align with gitlabform's `--terminate` CLI option. By default this option is not set. I've updated your logic to raise the exception and added couple of tests for both group and avatar setting. 

```yaml
projects_and_groups:
  gitlabform_group_lustily/*:
    group_settings:
      avatar: '/path/to/nonexistent/image.png'
      description: "Some description about this project"
  gitlabform_group_useable/*:
    group_settings:
       # some config here
```

Here's what happens when `--terminate` is not used or set to false (default) with the above config. Even though first group's processing fails due to avatar processing issue, it will continue to process the 2nd group but exit with non-zero to indicate there were issues (instead of silently failing).

```sh
tests/acceptance/standard/test_group_avatar.py Reading config from the provided string.
Connected to GitLab version: 18.1.1-ee (ceb07b24cb0)
Connected as: root, admin: yes
Getting user ids for users defined in protect_environments config
Getting user ids for users defined in merge_requests_approval_rules config
>>> Getting ALL groups and projects DEFINED in the configuration...
:: # of groups to process: 2
groups: ['gitlabform_group_lustily', 'gitlabform_group_useable']
:: # of projects to process: 0
projects: []
@ (1/2) Processing group: gitlabform_group_lustily
Skipping section 'group_variables' - not in config.
Processing section 'group_settings'
Skipping section 'group_members' - not in config.
Skipping section 'group_ldap_links' - not in config.
Skipping section 'group_badges' - not in config.
Skipping section 'saml_group_links' - not in config.
Skipping section 'group_labels' - not in config.
Skipping section 'group_hooks' - not in config.
Skipping section 'group_push_rules' - not in config.
@ (2/2) Processing group: gitlabform_group_useable
Skipping section 'group_variables' - not in config.
Processing section 'group_settings'
Warning: Error occurred while processing group gitlabform_group_useable, exception:

Group avatar file not found: /path/to/nonexistent/image.png
:: # of groups processed successfully: 1
:: # of projects processed successfully: 0
:: # of groups failed: 1 
:: Failed group 2: gitlabform_group_useable 
```

If we use the same config with `--terminate` CLI option enabled, here's the result. The first group processing fails because of avatar setting issue. Because of `--terminate` gitlabform will stop immediately. The 2nd or subsequent project/resource is not processed.

```sh
 tests/acceptance/standard/test_group_avatar.py Reading config from the provided string.
Connected to GitLab version: 18.1.1-ee (ceb07b24cb0)
Connected as: root, admin: yes
Getting user ids for users defined in protect_environments config
Getting user ids for users defined in merge_requests_approval_rules config
>>> Getting ALL groups and projects DEFINED in the configuration...
:: # of groups to process: 2
groups: ['gitlabform_group_chemo', 'gitlabform_group_wager']
:: # of projects to process: 0
projects: []
@ (1/2) Processing group: gitlabform_group_chemo
Skipping section 'group_variables' - not in config.
Processing section 'group_settings'
Error: Error occurred while processing group gitlabform_group_chemo, exception:

Group avatar file not found: /path/to/nonexistent/image.png
```

Please review the 2 new commits I've pushed and what I explained above. Let me know if this aligns with you or not. It's possible we may have gaps in other processors regarding handling of exceptions. We don't really have this documented to help with incoming contributions. And it gets a bit more complicated if avatar handling is done before other settings processor and what how we'd expect the `--terminate` option to behave. This is kind of unknown right now.

Hope all of that makes sense.

@TimKnight-DWP @gdubicki - FYI.

rickbrouwer:
That looks great. Thanks so much @amimas for this explanation and addition!

amimas (approved):

PR #988, opened 2025-03-29
Inform future contributors that they should use python-gitlab

gdubicki:
as decided in https://github.com/gitlabform/gitlabform/issues/73

This should prevent cases like https://github.com/gitlabform/gitlabform/pull/987#discussion_r2019786718

### DIFF
--- a/docs/contrib/implementation_design.md
+++ b/docs/contrib/implementation_design.md
@@ -88,6 +88,10 @@ to `ProjectProcessors` in `project > __init__.py`.
 
 ### gitlabform.gitlab.*
 
+!!! important
+
+    We have decided to move away from our own homegrown GitLab API client to the `python-gitlab` library. (See [issue #73](https://github.com/gitlabform/gitlabform/issues/73) for info.) Therefore, you should **NOT** add any more code to this package!
+
 With the basics for making requests to the GitLab API covered in the `GitLabCore` class in `core.py`, all other code
 is simple (most of the time).
 


docs/contrib/implementation_design.md
```diff
@@ -88,6 +88,10 @@ to `ProjectProcessors` in `project > __init__.py`.
 
 ### gitlabform.gitlab.*
 
+!!! important
+
+    We have decided to move away from our own homegrown GitLab API client to the `python-gitlab` library. (See [issue #73](https://github.com/gitlabform/gitlabform/issues/73) for info.) Therefore, you should **NOT** add any more code to this package!
+
```
jimisola, line 94: It does not seem to be rendered as intended. It's all "code" and hence the link text is not rendered properly.

image

amimas: I think that's just how it's being rendered by GitHub. When the site is generated using MKDocs, it should not look like this as it uses special parsing following this syntax. Below is another doc example that shows the same way as here.

https://github.com/gitlabform/gitlabform/blob/793951fac9e47b145102672e01e523b1bcddf242/docs/reference/archive_unarchive.md

jimisola: Ok. Then I'll just approve.

jimisola (approved):

amimas (commented):
I suggest moving the note about using `python-gitlab` higher up. And adding another note about `SingleEntityProcessor` or `MultipleEntityProcessor` being deprecated. What do you think?

jimisola:
> I suggest moving the note about using `python-gitlab` higher up. And adding another note about `SingleEntityProcessor` or `MultipleEntityProcessor` being deprecated. What do you think?

That's better. The further up the better to minimize the risk of someone missing it.

TimKnight-DWP (approved):

gdubicki:
Thanks for the suggestions how to improve this, guys. I will merge this for now as it is to not let it wait more.

PR #1016, opened 2025-05-06
feat: implement "exclude-sections" CLI parameter

### LINKED ISSUE #1005: Add CLI parameter to skip sections
We currently have a CLI parameter `-os` to ONLY run certain sections of the config, otherwise all sections are run.

For large Gitlab instances and configurations we may want to split out into several runs such as:

`gitlabform -c ${GLF_CONFIG} ${GLF_SCOPE} -os group_settings,group_members,group_badges,group_labels,group_variables`

Process the main group settings upfront and then run gitlabform again excluding those sections. If we don't exclude the sections they will get run a second time, introducing latency.

With large repos you may need to process group_settings and group_members upfront to ensure all Access Rights are correctly provisioned before the remaining run, otherwise state can get into a bit of a mess.

Propose to add `-es` to `EXCLUDE sections` as a CLI parameter

TimKnight-DWP:
- resolves #1005

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -50,10 +50,10 @@ To view the full set of Command Line Options you can run `gitlabform --help`
 
 ### Exclude Sections
 
-To exclude certain sections of the configuration from a given gitlabform run, you can pass a list of comma-delimited names via the `--exclude_sections` parameter.
+To exclude certain sections of the configuration from a given gitlabform run, you can pass a list of comma-delimited names via the `--exclude-sections` parameter.
 
 ```shell
-gitlabform --exclude_sections group_settings,project_settings
+gitlabform --exclude-sections group-settings,project-settings
 ```
 
 !!! warning
--- a/docs/running.md
+++ b/docs/running.md
@@ -56,4 +56,6 @@ To exclude certain sections of the configuration from a given gitlabform run, yo
 gitlabform --exclude_sections group_settings,project_settings
 ```
 
-**Warning**, the maintainers of gitlabform cannot guarantee consistent functionality when excluding and including different sections in executions of the tool, as Gitlab itself may require a specific set of operations. For example, provisioning of User and Group Permissions often needs to occur prior to other operations, we maintain a list within the `ProjectProcessors` and `GroupProcessors` classes in valid order of operations.
+!!! warning
+
+    gitlabform cannot guarantee consistent functionality when excluding and including different sections in executions of the tool, as Gitlab itself may require a specific set of operations. For example, provisioning of User and Group Permissions often needs to occur prior to other operations, we maintain a list within the `ProjectProcessors` and `GroupProcessors` classes in valid order of operations.
--- a/docs/running.md
+++ b/docs/running.md
@@ -43,3 +43,17 @@ gitlabform -h
 ```
 
 ...to see the current set of supported command line parameters.
+
+## Command Line Options
+
+To view the full set of Command Line Options you can run `gitlabform --help`
+
+### Exclude Sections
+
+To exclude certain sections of the configuration from a given gitlabform run, you can pass a list of comma-delimited names via the `--exclude_sections` parameter.
+
+```shell
+gitlabform --exclude_sections group_settings,project_settings
+```
+
+**Warning**, the maintainers of gitlabform cannot guarantee consistent functionality when excluding and including different sections in executions of the tool, as Gitlab itself may require a specific set of operations. For example, provisioning of User and Group Permissions often needs to occur prior to other operations, we maintain a list within the `ProjectProcessors` and `GroupProcessors` classes in valid order of operations.
--- a/gitlabform/__init__.py
+++ b/gitlabform/__init__.py
@@ -75,6 +75,7 @@ class GitLabForm:
             self.just_show_version = False
             self.terminate_after_error = True
             self.only_sections = "all"
+            self.exclude_sections = []
             self.recurse_subgroups = recurse_subgroups
 
             self._configure_output(tests=True)
@@ -97,6 +98,7 @@ class GitLabForm:
                 self.just_show_version,
                 self.terminate_after_error,
                 self.only_sections,
+                self.exclude_sections,
                 self.recurse_subgroups,
             ) = self._parse_args()
 
@@ -276,6 +278,15 @@ class GitLabForm:
         )
 
         parser.add_argument(
+            "-es",
+            "--exclude-sections",
+            dest="exclude_sections",
+            default="none",
+            type=str,
+            help="exclude sections with these names (comma-delimited)",
+        )
+
+        parser.add_argument(
             "-r",
             "--recurse-subgroups",
             dest="recurse_subsgroups",
@@ -287,6 +298,11 @@ class GitLabForm:
 
         if args.only_sections != "all":
             args.only_sections = args.only_sections.split(",")
+
+        if args.exclude_sections != "none":
+            args.exclude_sections = args.exclude_sections.split(",")
+        else:
+            args.exclude_sections = []
 
         return (
             args.target,
@@ -304,6 +320,7 @@ class GitLabForm:
             args.just_show_version,
             args.terminate_after_error,
             args.only_sections,
+            args.exclude_sections,
             args.recurse_subsgroups,
         )
 
@@ -402,6 +419,7 @@ class GitLabForm:
                 diff_only_changed=self.diff_only_changed,
                 effective_configuration=effective_configuration,
                 only_sections=self.only_sections,
+                exclude_sections=self.exclude_sections,
             )
 
         for group in groups:
@@ -437,6 +455,7 @@ class GitLabForm:
                     diff_only_changed=self.diff_only_changed,
                     effective_configuration=effective_configuration,
                     only_sections=self.only_sections,
+                    exclude_sections=self.exclude_sections,
                 )
 
                 successful_groups += 1
@@ -495,6 +514,7 @@ class GitLabForm:
                     diff_only_changed=self.diff_only_changed,
                     effective_configuration=effective_configuration,
                     only_sections=self.only_sections,
+                    exclude_sections=self.exclude_sections,
                 )
 
                 successful_projects += 1
--- a/gitlabform/__init__.py
+++ b/gitlabform/__init__.py
@@ -283,7 +283,7 @@ class GitLabForm:
             dest="exclude_sections",
             default="none",
             type=str,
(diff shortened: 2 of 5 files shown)


amimas:
@TimKnight-DWP  - is this ready? Will you be adding some test for this new cli option?

TimKnight-DWP:
@amimas more than happy to 👍

TimKnight-DWP:
@amimas added some unit tests and an acceptance test for the parsing of cli args

gitlabform/processors/__init__.py
```diff
@@ -25,16 +25,20 @@ def process_entity(
         dry_run: bool,
         diff_only_changed: bool,
         effective_configuration: EffectiveConfigurationFile,
-        only_sections: List[str],
+        only_sections: List[str] | str,
+        exclude_sections: List[str],
```
amimas, line 29: What's the reason for changing the type for `only_section`? And shouldn't `exclude_section` type be same as `only_section`?

tests/acceptance/standard/test_running.py
amimas, line 1: Do we need changes to this file? Looks like couple of `import` added but no additional code change.

tests/unit/processors/test_abstract_processors.py
amimas, line 1: I think there's already a test file at this location called `test_abstract_processor.py`, right? Can we add the new tests there? If these should be organized in a separate file, maybe let's rename this to make it more intuitive.

gitlabform/__init__.py
```diff
@@ -275,6 +277,15 @@ def _parse_args() -> Tuple:
             help="process only section with these names (comma-delimited)",
         )
 
+        parser.add_argument(
+            "-es",
+            "--exclude-sections",
+            dest="exclude_sections",
+            default="none",
+            type=str,
+            help="exclude sections with these names (comma-delimited). Warning odd functionality may be experienced by enforcing different orders of operation with -es compared to with what Gitlab expects",
```
amimas:
```suggestion
            help="exclude sections with these names (comma-delimited). Warning: odd functionality may be experienced by enforcing different orders of operation with -es compared to with what Gitlab expects",
```

gitlabform/processors/__init__.py
TimKnight-DWP: @amimas `only_sections` expects to be passed either `all` or a list of strings, if no CLI arg is passed, it sets it to `all` -> not a list of strings with just `all` in it, simply the string. I didn't want to change the functionality but I did want to type it correctly

With exclude_section, either you have an empty list and nothing to exclude, or a list of somethings to exclude, I didn't want to add a `none` string variant to match how only_sections work as it feels off

tests/acceptance/standard/test_running.py
TimKnight-DWP: @amimas no, probably a hangover from me trying something but not rolling back fully

tests/unit/processors/test_abstract_processors.py
TimKnight-DWP: I'll remind myself, I did look at it, but if I recall `abstract_processor` class is distinct from `abstract_processors` file so it does need a separate test

TimKnight-DWP: Ah yes I remember now @amimas abstract_processor is it's own class/file, the test I added validates logic in the __init.py__ of `processors/` which is class `AbstractProcessors` so naming the test file wasn't easy

gitlabform/__init__.py
```diff
@@ -275,6 +277,15 @@ def _parse_args() -> Tuple:
             help="process only section with these names (comma-delimited)",
         )
 
+        parser.add_argument(
```
adam-moss: Please make this mutually exclude with only_sections, lest we start seeing things like `-os=project_settings -es=project_settings`

If exclude is in use then include should be `all`

TimKnight-Opencast: @adam-moss updated

adam-moss (approved):

docs/running.md
```diff
@@ -43,3 +43,17 @@ gitlabform -h
...
+
+### Exclude Sections
+
+To exclude certain sections of the configuration from a given gitlabform run, you can pass a list of comma-delimited names via the `--exclude_sections` parameter.
+
+```shell
+gitlabform --exclude_sections group_settings,project_settings
+```
+
+**Warning**, the maintainers of gitlabform cannot guarantee consistent functionality when excluding and including different sections in executions of the tool, as Gitlab itself may require a specific set of operations. For example, provisioning of User and Group Permissions often needs to occur prior to other operations, we maintain a list within the `ProjectProcessors` and `GroupProcessors` classes in valid order of operations.
```
amimas: Let's use the warning callout so that it's rendered accordingly in a box. See `members.md` as a reference.

```suggestion
!!! warning

    gitlabform cannot guarantee consistent functionality when excluding and including different sections in executions of the tool, as Gitlab itself may require a specific set of operations. For example, provisioning of User and Group Permissions often needs to occur prior to other operations, we maintain a list within the `ProjectProcessors` and `GroupProcessors` classes in valid order of operations.
```

> we maintain a list within the `ProjectProcessors` and `GroupProcessors` classes in valid order of operations.

I'm not sure if that's true. I don't think we have such logic/list, unless I misunderstood what you meant. Maybe remove this or clarify for me?

amimas: Also, the latest commit added `add_mutually_exclusive_group` config for the argument parser. I'm new to it, but I think this means following command is invalid or gitlabform will not run:

```
gitlabform --only-sections group_settings --exclude-sections variables
```

Is this right? If yes, then the content above is no longer a "warning". We're forcing users to not be able to use them together. Maybe use `important` instead of `warning` in the docs callout and rephrase the content to explain why this is enforced instead of suggesting that this shouldn't be done.

gitlabform/__init__.py
```diff
@@ -275,6 +279,15 @@ def _parse_args() -> Tuple:
             help="process only section with these names (comma-delimited)",
         )
 
+        sections_group.add_argument(
+            "-es",
+            "--exclude-sections",
+            dest="exclude_sections",
+            default="none",
+            type=str,
+            help="exclude sections with these names (comma-delimited). Warning: may result in odd functionality when enforcing different oders of operation.",
```
amimas: Is the warning still valid? Also, still unsure if this is captured in the document.

gitlabform/processors/__init__.py
amimas: That makes sense. Thanks for explaining.

docs/running.md
TimKnight-DWP: You are correct @amimas you can't use both, but in the ProjectProcessors/GroupProcessors we have an ordered array of individual Processors which we run in a specific order to avoid permissions issues etc (i.e. you generally must update members first).

So I think this warning is still valid?

gitlabform/__init__.py
TimKnight-DWP: See above, perhaps the warning in CLI should point to README for full details 🤔

docs/running.md
amimas: Hi @TimKnight-DWP . Are you referring to this from `ProjectProcessors`?

https://github.com/gitlabform/gitlabform/blob/c3dc6f80ce6cd1cfef8da3f82105913bde9b8fe1/gitlabform/processors/project/__init__.py#L48-L62

If yes, I understand what you're saying. If you exclude `members` section but `branches` section contains permission related config such as `allowed_to_merge` for certain user, that config might not be processed if the user is not already a member of the project.

What do you think of adding examples like that in the doc? That way it's clearer, I think. What threw me off was the reference to "excluding and including" reference and I kept thinking of using `--only-sections` and `--exclude-sections` parameters together.

Either way, let's update the format using `!!!` so that in the website it's rendered accordingly.

gitlabform/__init__.py
amimas: I think reference to readme for full details is a good idea but that can be applicable for all the parameters (once we actually finish documenting them all). In that case, that reference will probably be at the beginning or end of the help output.

What do you think of rephrasing it like this?

```suggestion
            help="exclude sections with these names (comma-delimited). Warning: may result in failure or odd functionality when excluding sections that are dependent on by other sections.",
```

docs/running.md
TimKnight-DWP: @amimas yeah I was really struggling to figure out how to write this in a way for someone who didn't look at the code 😅

amimas (approved):
Thanks for looking into it and adding the unit tests. I hope we can finish documenting rest of the CLI parameters.

docs/running.md
```diff
@@ -43,3 +43,19 @@ gitlabform -h
...
+
+To view the full set of Command Line Options you can run `gitlabform --help`
+
+### Exclude Sections
+
+To exclude certain sections of the configuration from a given gitlabform run, you can pass a list of comma-delimited names via the `--exclude_sections` parameter.
+
+```shell
+gitlabform --exclude_sections group_settings,project_settings
+```
```
amimas: @TimKnight-DWP - Can you please confirm if the example is correct? The parameter name is `--exclude-sections`, right? Not `--exclude_sections`?

TimKnight-DWP: ah yes cli is kebab-case but python is snake_case, so I probably just typed what I'd been writing into the python

docs/running.md
```diff
@@ -43,3 +43,19 @@ gitlabform -h
...
+## Command Line Options
+
+To view the full set of Command Line Options you can run `gitlabform --help`
+
+### Exclude Sections
+
+To exclude certain sections of the configuration from a given gitlabform run, you can pass a list of comma-delimited names via the `--exclude_sections` parameter.
+
+```shell
+gitlabform --exclude_sections group_settings,project_settings
```
TimKnight-DWP:
```suggestion
gitlabform --exclude-sections group-settings,project-settings
```

docs/running.md
```diff
@@ -43,3 +43,19 @@ gitlabform -h
...
 
 ...to see the current set of supported command line parameters.
+
+## Command Line Options
+
+To view the full set of Command Line Options you can run `gitlabform --help`
+
+### Exclude Sections
+
+To exclude certain sections of the configuration from a given gitlabform run, you can pass a list of comma-delimited names via the `--exclude_sections` parameter.
```
TimKnight-DWP:
```suggestion
To exclude certain sections of the configuration from a given gitlabform run, you can pass a list of comma-delimited names via the `--exclude-sections` parameter.
```

Issue #791, opened 2024-06-24
GitLab now supports username natively when adding member to project


tmeijn:
Hey, just to make you aware, with 17.1 GitLab now supports username natively when adding member to project. I think in the current implementation we fetch the member_id based on the username?

If so, seems like this could reduce some API calls.

References:

* [Announcement in blogpost](https://about.gitlab.com/releases/2024/06/20/gitlab-17-1-released/?utm_source=pocket_shared#add-members-by-username-with-the-members-api)
* [Implementation MR](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/147080)
* [API doc](https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project)

amimas:
Thanks for the heads up. Would you be able to help with the update? PR is welcome if anyone else can pick this up.

jimisola:
> Thanks for the heads up. Would you be able to help with the update? PR is welcome if anyone else can pick this up.

How do we handle features like this that are only available from a particular GitLab version?

amimas:
> How do we handle features like this that are only available from a particular GitLab version?

I'd say in a backward-compatible way. So, there should be logic to check the version of gitlab and then apply the change accordingly. In the next major version release, the extra logic can be removed. Or, if it's not done in backward-compatible way as mentioned, a new major version will need to be released.

PR #1017, opened 2025-05-06
perf: use the new native parameter to add new members instead of using the user id

BenediktBertsch:
- resolves https://github.com/gitlabform/gitlabform/issues/791

I'd like to discuss the new `_process_gitlab_version_for_native_members_call` - I think it would be better if we define the function on an upper level or statically somewhere so every future need can be covered with it.

E.g. the function call could look something like that: `_check_version_compatibility(current_version: str, needed_version: str) -> bool`. With these parameters we can check generically if the current version is compatible. Sadly I dont have enough experience with this project to know what kind of place holds more generic functions that are shared across multiple files.

### DIFF
--- a/gitlabform/processors/project/members_processor.py
+++ b/gitlabform/processors/project/members_processor.py
@@ -79,6 +79,23 @@ class MembersProcessor(AbstractProcessor):
         else:
             verbose("Not enforcing group members.")
 
+    def _process_gitlab_version_for_native_members_call(self, version: str) -> bool:
+        # this call is only needed for the newly introduced (17.1) member GET call
+        # instead of needing an additional call to get the member id we can use this now
+        # not completely happy about this function, would prefer a more generic approach
+        # so other internal processes can use this kind of approach as well for other API
+        splitted_version = version.split(".")
+        major_version = int(splitted_version[0])
+        minor_version = int(splitted_version[1])
+
+        if major_version > 17:
+            return True
+
+        if major_version >= 17 and minor_version >= 1:
+            return True
+
+        return False
+
     def _process_users(
         self,
         project_and_group: str,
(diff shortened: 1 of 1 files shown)


BenediktBertsch:
Not quite sure how to write tests that are covering both cases... should I create a matrix for it and run two versions of gitlab then via GitHub actions (considering the move, does it make sense to invest here)?

TimKnight-DWP:
I think just unit tests work here, I think only self-managed Gitlab could be behind 17 now, their update cadence means gitlab.com is on 18, and "dedicated" instances will be moving to 18 in the next month. So not sure how many people would be on 16 @BenediktBertsch

BenediktBertsch:
@TimKnight-DWP have you read the worry I had about the placement of the function that checks for the gitlab version? Any thoughts?

amimas:
@BenediktBertsch - Thanks for helping with gitlabform development.

I haven't had time to review this PR. But wanted to drop this note here for attention. I had a chance to review another PR recently that is trying to do something similar - different behavior based on gitlab version. Please see my comment here:

https://github.com/gitlabform/gitlabform/pull/996#discussion_r2051733171

I'd like us to not come up with multiple approaches. Seems like version checking is becoming more relevant than before. So, ideally we implement it so that it's not catered for individual features of gitlabform but as a whole. Hope that makes sense.

amimas:
> Not quite sure how to write tests that are covering both cases... should I create a matrix for it and run two versions of gitlab then via GitHub actions (considering the move, does it make sense to invest here)? Or should I create unit only tests here?

That makes sense to me. I think specific version of gitlab should only run tests that are relevant (e.g. not the entire suite). We run entire suite of tests using the latest gitlab version.

My reasoning is that version specific behavior is applied at the processors which are always tested using live gitlab instance (acceptance tests).

@TimKnight-DWP - thoughts?

TimKnight-DWP:
@BenediktBertsch I think where you phyiscally put the code and when you call it makes the most sense, I can't think of any neater ways of doing so

TimKnight-DWP:
Unless we set the version as a global variable on initialisation - in case we need them in other places in future

TimKnight-DWP (approved):

TimKnight01:
@amimas agreed let's focus on getting the centralisaton piece done

PR #1018, opened 2025-05-06
Fix: verbose output for members_processor

BenediktBertsch:
I've seen while I worked on another PR that the verbose output doesn't work as intended e.g.:

> Processing user 'testuser'...
> http://localhost:80 "GET /api/v4/users?username=testuser HTTP/1.1" 200 None
> http://localhost:80 "GET /api/v4/users/35 HTTP/1.1" 200 None
> Nothing to change for user '%s' - same config now as to set. testuser

It looks like the verbose does not handle the %s accordingly, thats why I changed it to format strings instead and now it looks like intended:

> Processing user 'testuser'...
> http://localhost:80 "GET /api/v4/users?username=testuser HTTP/1.1" 200 None
> http://localhost:80 "GET /api/v4/users/35 HTTP/1.1" 200 None
> Nothing to change for user 'testuser' - same config now as to set.

If wanted I could look through the whole project and replace them, but not quite sure if thats wanted.

### DIFF
--- a/gitlabform/processors/project/members_processor.py
+++ b/gitlabform/processors/project/members_processor.py
@@ -49,15 +49,10 @@ class MembersProcessor(AbstractProcessor):
                     and expires_at == current_groups[common_group_name]["expires_at"]
                     and access_level == current_groups[common_group_name]["group_access_level"]
                 ):
-                    verbose(
-                        "Ignoring group '%s' as it is already a member",
-                        common_group_name,
-                    )
-                    verbose(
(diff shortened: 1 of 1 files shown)


TimKnight-DWP:
I've seen these as well, thanks for looking into it

TimKnight-DWP:
@BenediktBertsch I'll approve these, if you felt like doing the whole project, I'd happily review that MR

TimKnight-DWP (approved):

Issue #1025, opened 2025-05-14
Rate limit and max_retries configuration


philippe-granet:
`python-gitlab` obeys the rate limit of the GitLab server by default.
On receiving a 429 response (Too Many Requests), python-gitlab sleeps for the amount of time in the Retry-After header that GitLab sends back.
If GitLab does not return a response with the Retry-After header, python-gitlab will perform an exponential backoff.

https://github.com/python-gitlab/python-gitlab/blob/main/docs/api-usage-advanced.rst#rate-limits

The problem is that the default 10 retries are not sufficient:
![Image](https://github.com/user-attachments/assets/bddcfa09-c7c6-45fe-a3fa-f1d13ea31dfa)

it would be nice if Gitlabform can be parametrized to override the `max_retries` parameter of `python-gitlab` API, to be able to increase the number of retries.

amimas:
Thanks for the suggestion @philippe-granet . In addition to `max_retries`, maybe `obey_rate_limit` should be configurable?

> If you don't want to wait, you can disable the rate-limiting feature, by supplying the obey_rate_limit argument.
> 
> ```py
> import gitlab
> import requests
> 
> gl = gitlab.Gitlab(url, token, api_version=4)
> gl.projects.list(get_all=True, obey_rate_limit=False)
> ```
> If you do not disable the rate-limiting feature, you can supply a custom value for max_retries; by default, this is set to 10. To retry without bound when throttled, you can set this parameter to -1. This parameter is ignored if obey_rate_limit is set to False.
> 
> ```py
> import gitlab
> import requests
> 
> gl = gitlab.Gitlab(url, token, api_version=4)
> gl.projects.list(get_all=True, max_retries=12)
> ```

Also, maybe `retry_transient_errors` should be configurable too? Right now it's hard coded to be enabled.

Anyways, I think it should be an option similar to `timeout` config. What do you think? I think it's relatively simple change. Can you help add this feature? Let us know if you need any help or have questions.

PR #1035, opened 2025-05-31
test: fix test case for job token scope

### LINKED ISSUE #1029: ci: gitlabform pipeline broken after gitlab version 18 release
The project's CI seems to be broken as multiple PRs have been failing to pass, preventing changes from being merged. For example: following dependency updates should not be failing. This started happening since the CI picked up latest version of gitlab, which happens to be v18 released by GitLab.

- #1026 
- #1027 
- #1028 

Snippet of the failure:

```
=========================== short test summary info ============================
FAILED tests/acceptance/standard/test_job_token_scope.py::TestProjectJobTokenScope::test__disable_limit_access_to_this_project - assert not True
 +  where True = <ProjectJobTokenScope>.inbound_enabled
======== 1 failed, 177 passed, 2 skipped, 3 rerun in 588.63s (0:09:48) =========
```

There have been breaking change introduced in GitLab v18 related to CI Job Token Scope. This might require refactor in GitLabForm's logic/test cases to reflect the change accordingly.

https://docs.gitlab.com/update/deprecations/?removal_milestone=18.0&breaking_only=true#cicd-job-token---authorized-groups-and-projects-allowlist-enforcement

https://docs.gitlab.com/update/deprecations/?removal_milestone=18.0&breaking_only=true#cicd-job-token---limit-access-from-your-project-setting-removal

Unfortunately cannot merge any new changes to this project until the CI is fixed.

amimas:
In a recent major release of GitLab (v18), the CI Job token scope
related setting introduced breaking changes. Going forward, the
option to configure 'limit access to this project' has been removed
and it's enabled by default. However, this option can be restored
by GitLab admin from an instance level setting.

See the [breaking change announcement](https://docs.gitlab.com/update/deprecations/?removal_milestone=18.0&breaking_only=true#cicd-job-token---authorized-groups-and-projects-allowlist-enforcement)
for more details.

After GitLab v18, the test case for gitlabform validating if the
setting mentioned above can be 'disabled', began to fail. This is
because the setting could no longer be configured. This change adds
necessary logic in the test case so that instance level setting is
updated to allow this test case to validate that the option can be
disabled.

In addition, job token scope related documentation has been updated for clarification.

fixes #1029

### DIFF
--- a/tests/acceptance/standard/test_job_token_scope.py
+++ b/tests/acceptance/standard/test_job_token_scope.py
@@ -35,16 +35,30 @@ class TestProjectJobTokenScope:
         project: Project,
     ):
         """
-        Test that project-level job token scope can be disabled when the instance setting allows it.
-        
-        This test requires:
-        1. GitLab instance setting 'enforce_ci_inbound_job_token_scope_enabled' to be configurable
-        2. Admin access to modify instance settings
+        Test that project-level 'Limit access to this project' setting can be configured
+        when the instance-level setting 'enforce_ci_inbound_job_token_scope_enabled' is disabled.
+
+        The test verifies that:
+        1. When instance setting is disabled, projects can control their own 'inbound_enabled' setting
+        2. The gitlabform config 'limit_access_to_this_project: false' correctly sets
+           the project's 'inbound_enabled' to false via the GitLab API
+
+        This test requires admin access to modify the instance setting.
         """
-        # Setup: Ensure instance setting allows project-level configuration
+        # Setup: Disable instance-level enforcement to allow project-level configuration
         instance_settings = gl.settings.get()
         instance_settings.enforce_ci_inbound_job_token_scope_enabled = False
-        instance_settings.save()
+        
+        # Retry settings update in case of temporary resource issues
+        max_retries = 3
+        for attempt in range(max_retries):
+            try:
+                instance_settings.save()
+                break
+            except Exception as e:
+                if attempt == max_retries - 1:
+                    raise Exception(f"Failed to update instance settings after {max_retries} attempts: {str(e)}")
+                print(f"Attempt {attempt + 1} failed, retrying...")
 
         try:
             # Setup: Enable limit access at project level
@@ -65,9 +79,12 @@ class TestProjectJobTokenScope:
 
         finally:
             # Cleanup: Restore instance setting to its original state
-            instance_settings = gl.settings.get()
-            instance_settings.enforce_ci_inbound_job_token_scope_enabled = True
-            instance_settings.save()
+            try:
+                instance_settings = gl.settings.get()
+                instance_settings.enforce_ci_inbound_job_token_scope_enabled = True
+                instance_settings.save()
+            except Exception as e:
+                print(f"Warning: Failed to restore instance settings: {str(e)}")
 
     def test__add_other_project_to_job_token_scope_by_name(
         self,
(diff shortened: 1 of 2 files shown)


tests/acceptance/standard/test_job_token_scope.py
```diff
@@ -31,25 +31,60 @@ def test__enable_limit_access_to_this_project(
...
+        # Retry settings update in case of temporary resource issues
+        max_retries = 3
+        for attempt in range(max_retries):
+            try:
+                instance_settings.save()
+                break
+            except Exception as e:
+                if attempt == max_retries - 1:
+                    raise Exception(f"Failed to update instance settings after {max_retries} attempts: {str(e)}")
+                print(f"Attempt {attempt + 1} failed, retrying...")
```
gdubicki: Do we really need this retry loop even though [we already use `retry_transient_errors`](https://github.com/gitlabform/gitlabform/issues/1025#issuecomment-2894233400)?

gdubicki (commented):
Is this code backwards-compatible with GitLab < 18?

tests/acceptance/standard/test_job_token_scope.py
amimas: Good question. Not sure if we really need this or not.

When I was running that test locally (not the entire suite), I'd occasionally get '500 Internal server error' when attempting to update gitlab application/instance setting.

This was probably because gitlab running in docker, but not sure. Since we do the same in the CI process, thought of adding the retry so that test is not randomly failing making the pipeline unstable.

Let me know if you have suggestions.

amimas:
> But is this code backwards-compatible with GitLab < 18?

Yes, I think so. The failing test was attempting to verify if job token scope can be disabled. The code is only around acceptance test.

As of gitlab v18, it's always enabled and not configurable by project settings. But it can be made configurable (allow to disable) if allowed by instance level setting.

So, the feature itself is same, but has more stricter control/default behavior. Hope that makes sense.

See my comment here for more details: https://github.com/gitlabform/gitlabform/issues/1029#issuecomment-2926250946

tests/acceptance/standard/test_job_token_scope.py
gdubicki: `retry_transient_errors` should help with 500s, quoting [the docs](https://python-gitlab.readthedocs.io/en/stable/api-usage-advanced.html#transient-errors):

> GitLab server can sometimes return a transient HTTP error. python-gitlab can automatically retry in such case, when retry_transient_errors argument is set to True. When enabled, HTTP error codes 500 (Internal Server Error), 502 (502 Bad Gateway), 503 (Service Unavailable), 504 (Gateway Timeout), and Cloudflare errors (520-530) are retried.

amimas: Yup. I agree with you. But, I'm just not sure if that is sufficient for gitlabform tests which uses gitlab in a container with very limited resources compared to how gitlab would be running in a real environment.

I'm happy to remove the retry logic and we can re-evaluate it if it doesn't work consistently in gitlabform ci, running gitlab in container.

amimas: @gdubicki - Removed the extra retry logic from the test in latest commit.

TimKnight-DWP (approved):

tests/acceptance/standard/test_job_token_scope.py
amimas: @gdubicki - Going to resolve this thread and merge this PR so that other PRs are not blocked because of broken CI pipeline. If you find something wasn't addressed here, please open a new issue.

Thanks for your review.

PR #1040, opened 2025-06-03
fix: Handle the case where an integration doesn't exist when trying to delete it. (fixes #1036)

### LINKED ISSUE #1036: Deleting a non-existant integration results in a 404 error.
If you delete an integration using config like...

```
  integrations:
    jira:
      delete: true
```

... but the integration was never setup on the project, then you get the following error when processing it...

```
Processing section 'integrations'
Deleting integration: jira
Warning: Error occurred while processing project groupname/projectname, exception:

404: 404 Integration Not Found
```

**GitLabForm version**

`4.2.5`

**GitLab version**

`17.11.2 (self managed, premium tier)`

andrewjw:


TimKnight-DWP (approved):

gitlabform/processors/project/integrations_processor.py
```diff
@@ -18,7 +19,15 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
 
             if configured_integrations[integration].get("delete"):
                 verbose(f"Deleting integration: {integration}")
-                gl_integration.delete()
+                try:
```
amimas, line 22: Looks like we're retrieving a list of all integrations before we attempt to delete the integration. That means we should be able to find out if the integration exists or not, right? So, we shouldn't need to capture exceptions - simply attempt to delete if it exists. This way we can save unnecessary API calls.

In addition, the line (18) where we're retrieving project integrations - wondering if that needs to happen inside the for loop. We should just get the list of integration once.

What do you think?

tests/acceptance/standard/test_integrations.py
```diff
@@ -68,6 +68,39 @@ def test__if_delete_works(self, project):
...
+    def test__if_delete_without_create_works(self, other_project):
+        for integration_name in ["jira", "slack"]:
+            try:
+                other_project.integrations.get(integration_name)
+            except GitlabGetError as e:
+                assert (
+                    e.response_code == 404
+                ), f"Integration {integration_name} should not exist, but got a different error: {e}"
+            else:
+                assert False, f"Integration {integration_name} should not exist."
```
amimas: I believe the `other_project` fixture's scope is class level. There's another fixture named `project_for_function`. If we use that fixture instead, we shouldn't need this extra setup here trying to validate the integrations don't exist before we start the tests - the project will be created fresh for this function without any pre-configured settings.

amimas (changes requested):
Thank you so much for the PR. Have some suggestions from my observation. Appreciate if you could please take a look. Thanks.

amimas:
> As far as I can tell the CI failures are not due to this change. Happy to correct my patch if you can point me in the right direction.

A fix has been merged. Please update your branch to be up-to-date with `main`. Thanks.

gitlabform/processors/project/integrations_processor.py
andrewjw: I'm a bit confused - line 18 is inside the for loop. We get all of the integration names from the config, then create an object representing it on the gitlab API - this is lazy so no API call occurs on line 18. It's only when we delete that we discover it doesn't actually exist.

We don't currently support config diffing, so the config for the integration is updated every run. If we changed line 18 to not be lazy (and therefore discovered it didn't exist at that point), in the more common case of setting the configuration for the integration this would become two calls rather than one.

If there's a bigger change to support config diffing and the common case becomes no update is needed then I agree we should change 18 to be eager, because that will keep the normal flow to be one request.

tests/acceptance/standard/test_integrations.py
andrewjw: I've made this fix.

tests/acceptance/standard/test_integrations.py
```diff
@@ -68,6 +68,29 @@ def test__if_delete_works(self, project):
...
+
+        for integration_name in ["jira", "slack"]:
+            try:
+                project_for_function.integrations.get(integration_name)
+            except GitlabGetError as e:
+                assert (
+                    e.response_code == 404
+                ), f"Integration {integration_name} should not exist, but got a different error: {e}"
+            else:
+                assert False, f"Integration {integration_name} should not exist."
```
amimas: Instead of catching `GitLabGetError` exception, couldn't we assert that 0 integrations exist for the project? Basically something like:

```py
project_integrations = project.integrations.list(get_all=True)
assert len(project_integrations) == 0
```

Please let me know if you think this misses some validation aspect.

gitlabform/processors/project/integrations_processor.py
amimas: Ahh... I missed some details. Thanks for clarifying it for me.

Also, didn't realize that the processor currently doesn't support config diffing. I think config diffing should be supported. That should probably be a separate PR though. That's when this logic can be changed. An example might be something like this:

https://github.com/gitlabform/gitlabform/blob/5d85eda32e8f9427ab460543995c7fe729eb4b69/gitlabform/processors/project/schedules_processor.py#L27-L34

Anyways, let's leave it out of this PR for now. What do you think?

tests/acceptance/standard/test_integrations.py
andrewjw: Good point, I've pushed this change.

amimas (approved):
Looks good.

gitlabform/processors/project/integrations_processor.py
amimas: Looks like my last reply somehow disappeared.

Thanks for clarifying it for me. I didn't look at the code closely. But I agree with you.

We should support config differing. That'll probably involve changing the logic. Schedule processor is a good example to follow. But let's leave that out of this PR. Will you be able to help out with that in a separate PR?

PR #1065, opened 2025-07-23
docs: update intro using common term

mfriedenhagen:
* configuration-as-code is a common term and the indefinite article in front of code had me read the line three times.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -9,7 +9,7 @@
   ![GitLabForm logo](images/gitlabform-logo.png)
 </figure>
 
-is a specialized configuration as a code tool for GitLab:
+GitLabForm is a specialized configuration-as-code tool for GitLab:
 
 * application settings,
 * groups,


docs/index.md
```diff
@@ -9,7 +9,7 @@
   ![GitLabForm logo](images/gitlabform-logo.png)
 </figure>
 
-is a specialized configuration as a code tool for GitLab:
+is a specialized configuration-as-code tool for GitLab:
```
amimas: Should we add name of the tool at the beginning? I'm thinking that probably makes it more complete sentence and also helps with accessibility, SEO, etc.

Right now you have to read the image and then the text, which seems strange.

Thoughts?

amimas (commented):
Thank you so much for the PR and helping to improve the docs.

amimas (approved):
Looks good.

Issue #1061, opened 2025-07-16
Unprotecting and re-protecting a branch resets manual approval rules to "All branches"


rickbrouwer:
**Describe the bug**

We require that the `develop` branch always is protected..
To ensure this, we use GitLabform to configure branch protection like this:

```
branches:
  develop:
    allow_force_push: false
    merge_access_level: maintainer
    protected: true
    push_access_level: no access
```

Teams are free to define their own manual approval rules.
For example, a team may add a rule called `security approval` specifically for the `develop` branch, managed outside of GitLabform.

However, we’ve noticed that GitLabform currently unprotects and then re-protects a branch during processing:
https://github.com/gitlabform/gitlabform/blob/f3893de5a9d668542421a5d60ec8b93b914d2883/gitlabform/processors/project/branches_processor.py#L57-L63

This unprotect/protect cycle causes manually added approval rules to be altered.
The rules themselves are not deleted, but their scope changes: instead of being specific to `develop`, they are reset to apply to `All branches`.

This behavior breaks team-defined compliance rules and causes unexpected changes to approval requirements.

**GitLabForm version**

GitLabForm version: 4.2.6

**GitLab version**

GitLab Enterprise Edition [v18.0.4-ee

TimKnight-DWP:
Should be resolved by: https://github.com/gitlabform/gitlabform/pull/1070

rickbrouwer:
Great! Thanks!!!

Issue #1034, opened 2025-05-30
Using the premium only branch protection rules causes unnecessary updating


andrewjw:
If you use the premium only branch protection rules (allowed_to_push, allowed_to_merge) then the diffing doesn't work and it unprotects and then reprotects the branches every run. This causes extra api load, and spams the project's audit logs.

Example config
```
    main:
      protected: true
      allowed_to_push:
        - user: username
      allowed_to_merge:
        - access_level: developer
      code_owner_approval_required: true
```

This gets turned into the following Python object...

```
{'protected': True, 'allowed_to_push': [{'user_id': 1767}], 'allowed_to_merge': [{'access_level': 30}], 'code_owner_approval_required': True}
```

The data returned from the GitLab API is:

```
{'project_id': '20306', 'id': 28843761, 'name': 'master', 'push_access_levels': [{'id': 28960972, 'access_level': 40, 'access_level_description': 'username', 'deploy_key_id': None, 'user_id': 1767, 'group_id': None}], 'merge_access_levels': [{'id': 28981208, 'access_level': 30, 'access_level_description': 'Developers + Maintainers', 'user_id': None, 'group_id': None}], 'allow_force_push': False, 'unprotect_access_levels': [{'id': 26711591, 'access_level': 40, 'access_level_description': 'Maintainers', 'user_id': None, 'group_id': None}], 'code_owner_approval_required': True, 'inherited': False}
```

and since they don't match, the branch is unprotected and reprotected.

**GitLabForm version**

`GitLabForm version: 4.2.5`

**GitLab version**

`17.11.2` (self managed, premium tier)

hiaik:
Hi @amimas  I can see you left 👀  for the initial comment.

Have you got a chance to look at the issue? Based on the description, it causes the same issues for us to recreate (remove and create protected branches again) for ~ 700 repos on each gitlabform run.

Happy to debug more and provide logs if you need!

hiaik:
@amimas @andrewjw It seems https://github.com/gitlabform/gitlabform/pull/996 might solve the issue reported in this ticket based on 

```
In this PR I try to implement diff detection, so that GitLabForm updates branch protection only when needed, and changes are applied only to rules/settings that are different from those set in GitLabForm config.
```

amimas:
I suggest we keep it open until the PR has been merged.

Issue #445, opened 2022-11-24
Update branch protection logic with API changes available in GitLab 15.6


amimas:
In 15.6 of GitLab, following changes were made to the branch protection API

> Previously, the UI was required to update the access levels of protected branches. The API required you to unprotect, then reprotect, a branch when updating its access levels. Now, the [protected branches API](https://docs.gitlab.com/ee/api/protected_branches.html#update-a-protected-branch) enables you to directly update which users or groups are allowed_to_push, allowed_to_merge, allowed_to_unprotect, and more. This one-step method decreases the risk of a bot changing this setting and leaving a branch unprotected.

Ref: https://about.gitlab.com/releases/2022/11/22/gitlab-15-6-released/#update-access-levels-from-protected-branch-api

I think there might be an opportunity for further peformance improvement of gitlabform since it doesn't have to go through unprotect and reprotect operation. It should make the project's audit logs cleaner too. Perhaps even check to see if new operation is needed if config is same as what's already set?

I don't think existing logic can be removed yet though as not everyone maybe up-to-date with GitLab version. For at least for some time, gitlabform may support both existing and new logic depending on which gitlab version is detected. Maybe existing logic can be removed in the next major release of gitlabform.

gdubicki:
Hi @amimas!

Thanks for the report, I didn't notice that change in GitLab's API.

I agree that such a change, done in a backward-compatible manner, doing it the new way for GitLab >=15.6 and the old way for <15.6 would be good and we are open to PRs with this. :)

PR #1070, opened 2025-08-04
fix!: do not unprotect branches when applying config changes and detect if it needs update correctly

TimKnight-DWP:
**Breaking Change**

When preventing GitlabForm from un-protecting and then re-protecting the Branch in order to apply configuration changes to the protection rules [#1070](https://github.com/gitlabform/gitlabform/pull/1070), we discovered some functionality that was previously hidden from the user.

Previously if you stopped defining `code_owner_approval_required` or `allow_force_push` in the YAML config, the settings would revert to the default Gitlab settings applied when first protecting the branch, see: [Branch Protection API](https://docs.gitlab.com/api/protected_branches/#protect-repository-branches)

With the changes to perform dynamic updates as required, we follow the GitlabForm convention of **leaving the Gitlab settings as is if the data is no longer defined in configuration**.

Therefore, if you wish to ensure the Protected Branch settings are applied as you intend when running GitlabForm you should define some defaults explicitly:

```yaml
projects_and_groups:
 my-group/*:
   branches:
     main:
       protected: true
       allow_force_push: false
       code_owner_approval_required: true
```

- Extension of: https://github.com/gitlabform/gitlabform/pull/996

Resolves:
- https://github.com/gitlabform/gitlabform/issues/445
- https://github.com/gitlabform/gitlabform/issues/1061
- https://github.com/gitlabform/gitlabform/issues/1034


TimKnight-DWP:
I **think** this is working code, but the integration tests seems very flaky, if I debug and step through 9/10 times it'll pass. But sometimes there seems to be an issue in the needs_update detection. I was hoping it was just that this machine is slow, but I will need to do so more digging

TimKnight-DWP:
@amimas @br0ziliy @rickbrouwer this should now resolve the various issues reported in the branch protection area of gitlabform.

Thanks for the ground work @br0ziliy you found a way to make the update API work with python-gitlab that had completely stumped me last week ❤️

br0ziliy:
Thanks @TimKnight-DWP for picking this up, and pushing towards completion. 

Please let me know if you’d like me to run some more tests/verifications - I have a GitLab instance with premium subscription available.

tests/acceptance/premium/test_branches.py
amimas, line 1: I noticed some of the tests have been changed to use fixtures that are scoped to the function. For example: `project_for_function` instead of `project`, `branch_for_function` instead of `branch`. Any particular reason for this change.

I like the advantage with the function scoped fixtures - start with a clean slate and I can run individual tests on demand.

But, I observed couple of issues as the project grew along with tests. When I run the tests (i.e. all tests for branches) locally using function scoped fixtures (i.e. `project_for_function`), often times I'd run into an error where the test fails to create a project because the name has been taken already. Using function scoped fixtures also increases the overall duration for the tests since those resources (i.e. project) has to be created first.

That's why lately I've been a bit reserved when it comes to using function scoped fixtures. In your change for this PR, are they necessary? Should/Could we try to stick with class scoped fixtures for some of the tests? What do you think?

tests/acceptance/premium/test_branches.py
```diff
@@ -1,53 +1,30 @@
...
 
 from gitlabform.gitlab import AccessLevel
 from tests.acceptance import get_only_branch_access_levels, run_gitlabform
+from tests.acceptance.conftest import create_project_member_developer
 
 pytestmark = pytest.mark.requires_license
 
 
 class TestBranches:
-    def test__code_owners_approval(self, project, branch):
```
amimas, line 13: Just to confirm, is this test moved to somewhere else? It's not removed completely, right?

tests/acceptance/conftest.py
```diff
@@ -317,15 +317,19 @@ def _make_user(level: AccessLevel = AccessLevel.DEVELOPER, add_to_project: bool
...
 
     yield user
 
     gl.users.delete(user.id)
 
 
+def create_project_member_developer(gl, project_to_add_user_to):
+    username = get_random_name("user")
+    user = _create_user_and_add_to_project(gl, project_to_add_user_to, username)
+    return user
```
amimas: I think this was created because the test cases started using `project_for_functon`, right? But, I can see this being useful even if we use `project` fixture. Thanks!

What do you think of making this function more generic so that we can pass the role? Right now the tests are using "developer" role everywhere, but maybe we can mix it up?

tests/acceptance/premium/test_branches.py
```diff
@@ -400,6 +381,18 @@ def test__modify_protection(self, project_for_function, group_for_function, bran
...
+        protected_branch: ProjectProtectedBranch = project_for_function.protectedbranches.get(branch_for_function)
+
+        approval_rules = project_for_function.approvalrules.list(get_all=True)
+        assert len(approval_rules) == 1
+        approval_rule = approval_rules[0]
+        assert approval_rule.name == "any"
+        assert approval_rule.approvals_required == 2
+        pb_ar = approval_rule.protected_branches[0]
+        assert pb_ar.get("id") == protected_branch.id
+
```
amimas: I think the expectation is that after the previous/latest run of gitlabform, branch protection shouldn't be doing "unprotect" and "reprotect" flow. Is that right?

If yes, could you please help me understand how these assertions are confirming that? I'm not able to follow the code comment - probably because there's something special going on with GitLab's API. Not sure what you mean by "should have a different protected branch id after the GLF run and therefore will not match the protected_branch_ids array on approval rule".

I think you had mentioned it somewhere - the branch protection API is really messy. Appreciate if you can clarify the logic here. Maybe rephrase the comment if needed.

tests/acceptance/premium/test_branches.py
```diff
@@ -400,6 +381,18 @@ def test__modify_protection(self, project_for_function, group_for_function, bran
...
         # Reset to standard rules
```
amimas, line 291: Do you think we need rest of the logic for this modify/update case?

tests/acceptance/standard/test_branches.py
```diff
@@ -60,19 +62,66 @@ def test__protect_and_unprotect(self, project, branch):
...
+            branches:
+              {branch}:
+                protected: true
+        """
+
+        run_gitlabform(config_remove_settings_for_branch, project.path_with_namespace)
+
+        protected_branch = project.protectedbranches.get(branch)
+        assert protected_branch.allow_force_push is False
+        assert protected_branch.code_owner_approval_required is False
```
amimas: Hmm... In this case, I would expect gitlabform to not change any of the settings for branch protection. The config simply removed 2 attributes (`allow_force_push`, `code_owner_approval_required`). Shouldn't these 2 attributes remain whatever their last config was? Why are they being changed?

tests/acceptance/standard/test_branches.py
```diff
@@ -60,19 +62,66 @@ def test__protect_and_unprotect(self, project, branch):
...
         config_protect_branch = f"""
          projects_and_groups:
-           {project.path_with_namespace}:
+           {project_for_function.path_with_namespace}:
              branches:
-               {branch}:
+               {branch_for_function}:
                  protected: true
-                 push_access_level: {AccessLevel.NO_ACCESS.value}
+                 push_access_level: {AccessLevel.NO_ACCESS}
```
amimas: I think this was an accidental change?

tests/acceptance/standard/test_branches.py
```diff
@@ -82,24 +131,87 @@ def test__modify_protection(self, project, branch):
...
 
-        config_unprotect_branch = f"""
+        config_modify_protection_on_branch = f"""
+         projects_and_groups:
+           {project_for_function.path_with_namespace}:
+             branches:
+               {branch_for_function}:
+                 protected: true
+                 push_access_level: {AccessLevel.MAINTAINER.value}
+                 unprotect_access_level: {AccessLevel.MAINTAINER}
```
amimas:
```suggestion
                 unprotect_access_level: {AccessLevel.MAINTAINER.value}
```

tests/acceptance/standard/test_branches.py
```diff
@@ -82,24 +131,87 @@ def test__modify_protection(self, project, branch):
...
+            merge_access_levels,
+            push_access_user_ids,
+            merge_access_user_ids,
+            _,
+            _,
+            unprotect_access_level,
+        ) = get_only_branch_access_levels(project_for_function, branch_for_function)
+        assert push_access_levels == [AccessLevel.MAINTAINER.value]
+        # If it is not defined in create or update, Gitlab defaults to "Maintainer" level
+        assert merge_access_levels == [AccessLevel.MAINTAINER.value]
```
amimas: Not sure if this is the right explanation or that's how the code works. I think here merge access level is "maintainer" because it's not in the config but earlier run of gitlabform included that config, which set it to "maintainer". Maybe we should use a different role for merge access level in earlier config and then test it.

tests/acceptance/standard/test_branches.py
```diff
@@ -82,24 +131,87 @@ def test__modify_protection(self, project, branch):
...
+        assert push_access_user_ids == []
+        assert merge_access_user_ids == []
+        assert unprotect_access_level is AccessLevel.MAINTAINER.value
+
+        config_remodify_protection_on_branch = f"""
```
amimas, line 145: What's the purpose of rest of the logic in this function? Not clear about the scenario being tested. Can we add a little snippet of comment mentioning the scenario where we are running multiple iterations of gitlabform?

tests/unit/processors/test_branches_processor.py
```diff
@@ -0,0 +1,262 @@
...
+        self.processor = BranchesProcessor(self.gitlab, self.strict)
+
+    def test_is_branch_name_wildcard(self):
+        # Test cases with wildcards
+        assert BranchesProcessor.is_branch_name_wildcard("branch*") is True
+        assert BranchesProcessor.is_branch_name_wildcard("*branch") is True
+        assert BranchesProcessor.is_branch_name_wildcard("br*nch") is True
+        assert BranchesProcessor.is_branch_name_wildcard("branch?") is True
+        assert BranchesProcessor.is_branch_name_wildcard("?branch") is True
+        assert BranchesProcessor.is_branch_name_wildcard("br?nch") is True
```
amimas: Is this actually supported by GitLab - usage of `?` as wildcard.

tests/unit/processors/test_branches_processor.py
```diff
@@ -0,0 +1,262 @@
...
+
+        # Test cases without wildcards
+        assert BranchesProcessor.is_branch_name_wildcard("branch") is False
+        assert BranchesProcessor.is_branch_name_wildcard("main") is False
+        assert BranchesProcessor.is_branch_name_wildcard("feature-123") is False
+
+    def test_convert_user_and_group_names_to_ids(self):
```
amimas, line 29: Aren't these already covered by acceptance tests? Any benefit to adding these unit tests using mock?

tests/acceptance/premium/test_branches.py
TimKnight-DWP: @amimas I was having some issues with not having a clean slate - but I'll go back through and see if those have resolved with other fixes

tests/acceptance/premium/test_branches.py
TimKnight-DWP: @amimas yes there's a standard test covering this now

tests/acceptance/conftest.py
TimKnight-DWP: yeah happy to refactor and pass a default access_level through

tests/acceptance/premium/test_branches.py
TimKnight-DWP: @amimas I wanted to really stress going between "Premium" and "Standard" feature definition in the yaml, as it feels like one of those edge-cases where Gitlab could be weird and I wanted to ensure it was covered. I remember the last part failing when the first parts passed until I made some logic changes.

tests/acceptance/standard/test_branches.py
TimKnight-DWP: @amimas I felt like if you didn't define them we should set to whatever the default value is when Gitlab first protects a branch if you don't supply them (which is both _False_) -> happy to leave at whatever the last value was though, it's an easy logic change

tests/acceptance/standard/test_branches.py
TimKnight-DWP: No I deliberately wanted to validate what happens if the YAML passes the words "No Access" or "Maintainer" through (which we support) rather than the integer representation, just in case there was missing logic or magic behind the scenes GL stuff

tests/acceptance/standard/test_branches.py
TimKnight-DWP: I can clarify the comment, see https://docs.gitlab.com/api/protected_branches/#protect-repository-branches - if you were to protect an unprotected branch, it would default to Maintainer level. So this assertion validates we retain that functionality even though we no longer unprotect / reprotect. 

I could make it clearer by setting the Merge Access Level to a different state when first protecting it though 👍

tests/unit/processors/test_branches_processor.py
TimKnight-DWP: Theoretically faster feedback loop if running tests locally, happy to strip them out

tests/acceptance/premium/test_branches.py
TimKnight-DWP: I test this in Standard too, so will remove from here and make the comment is more useful

tests/acceptance/standard/test_branches.py
TimKnight-DWP: Ah there is an existing separate test for that which I'd missed will remove

tests/unit/processors/test_branches_processor.py
TimKnight-DWP: Nope only * : https://docs.gitlab.com/user/project/repository/branches/protected/#use-wildcard-rules

Will have this return false

tests/acceptance/standard/test_branches.py
TimKnight-DWP: done

tests/acceptance/standard/test_branches.py
amimas: @TimKnight-DWP - I haven't looked at the code yet, but I believe in general GitLabForm doesn't know what is the default value for an attribute in an API.

If you follow GitLabForm's "[raw parameter passing](https://gitlabform.github.io/gitlabform/reference/#raw-parameters-passing)" design, it'll simply specify the attributes in API call that are in the config. The rest is either set by GitLab as default or set manually - GitLabForm doesn't interfere.

I'd say let's stick with that pattern for consistency instead of introducing exception here. It also keeps things simple. If I'm mistaken or have wrong understanding, please let me know.

tests/acceptance/standard/test_branches.py
amimas: Ahh... I see. So, we're trying to validate the default value set by gitlabform.

This seems related to the earlier discussion: https://github.com/gitlabform/gitlabform/pull/1070#discussion_r2255769632

Building on the linked comment, I feel we shouldn't have to test GitLab's default behaviour. If our logic is explicitely doing something around default behaviour, then it'd make sense, but I feel we shouldn't be doing that.

tests/unit/processors/test_branches_processor.py
amimas: Let's keep them since you've added already. Like you said, it does have some advantage.

amimas:
> > Tests run absolutely fine locally both individually and as part of the full acceptance test run (checked on two different machines).
> > So I'm a bit stumped as to why the fail when run on Github Actions
> 
> Fixed, I was trying to use Premium features in the standard tests. Doh!

I made the same mistake in the other PR for fixing variable deletion issue. It took a long time till I realized the above 🤦

tests/acceptance/conftest.py
```diff
@@ -317,15 +317,19 @@ def _make_user(level: AccessLevel = AccessLevel.DEVELOPER, add_to_project: bool
...
 
     yield user
 
     gl.users.delete(user.id)
 
 
+def create_project_member(gl, project_to_add_user_to, access_level_value: int = AccessLevel.DEVELOPER.value):
+    username = get_random_name("user")
+    user = _create_user_and_add_to_project(gl, project_to_add_user_to, username)
+    return user
```
amimas: I think you need to pass the `access_level_value` parameter when calling the `_create_user_and_add_to_project` function. Otherwise it'll always be the default value (i.e. developer).

tests/acceptance/premium/test_branches.py
```diff
@@ -1,247 +1,146 @@
...
                 protected: true
-                push_access_level: maintainer
-                merge_access_level: developer
-                unprotect_access_level: maintainer
+                allow_force_push: false
                 code_owner_approval_required: true
         """
 
-        run_gitlabform(protect_branch_with_code_owner_approval_required, project)
+        run_gitlabform(config_protect_branch, project.path_with_namespace)
```
amimas, line 36: I believe all other acceptance tests use the following pattern when calling the `run_gitlabform` function:

```
run_gitlabform(config_variable_name, project)
```

Why are we doing `project.path_with_namespace` here and other tests? What changed?

tests/acceptance/premium/test_branches.py
```diff
@@ -1,247 +1,146 @@
...
-        )
-        assert merge_access_user_ids == []
+        # If we were protecting a branch for the first time and not passing these values, Gitlab would default both
+        # to "False" as per: https://docs.gitlab.com/api/protected_branches/#protect-repository-branches
+        # We retain that decision-making rather than leaving the values at their previous state, otherwise
+        # if a User manually unprotected a branch, and restored protection by re-running Gitlabform, the flags would
+        # have silently changed
+        protected_branch = project.protectedbranches.get(branch)
+        assert protected_branch.allow_force_push is False
+        assert protected_branch.code_owner_approval_required is False
```
amimas: As mentioned in another thread, GitLabForm follows "raw parameter passing" design. That means, it'll pass on to GitLab API, whatever is given in the config. If some attributes are not provided and they are not required, it's up to GitLab on how to process them (i.e. what default to set).

We shouldn't need to validate the default values, which also **shouldn't change** based on attributes not being present in the config.

tests/acceptance/premium/test_branches.py
```diff
@@ -1,247 +1,146 @@
...
-                  - group_id: {group_for_function.id}
+                  - user_id: {first_user.id}
+                  - access_level: {AccessLevel.NO_ACCESS.value} 
+                  - user_id: {second_user.id}
+                  - user: {third_user.username}
                 allowed_to_merge:
                   - access_level: {AccessLevel.MAINTAINER.value}
-                  - user: {project_user_allowed_to_merge.username}
-                  - group_id: {group_for_function.id}
         """
```
amimas: Both `allowed_to_merge` and `allowed_to_push` are premium license feature. Let's test both of them accepting different inputs (i.e. username, user id, group name, group id, access level, etc.).

gitlabform/processors/project/branches_processor.py
```diff
@@ -33,9 +39,11 @@ def process_branch_protection(self, project: Project, branch_name: str, branch_c
         """
         protected_branch: Optional[ProjectProtectedBranch] = None
 
-        if not self.is_branch_name_wildcard(branch_name):
+        # If protected branch name contains a supported wildcard do not try looking it up
+        if not self.is_branch_name_supported_protected_branch_wildcard(branch_name):
```
amimas: Minor comment: not sure if the new function name `is_branch_name_supported_protected_branch_wildcard` is any better/clearer than existing name.

We're just checking if the branch name has a wildcard pattern, right? Maybe just call it like that (i.e. `branch_name_contains_supported_wildcard`)?

tests/acceptance/premium/test_branches.py
TimKnight-DWP: @amimas I agree on the design of GLF, however in this case we would potentially introduce a regression, as partially described in the comment.

Current functionality due to Unprotect - Reprotect, would be, if I removed `allow_force_push: True` from my config, it would end up being set to "False" when GLF is re-run, because when Protecting a branch, if `allow_force_push` is not supplied in the API request then Gitlab defaults to "False".

We would be requiring user's to set a value they may not already have done, due to getting it by default within both Gitlab and GitLabForm. If it's the decision as a whole we want to go down that route, fine, but we should consider making this a breaking change instead, as it will result in a change in functionality.


**Example:**
We on DWP automatically trigger a run of gitlabform against a given repository on reciept of a Webhook event from Gitlab describing a manual change in the UI to some repository setting. So let's say a Malicious User deliberately sets "allow_force_push" to True, in order to mess with the git history and hide something, and we haven't defined "allow_force_push" in the config (because it defaults to False), then we are unable to revert the manual change. Now we probably **should** define the configuration as we want it to be (and probably have).

tests/acceptance/premium/test_branches.py
TimKnight-DWP: Interesting... the standard tests all have always used project.path_with_namespace, but yeah the premium tests just used project. I assume our `run_gitlabform` function can handle both

TimKnight-DWP: Oh yeah inside `run_gitlabform` it does

```
    # Allow passing in GitLab REST objects; assume a full path string otherwise.
    if isinstance(target, Group):
        target = target.full_path  # type: ignore
    elif isinstance(target, Project):
        target = target.path_with_namespace  # type: ignore
```

So the only difference is doing the call outside that function and consistency with the standard tests. Honestly I have no strong opinions about which way we right them. I'll make it at least internally consistent to this file

amimas: Ahh... okay, that is interesting. Thanks for the context.

tests/acceptance/premium/test_branches.py
amimas: Thanks @TimKnight-DWP for the explanation. I can see why you're saying it's a regression, although I'm looking at that more like a bug - but I understand.

I suggest we follow the design/architecture of gitlabform. If that requires the release to be a major version, that should be fine in my opinion.

What do you think?

TimKnight-DWP: Works for me, I will update both the code and add some documentation explaining **Why** a major release and **what** people should do to migrate / ensure functionality

TimKnight-DWP: @amimas Done

docs/upgrade.md
```diff
@@ -2,6 +2,32 @@
 
 Some of these changes between major application versions may affect the effective configuration that would be applied if you run the application. Therefore to do the upgrade safely please follow this procedure.
 
+## From 3.\* to 4.\*
```
amimas: I think you meant this to be upgrading from version 4.x to 5.x, right? Currently released major version is 4 and I believe latest discussion is in this PR is to release it under next major release, which would be 5.x.

docs/upgrade.md
```diff
@@ -2,6 +2,32 @@
...
+
+### Branch Protection Changes
+
+When preventing GitlabForm from un-protecting and then re-protecting the Branch in order to apply configuration changes to the protection rules [#1070](https://github.com/gitlabform/gitlabform/pull/1070), we discovered some functionality that was previously hidden from the user.
+
+Previously if you stopped defining `code_owner_approval_required` or `allow_force_push` in the YAML config, the settings would revert to the default Gitlab settings applied when first protecting the branch, see: [Branch Protection API](https://docs.gitlab.com/api/protected_branches/#protect-repository-branches)
+
+With the changes to perform dynamic updates as required, we follow the GitlabForm convention of **leaving the Gitlab settings as is if the data is no longer defined in configuration**.
+
+Therefore, if you wish to ensure the Protected Branch settings are applied as you intend when running GitlabForm you should define some defaults explicitly:
```
amimas: Thanks for the writing this up. Couple of thoughts:

The topic seems more relevant for changelog since content seems to be explanation for the change/fix as opposed to "how to upgrade".

As for the content, I found it a little difficult to follow. Let me know if the following suggestion makes sense or feel free to update it as needed.

BTW, the unprotect and reprotect flow still exists for gitlab version older than X, right? We should probably call that out too.


```suggestion
In GitLabForm v4, branch protection config introduced a bug. GitLabForm would remove and re-add branch protection on every run, because GitLab's API did not support in-place updates. Besides causing audit noise because of constant change, this approach reset all unspecified attributes to GitLab defaults, breaking GitLabForm’s core behavior of [raw parameter passing](reference/index.md#raw-parameters-passing) — where only explicitly configured values are sent, and unspecified ones are left untouched.

This unintended consequence meant:

- A user configuring only some protection attributes (e.g. `push_access_level`) would find others reset to default (e.g. `code_owner_approval_required`) on re-run.
- GitLabForm was no longer idempotent unless all settings were always provided.

This issue is being fixed by following original intended behaviour of GitLabForm. Therefore, if you wish to ensure the Protected Branch settings are applied as you intend when running GitlabForm you should define some defaults explicitly:
```

amimas (commented):
Hey @TimKnight-DWP - Have become a little busy again. So, haven't been able to review the whole PR. Left couple of comments for the docs after a quick peak at it. Will try to catch up with it.

docs/upgrade.md
TimKnight-DWP: yes 😅

docs/upgrade.md
```diff
@@ -2,6 +2,33 @@
 
 Some of these changes between major application versions may affect the effective configuration that would be applied if you run the application. Therefore to do the upgrade safely please follow this procedure.
 
+## From 3.\* to 4.\*
```
TimKnight-DWP:
```suggestion
## From 4.\* to 5.\*
```

gitlabform/processors/project/branches_processor.py
```diff
@@ -50,35 +58,102 @@ def process_branch_protection(self, project: Project, branch_name: str, branch_c
...
+                    self.protect_branch(project, branch_name, branch_config, False)
+                else:
+                    # Update the Branch Protection Rules
+                    verbose(f"Updating branch protection for {branch_name}")
+
+                    code_owner_approval_required_data = self.get_code_owner_approval_required_data(
+                        branch_config, protected_branch
+                    )
+
+                    allow_force_push_data = self.get_allow_force_push_data(branch_config, protected_branch)
```
amimas: Related to this discussion: https://github.com/gitlabform/gitlabform/pull/1070#discussion_r2261837750

We're just passing to GitLab API if these attributes (e.g. `code_owner_approval_required`, `allow_force_push`) are set in the config and their values (e.g. true, false). Why do we still need special logic here?

gitlabform/processors/project/branches_processor.py
```diff
@@ -50,35 +58,102 @@ def process_branch_protection(self, project: Project, branch_name: str, branch_c
...
+                        info(f"Updating protected branch {branch_name} with {branch_config_prepared_for_update}")
+                        self.protect_branch(project, branch_name, branch_config_prepared_for_update, True)
             elif not protected_branch:
-                self.protect_branch(project, branch_name, branch_config)
+                info(f"Creating branch protection for {branch_name}")
+                self.protect_branch(project, branch_name, branch_config, False)
         elif protected_branch and not branch_config.get("protected"):
-            self.unprotect_branch(protected_branch)
+            info(f"Removing branch protection for {branch_name}")
+            self._unprotect_branch(protected_branch)
```
amimas: I know there's no change in logic here, but reading through the code got me thinking...

Here it seems like we're treating `protected` key to be optional. For example, imagine following config where we omitted or removed the 'protected' key:

```diff
projects_and_groups:
  group_1/project_1:
    branches:
      main:
-       protected: true
        push_access_level: no access
        merge_access_level: developer
        unprotect_access_level: maintainer
```

If the branch is already protected, absence of `protected` key would mean the branch protection should be removed. You can probably prove it by adding a simple test. I feel that's wrong or not intuitive at all. And if we read [gitlabform's documentation on branch protection](https://gitlabform.github.io/gitlabform/reference/protected_branches/), `protected` key is mandatory.

Since this PR is making breaking change anyways, now is probably the best time to address this. To do that, similar to `self._needs_update` function, I think we can utilize `self._can_proceed` function. It needs to be defined with appropriate logic to return boolean (e.g. ensure `protected` key exists and is a boolean type).

Thoughts?

gitlabform/processors/project/branches_processor.py
```diff
@@ -50,35 +58,102 @@ def process_branch_protection(self, project: Project, branch_name: str, branch_c
...
-                # protected branch is not the same. So, removing existing branch protection and
-                # reconfiguring branch protection seems simpler.
-                self.unprotect_branch(protected_branch)
-                self.protect_branch(project, branch_name, branch_config)
+            if protected_branch:
+                # Branch matching this name exists, so check for updates
+
+                # Transform the config from "allowed_to_push" or "push_access_level" into "push_access_levels" array to
+                # match how the data is stored in the Gitlab backend
+                branch_config_for_needs_update = self.transform_branch_config_access_levels(branch_config)
```
amimas: Minor comment: can we come up with a better name for this variable 😄 . Don't know if this is any better or if it makes sense in this context, but how about:

```suggestion
                branch_protection_api_payload = self.transform_branch_config_access_levels(branch_config)
```

gitlabform/processors/project/branches_processor.py
```diff
@@ -50,35 +58,102 @@ def process_branch_protection(self, project: Project, branch_name: str, branch_c
...
+                        branch_config_prepared_for_update["allow_force_push"] = allow_force_push_data
+
+                    if branch_config_prepared_for_update != {}:
+                        # We have some updates to Access Levels to make
+                        info(f"Updating protected branch {branch_name} with {branch_config_prepared_for_update}")
+                        self.protect_branch(project, branch_name, branch_config_prepared_for_update, True)
             elif not protected_branch:
-                self.protect_branch(project, branch_name, branch_config)
+                info(f"Creating branch protection for {branch_name}")
+                self.protect_branch(project, branch_name, branch_config, False)
```
amimas: When I was reading through the `process_branch_protection` function, it was a bit confusing because `protect_branch` function is being called in different scenarios, but the branch config format or data structure looked different. I think you mentioned somewhere that it's because of GitLab's API not being consistent between create vs update.

If my understanding is correct, could we add some comment about it? Not sure what location is the best for this.

gitlabform/processors/project/branches_processor.py
```diff
@@ -87,22 +162,19 @@ def protect_branch(self, project: Project, branch_name: str, branch_config: dict
                     exit_code=EXIT_PROCESSING_ERROR,
                 )
             else:
-                warning(message)
+                error(message)
 
-    def unprotect_branch(self, protected_branch: ProjectProtectedBranch):
+    def _unprotect_branch(self, protected_branch: ProjectProtectedBranch):
```
amimas: I suggest reverting this change. It's kind of inconsistent with `protect_branch` function's naming pattern. Or, we update the other functions to start with `_`.

gitlabform/processors/project/branches_processor.py
```diff
@@ -137,5 +209,270 @@ def transform_branch_config(self, branch_config: dict):
...
+
+            # Create a new record for the new Access Level required and mark the existing records for deletion
+            access_items_to_patch.append(
+                {
+                    "access_level": access_level_value,
+                }
+            )
+
+            # Mark records for deletion
+            access_items_to_patch.extend(BranchesProcessor.create_delete_data(existing_records, access_items_to_patch))
```
amimas: I'm new to this pattern. Just for my learning, could you point me to a document/tutorial so that I can understand why `acces_items_to_patch.extends` function's first parameter is referring to `BranchesProcess.create_delete_data` instead of `create_delete_data`?

TimKnight-DWP: `.extends` adds a whole array to the end of itself (rather than for each x, array.add(x)).

create_delete_data can probably be named a bit more coherently

gitlabform/processors/project/branches_processor.py
TimKnight-DWP: @amimas agreed, we are iterating through a list of branches defined in config here so protected should be present, if it's "None" we should throw an exception

gitlabform/processors/project/branches_processor.py
TimKnight-DWP: @amimas I have moved some code around, and extracted a function to hopefully make it a bit easier to read, fewer nested if elifs

gitlabform/processors/project/branches_processor.py
TimKnight-DWP: Reworked since I was already mutating one of the input arrays, there was no need to try and make a totally "functional" method here, and as such have also removed extra loop

docs/upgrade.md
```diff
@@ -2,6 +2,48 @@
...
+```yaml
+projects_and_groups:
+ my-group/*:
+   branches:
+     main:
+       allow_force_push: false
+       code_owner_approval_required: true
+     master:
+       protected: true
+```
```
amimas: Did you intend to include `master` in the example config? If we're following gitlab's API, `master` branch config is still valid and shouldn't fail. But, `main` branch config should fail since it's missing gitlabform's required attribute `protected`.

Let's clarify the explanation or remove `master` from the example.

gitlabform/processors/project/branches_processor.py
amimas: Hi @TimKnight-DWP - Not sure if you've already explained this. I'm still seeing special logic in the code like below in latest commit.

```
            verbose("Creating data to update code_owner_approval_required as necessary")
            code_owner_approval_required_patch_data = None
            code_owner_approval_required_config = transformed_branch_config.get("code_owner_approval_required")
            if (
                code_owner_approval_required_config is not None
                and protected_branch.code_owner_approval_required != code_owner_approval_required_config
            ):
                code_owner_approval_required_patch_data = transformed_branch_config.get("code_owner_approval_required")

            verbose("Creating data to update allow_force_push as necessary")
            allow_force_push_patch_data = None
            allow_force_push_config = transformed_branch_config.get("allow_force_push")
            if allow_force_push_config is not None and protected_branch.allow_force_push != allow_force_push_config:
                allow_force_push_patch_data = allow_force_push_config

...

            if code_owner_approval_required_patch_data is not None:
                protected_branch_api_patch_data["code_owner_approval_required"] = (
                    code_owner_approval_required_patch_data
                )

            if allow_force_push_patch_data is not None:
                protected_branch_api_patch_data["allow_force_push"] = allow_force_push_patch_data

```

Why do we still need to do this?

gitlabform/processors/project/branches_processor.py
```diff
@@ -137,5 +258,218 @@ def transform_branch_config(self, branch_config: dict):
...
+            # User has not defined either x_access_level OR allowed_to_x in their configuration
+            # We should follow Gitlab convention and ensure the protected branch is set to Maintainer access level
+            access_level_value = AccessLevel.MAINTAINER.value
+
+            # Create a new record for the new Access Level required and mark the existing records for deletion
+            patch_data.append(
+                {
+                    "access_level": access_level_value,
+                }
+            )
```
amimas: GitLabForm shouldn't decide or set a default value. We should leave that up to GitLab.

TimKnight-DWP: @amimas unfortunately we **have** to pass something if we are using the Update() method, otherwise Gitlab will reject the full data packet.

So I have set it to the Default that Gitlab would set to if initially creating a Protected Branch

amimas: Thanks @TimKnight-DWP for clarifying. Not sure I follow or understand it though. Could you share a config example and the steps when update method will be called but there's no data? Also, I think there's logic to call the update method only when needed, right - instead of making that API call in every run of gitlabform?

amimas:
> Won't let me reply to the one about allow_force_push:
> 
> This is all about not making changes that we don't need to
> 
> So let's say _nothing at all_ has changed, current logic means that it'll run through and find no update to make, so won't make an API call
> 
> Let's say only push_access_level has changed, we don't _want_ to send data related to the code_owner_approval_required setting etc, because those have not changed.

Thanks @TimKnight-DWP  for explaining. Few follow up questions:

> Let's say only push_access_level has changed, we don't _want_ to send data related to the code_owner_approval_required setting etc, because those have not changed.

What's wrong with including all the data where some attributes didn't change? Isn't that how all the other processors work today? My understanding is that all attributes are sent where at least one attribute changed. And update is called only if one of the attributes have changed.

Another issue is that now we have special logic for individual attributes of branch protection. With this pattern, doesn't it mean we have to add similar logic if/when gitlab adds or removes attributes from the API as part of feature add, removal, etc? Unlike most other cases, the processor's logic now needs to be in synch with gitlab's feature.

Please let me know if I missed something. Thanks!

TimKnight-DWP:
> What's wrong with including all the data where some attributes didn't change? Isn't that how all the other processors work today? My understanding is that all attributes are sent where at least one attribute changed. And update is called only if one of the attributes have changed.

I can give it another try, my recollection from running the service tests is that Gitlab would return a `200` but not actually make the changes if you gave it more than it needed. I will double check. I can try and tidy up.

> Another issue is that now we have special logic for individual attributes of branch protection. With this pattern, doesn't it mean we have to add similar logic if/when gitlab adds or removes attributes from the API as part of feature add, removal, etc? Unlike most other cases, the processor's logic now needs to be in synch with gitlab's feature.

So I am not using _needs_update for this section deliberately as even with the great update made at the start of the branch, it was very flaky in detecting if updates were actually needed (sometimes not updating when it was needed), I think due to how Gitlab returns data compared to how it takes it.

I 100% agree it's tightly coupled to Gitlab's API, I would prefer it not to, but their weird API choices here puts us in a bind, at some point we have to tightly couple between the GET differing to the UPDATE/PUT :( 

@amimas - thanks for all the comments and challenges, they're great, I want us to get to a point this is maintainable, whilst navigating a tricky API from GL

gitlabform/processors/project/branches_processor.py
TimKnight-DWP: So in this specific example, let's say you had:
```yaml
        projects_and_groups:
          {project_for_function.path_with_namespace}:
            branches:
              {branch_for_function}:
                protected: true
                allowed_to_push:
                  - user_id: {first_user.id}
                  - access_level: {AccessLevel.NO_ACCESS.value} 
                  - user_id: {second_user.id}
                  - user: {third_user.username}
```

If you now change that to:
```yaml
        projects_and_groups:
          {project_for_function.path_with_namespace}:
            branches:
              {branch_for_function}:
                protected: true
```

What do we do? My belief is that the intent as a user is to remove `allowed_to_push` from all those previously defined. At which point we have to set `allowed_to_push` to _something_ - when unprotecting and reprotecting the branch as before, it would default back to "Maintainer" access as that's what Gitlab does by default. In terms of logic in the processor we have to build packets to delete the other push_access_level entries: https://docs.gitlab.com/api/protected_branches/#example-delete-a-push_access_level-record and then add one to create the maintainer level entry.


Now should we mandate one of `allowed_to_push` with at least an `acces_level` entry or a `push_access_level` entry in the config when defining Gitlab config.
I personally feel that's even more of a breaking change than the one to mandate `protected: true/false` as these entries are **not** mandatory in the Gitlab API

amimas: Thanks for sharing that example. That helps me to understand your point of view.

> What do we do? My belief is that the intent as a user is to remove allowed_to_push from all those previously defined.

This is a good example and exactly why we need to follow the raw parameter passing design that gitlabform started with. In your example, if the user removed an attribute from the config (or manually created branch protection and now using gitlabform by using only some config options), gitlabform shouldn't do anything with unconfigured attributes - because the intent is not clear.

GitlabForm should simply call the update function, if needed, and pass all the attributes that are in the config. The rest is up to GitLab's API on what it will do.

Hope that makes sense.

amimas: I just had another thought.

I think the scenario/example you gave is similar to gitlabform feature in some processors supporting `enforce` key or in the case of members config, `keep_bots` that compliments the `enforce` key.

I can see something like that being used to support your example - but the point is it's done via explicit config. There's no ambiguity. In this case `enforce` probably doesn't make sense - maybe `reset_to_default` or something like it. Not sure if we need it though but that's a discussion for separate issue.

TimKnight-DWP: @amimas I agree it does make more sense with an `enforce` flag as the user is saying "No I mean enforce this specific configuration - regardless of what is in GL but no longer defined here". 

That would solve _our_ specific need without changing the fundamental paradigm of gitlabform.

There might be a better word but `enforce` is a consistently used keyword in GitlabForm outside of the GL API so I would suggest we keep it

amimas: Let's keep this PR as minimal as possible please. The goal of this PR is to address some issues.

The enforce option would be a new feature from gitlabform and should be done as a separate PR. To be frank, not sure we need this feature, as I mentioned above but that can be discussed in separate issue and addressed via separate PR.

TimKnight-DWP: Sounds good @amimas - how would you like me to handle the "defaulting" aspect - leave as currently implemented (which is consistent with the functionality prior to this change even if it wasn't quite how GLF likes to operate)?

amimas: I think we should remove any explicit logic involving default settings or configuring attributes that are not present in gitlabform's config.

If I understand correctly, the main challenge is to detect if an update is needed. If update is needed, pass all the attributes from the config in a manner acceptable by GitLab API.

Let me try give a bit more context or explanation around the topic of options getting reset to default based on my understanding. This is not a feature of gitlabform - it's a result of gitlab issue. Gitlab didn't have an API for updating branch protection. So gitlabform always performed unprotect-reptotect flow. And during reprotect, gitlabform follows raw parameter passing pattern instead of maintaining API attributes in the code. So, during reprotect, unconfigured attributes gets set to default values by GitLab. This is how it always has worked. Unfortunately in the recent update to gitlabform's branch protection processor, the "needs update" logic didn't work as well as it did before. So, it amplified that reset issue along with audit noise. Plus, keep in mind that gitlab's API for updating branch protection only exist for certain version and above. So this issue will continue to exist.

Please let me know if I'm mistaken.

TimKnight-DWP: The unprotect-reprotect flow also resulted in a security-flaw, where if the API timed-out/rate-limited it could leave something unprotected (we had someone push straight to `main` as a result).

I _think_ it should be possible to leave the existing setting as is if it's removed from config. That was where the defaulting came in. I'll crack on the testing 👍

TimKnight-DWP: @amimas updated - all the tests run as expected (had to change one assertion that's all)

amimas: Thanks for the update. Let's close this thread.

TimKnight-DWP:
Re:
`Gitlab would return a 200 but not actually make the changes if you gave it more than it needed`

I think I have now fixed that, so it will happily send more across, that comment was probably a remnant in my thoughts from when I was using the PATCH api via direct HTTP call rather than python-gitlab. Either way _most_ of the logic now follows pure Raw Parameter passing

PR #1076, opened 2025-08-05
fix: group saml links config changes should be applied

### LINKED ISSUE #994: SAML: saml_group_links does not update access_level for existing groups
https://github.com/gitlabform/gitlabform/blob/f96efbe89ad4ae707a5d55ff476be825b8bb16d2/gitlabform/processors/group/group_saml_links_processor.py#L33

I have noticed that when configuring `saml_group_links` for a group, any later updates to `access_level` for those groups are ignored (it seems like the desired behavior would be to delete and re-create the SAML link with the new `access_level`

amimas:
This PR addresses the bug reported in #994 . Currently config changes to saml links are being silently ignored. It simply "creates" a saml link if it does not exist.

With the changes in this PR, updates will be performed if required. However, GitLab's API currently does not support updates to an existing link. So, in this PR update is performed by deleting the existing link and recreating with latest config.

Additionally, moved the acceptance tests into its own separate test suite to follow this repo's test convention.

fixes #994

### DIFF
--- a/gitlabform/processors/group/group_saml_links_processor.py
+++ b/gitlabform/processors/group/group_saml_links_processor.py
@@ -1,9 +1,8 @@
 from logging import debug
-from typing import List, Union
+from typing import List
 
 from gitlabform.gitlab import GitLab
-from gitlab.base import RESTObject, RESTObjectList
-from gitlab.v4.objects import Group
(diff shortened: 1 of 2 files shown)


TimKnight-DWP:
Neat and simple fix @amimas  nice

PR #1097, opened 2025-09-08
feat!: release version 5

TimKnight-DWP:
PR opened so it's easier to rebase onto `main` for dependabot changes and v4 features

----
Should be released when we're happy with stability and more v5 features added

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,37 @@
 # Changelog
 
+## 5.0.0-beta.1
+
+### Breaking Changes
+
+#### Branch Protection Updates [#1070](https://github.com/gitlabform/gitlabform/pull/1070) ([TimKnight-DWP](https://github.com/TimKnight-DWP) & [br0ziliy](https://github.com/br0ziliy))
+When preventing GitlabForm from un-protecting and then re-protecting the Branch in order to apply configuration changes to the protection rules [#1070](https://github.com/gitlabform/gitlabform/pull/1070), we discovered some functionality that was previously hidden from the user.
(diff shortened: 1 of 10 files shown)


docs/changelog.md
```diff
@@ -49,13 +107,6 @@ Thanks to all the contributors of this release!
 
 * Suppress false positive bandit warning for private_token [#1163](https://github.com/gitlabform/gitlabform/pull/1163). ([amimas](https://github.com/amimas))
 
-### Dependencies
-
-* Update various dependencies to newer version.
-
-Thanks to all the contributors of this release!
-
-
```
amimas, line 58: Somehow this got removed. I'm guessing a mishap during rebase. We should restore it before merging.

docs/changelog.md
```diff
@@ -1,5 +1,63 @@
...
+
+## 5.0.0-beta.4
+
+- Rebased onto v4.7.0
+
+## 5.0.0-beta.3
+
+### Chore
+
+- Rebased onto v4.6.1
```
amimas: I think one of these weren't just rebase. They had some changes. We should correct it before merging it.

amimas (commented):
I wonder if the changelog for v5 needs a bit of refactoring. For example, beta 1 and beta 2 includes details about why the change is made and what config should be used. Maybe these belongs in the upgrade doc?

And for merging this to main branch, should we keep the changelog for all the beta? We need a clear changelog for v5.0.0. This should include all the info from the beta versions.

What's your thought @TimKnight01

PR #1140, opened 2025-12-11
feat: Allow for any kwarg to be passed to python-gitlab by GitlabWrapper

thomascizeron:
Solves #1025

This PR enables GitlabForm to pass any additional configuration parameters from the `gitlab` section directly to the underlying python-gitlab Gitlab client. This provides users with flexibility to configure advanced `python-gitlab` features without requiring GitLabForm updates.

### DIFF
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -49,7 +49,7 @@ gitlab:
   # For following migration to python-gitlab, see: https://github.com/gitlabform/gitlabform/issues/73
   #
   # For example:
-  # retry_transient_errors: false # retry requests that fail due to transient errors
+  # retry_transient_errors: false # (defaults to `true`) retry requests that fail due to transient errors
   # obey_rate_limit: true  # automatically wait if rate limit is hit
   # max_retries: 5      # number of times to retry failed requests
 


docs/reference/index.md
```diff
@@ -41,6 +41,16 @@ gitlab:
...
+  # whether to retry requests that fail due to transient errors (network issues, rate limits, etc.)
+  # (defaults to `true`)
+  retry_transient_errors: true
+  
+  # ** advanced parameters **
+  # Any additional parameters specified here will be passed directly to the python-gitlab library.
+  # See https://python-gitlab.readthedocs.io/en/stable/api-usage.html#gitlab-client for available options.
+  # For example:
+  # obey_rate_limit: true  # automatically wait if rate limit is hit
+  # max_retries: 5      # number of times to retry failed requests
```
amimas: Couple of comments on this doc:

GitLabForm hasn't yet 100% migrated to use python-gitlab. Some of the processors are still using legacy code. So, we should clarify that these additional options won't be applied for those processors. It's only applicable for the processors that are using the python-gitlab library under the hood. What's the best way to clearly communicate this?

Why are you explicitly adding `retry_transient_errors` in the config instead of in the example section below? Let's keep it consistent. I suggest moving this to the example. Otherwise it appears as though this is supported globally - see my note above.

gitlabform/gitlab/core.py
amimas, line 1: For the purpose of this PR (extending support for python-gitlab configuration), are the changes in this file necessary? Just trying to understand what is being fixed or improved.

amimas (commented):
Thank you so much for looking into this and the PR. Like the extra tests being included. Left some comments. Please let me know if you have any questions.

gitlabform/gitlab/core.py
thomascizeron: They are needed as a self.gitlab_config is construct here to be passed to the GitlabWrapper. It seemed the simpler way to do it

docs/reference/index.md
thomascizeron: Will try to do it, is there a list/advancement report on the python-gitlab migration ? Maybe we should reference #73 to refer readers to the ongoing progress ?

It's because it's one of the default value, but i agree it is simpler in the example section

thomascizeron:
```suggestion
  # ** advanced parameters **
  # Any additional parameters specified here will be passed directly to the python-gitlab library.
  # See https://python-gitlab.readthedocs.io/en/stable/api-usage.html#gitlab-client for available options.
  #
  # GitlabForm hasn't yet 100% migrated to use python-gitlab. Some processors are not using it and wont use these parameters.
  # For following works on the migration, see: https://github.com/gitlabform/gitlabform/issues/73
  #
  # For example:
  # retry_transient_errors: false # retry requests that fail due to transient errors
  # obey_rate_limit: true  # automatically wait if rate limit is hit
  # max_retries: 5      # number of times to retry failed requests
```

docs/reference/index.md
```diff
@@ -41,6 +41,17 @@ gitlab:
   ssl_verify: true
   # timeout for the whole requests to the GitLab API, in seconds
   timeout: 10
+  # ** advanced parameters **
+  # Any additional parameters specified here will be passed directly to the python-gitlab library.
+  # See https://python-gitlab.readthedocs.io/en/stable/api-usage.html#gitlab-client for available options.
+  #
+  # GitlabForm hasn't yet 100% migrated to use python-gitlab. Some processors are not using it and wont use these parameters.
+  # For following works on the migration, see: https://github.com/gitlabform/gitlabform/issues/73
```
thomascizeron:
```suggestion
  # For following migration to python-gitlab, see: https://github.com/gitlabform/gitlabform/issues/73
```

docs/reference/index.md
amimas: Yeah, that's the closest issue but I can see it's already slightly outdated. There's also this project but didn't get a chance to populate all the todos yet.

https://github.com/orgs/gitlabform/projects/1/views/1

gitlabform/gitlab/core.py
amimas: Thanks for clarifying.

docs/reference/index.md
```diff
@@ -41,6 +41,17 @@ gitlab:
...
   timeout: 10
+  # ** advanced parameters **
+  # Any additional parameters specified here will be passed directly to the python-gitlab library.
+  # See https://python-gitlab.readthedocs.io/en/stable/api-usage.html#gitlab-client for available options.
+  #
+  # GitlabForm hasn't yet 100% migrated to use python-gitlab. Some processors are not using it and wont use these parameters.
+  # For following migration to python-gitlab, see: https://github.com/gitlabform/gitlabform/issues/73
+  #
+  # For example:
+  # retry_transient_errors: false # retry requests that fail due to transient errors
```
amimas: The `retry_transient_errors` is set to `False` as default in python-gitlab but it's set to `True` as default from gitlabform - right?

I liked that you had specified earlier about the default config. Can we keep it that way?

thomascizeron: Yes it defaults to false in python-gitlab (https://python-gitlab.readthedocs.io/en/stable/api/gitlab.html)
and true in gitlabform

I've added it to the documentation

docs/reference/index.md
thomascizeron: I've refered to the issue, the project is linked in it but if you prefer I can directly link the board 😊

amimas: Sounds good.

PR #1151, opened 2026-01-02
fix: Split kwargs passed to gitlab client and graphql client

thomascizeron:
Linked to #1140 

Problem: Turns out python-gitlab dont pass kwargs to the graphql client, some kwargs would crash the init of the client.

Solution: Added explicit parameter whitelists in GitlabWrapper to pass parameters to the gitlab client or the graphql client

I have tested setting max_retries against a personnal gitlab instance and it worked. Dont know if an acceptance test would be useful here, let me know

### DIFF
--- a/.github/workflows/_prs.yml
+++ b/.github/workflows/_prs.yml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           ref: ${{ github.event.pull_request.head.sha }}
       # Initializes the CodeQL tools for scanning.
@@ -67,7 +67,7 @@ jobs:
       fail-fast : false
     steps:
       - name: Checkout code from the PR
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 2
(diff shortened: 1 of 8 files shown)


gitlabform/gitlab/__init__.py
```diff
@@ -67,10 +67,48 @@ class GitLab(
...
+    GRAPHQL_PARAMS = {
+        "ssl_verify",
+        "client",
+        "timeout",
+        "user_agent",
+        "fetch_schema_from_transport",
+        "max_retries",
+        "obey_rate_limit",
+        "retry_transient_errors",
+    }
```
amimas: Is it possible to fix it in a way that gitlabform does not need to maintain a list of attributes for python-gitlab? I'm not experienced enough. From a very quick/short research, could we possibly use python's `inspect` module? It provides a `signature` method that can possibly let us detect the method signature for graphql  client.

- https://realpython.com/ref/stdlib/inspect/
- https://docs.python.org/3/library/inspect.html#inspect.signature

amimas (commented):
Thank you so much for looking into this. I only have one concern about maintaining a hard coded list of attributes for an external library. This is probably okay as a last resort. Would you be able to explore other alternatives?

I'm not experienced enough to suggest required code changes but left a comment below with a suggestion. Please take a look and let me know your thoughts or if you could try it out.

gitlabform/gitlab/__init__.py
thomascizeron: It's a great idea ! didn't thought of it, I've added it

amimas: Thanks for exploring the suggestion. Glad you've made it dynamic.

amimas (approved):
Looks good. Thank you again for the quick fix. I'll merge it now and will look into releasing this soon along with the original feature you've contributed.

PR #1152, opened 2026-01-03
refactor: migrate project security settings to python-gitlab

thomascizeron:
This MR migrates the `ProjectSecuritySettingsProcessor` from the legacy `SingleEntityProcessor` pattern to use the python-gitlab library directly, as part of issue #73.

- Migrated processor: `ProjectSecuritySettingsProcessor` now extends `AbstractProcessor` and uses `python-gitlab`'s `http_get`/`http_put` methods instead of legacy `_make_requests_to_api()`
- Removed legacy code: Deleted `GitlabProjectSecuritySettings` wrapper class and removed it from `GitLab` class inheritance
- Implemented dry-run support: Added _print_diff method using `DifferenceLogger` to show configuration changes in dry-run mode
- Updated tests: Acceptance test now instantiates and tests the processor directly

### DIFF
--- a/gitlabform/processors/project/project_security_settings.py
+++ b/gitlabform/processors/project/project_security_settings.py
@@ -41,7 +41,7 @@ class ProjectSecuritySettingsProcessor(AbstractProcessor):
             warning(f"Failed to get project security settings for project {project.path_with_namespace}: {e}")
             return {}
 
-    def _update_project_security_settings(self, project: Project, initial_settings: dict, bsettings: dict) -> None:
+    def _update_project_security_settings(self, project: Project, initial_settings: dict, settings: dict) -> None:
         """Update project security settings using python-gitlab."""
         try:
             # TODO: python-gitlab does not yet support updating project security settings


gitlabform/processors/project/project_security_settings.py
```diff
@@ -1,12 +1,53 @@
...
+class ProjectSecuritySettingsProcessor(AbstractProcessor):
     def __init__(self, gitlab: GitLab):
-        super().__init__(
-            "project_security_settings",
-            gitlab,
-            get_method_name="get_project_security_settings",
-            edit_method_name="put_project_security_settings",
+        super().__init__("project_security_settings", gitlab)
+
+    def _process_configuration(self, project_or_project_and_group: str, configuration: dict) -> None:
```
amimas: In the context of this processor, we know the entity type is project - not a group. So, let's rename `project_or_project_and_group` to `project`. It'll make the code easier to read.

gitlabform/processors/project/project_security_settings.py
```diff
@@ -1,12 +1,53 @@
...
+        debug("Processing project security settings...")
+        project: Project = self.gl.get_project_by_path_cached(project_or_project_and_group)
+
+        security_settings_in_config = configuration.get("project_security_settings", {})
+        security_settings_in_gitlab = self.get_project_security_settings(project_or_project_and_group)
```
amimas: I think you can pass the `project` variable to the function. That way you already have access to the project object and won't need to retrieve it again in that function.

gitlabform/processors/project/project_security_settings.py
```diff
@@ -1,12 +1,53 @@
...
+
+    def get_project_security_settings(self, project_path: str) -> dict:
+        """Get project security settings from GitLab using python-gitlab."""
+        project: Project = self.gl.get_project_by_path_cached(project_path)
+        # Use lower-level http_get as security_settings doesn't have a dedicated manager
+        path = f"/projects/{project.encoded_id}/security_settings"
+        result = self.gl.http_get(path)
+        # http_get can return Response for streamed requests, but we're not streaming
+        # so it will always be a dict
+        return cast(dict[str, Any], result)
```
amimas: Let's wrap it in `try-except` block. I'd also add a todo for future reference. I had to do something like this recently. Have a look at here. Let me know if this makes sense.

https://github.com/gitlabform/gitlabform/blob/7cb585a7a4e20af2b06570253a893f45f1e01bf0/gitlabform/processors/project/remote_mirrors_processor.py#L147-L159

gitlabform/processors/project/project_security_settings.py
```diff
@@ -1,12 +1,53 @@
...
+
+    def _update_project_security_settings(self, project: Project, settings: dict) -> None:
+        """Update project security settings using python-gitlab."""
+        path = f"/projects/{project.encoded_id}/security_settings"
+        self.gl.http_put(path, post_data=settings)
```
amimas: Same comment as above. Let's make sure we catch errors/exceptions and handle it properly.

gitlabform/processors/project/project_security_settings.py
```diff
@@ -1,12 +1,53 @@
...
     def __init__(self, gitlab: GitLab):
-        super().__init__(
-            "project_security_settings",
-            gitlab,
-            get_method_name="get_project_security_settings",
-            edit_method_name="put_project_security_settings",
+        super().__init__("project_security_settings", gitlab)
+
+    def _process_configuration(self, project_or_project_and_group: str, configuration: dict) -> None:
+        debug("Processing project security settings...")
```
amimas: The `AbstractProcessor` already takes care of logging this. So, let's remove it from here.

gitlabform/processors/project/project_security_settings.py
```diff
@@ -1,12 +1,53 @@
...
+
+        security_settings_in_config = configuration.get("project_security_settings", {})
+        security_settings_in_gitlab = self.get_project_security_settings(project_or_project_and_group)
+        debug(security_settings_in_gitlab)
+        debug("project_security_settings BEFORE: ^^^")
+
+        if self._needs_update(security_settings_in_gitlab, security_settings_in_config):
+            debug("Updating project security settings")
+            self._update_project_security_settings(project, security_settings_in_config)
+            debug("project_security_settings AFTER: ^^^")
```
amimas: The `_update_project_security_settings` function doesn't print anything. So the next line showing "AFTER: ^^^" won't refer to anything.

tests/unit/processors/test_project_security_settings_processor.py
amimas, line 1: Typically in this project we have acceptance tests for processors. It uses live gitlab instance to test out the processor logic. What's the rational for this unit test here? If something is not covered by the acceptance tests, maybe we should add it there?

gitlabform/processors/single_entity_processor.py
amimas, line 1: Should we remove it separately in a different PR? We'd need to update the contributing docs too.

gitlabform/processors/project/project_security_settings.py
thomascizeron: As `project` is already taken in the function, I opted for `project_name`

gitlabform/processors/project/project_security_settings.py
thomascizeron: Tell me if the new version match what you meant :)

tests/unit/processors/test_project_security_settings_processor.py
thomascizeron: I thought that some unit tests won't hurt, but I see the point, there is not much internal logic here, acceptance tests should be sufficient

tests/acceptance/standard/test_project_settings.py
amimas, line 1: Nothing wrong here with the update in this file adding docstrings. But it's not related to project security settings refactoring. So, let's take it out of this PR. Want to minimize/isolate changes that are not related.

gitlabform/processors/project/project_security_settings.py
```diff
@@ -1,12 +1,66 @@
...
+            debug("No update needed for project security settings")
+
+    def get_project_security_settings(self, project: Project) -> dict:
+        """Retrieve project security settings using python-gitlab."""
+        try:
+            # TODO: python-gitlab does not yet support retrieving project security settings
+            # via its dedicated manager, so we use a lower-level http_get method here.
+            # Switch to native method once supported by python-gitlab.
+
+            path = f"/projects/{project.encoded_id}/security_settings"
```
amimas: So far the pattern used is `project.id`. Any specific reason for using `project.encoded_id`?

gitlabform/processors/project/project_security_settings.py
amimas: Looks like latest update removed that log entirely. Maybe after the update, we should get the setting and then print that debug line to keep the functionality same as before?

tests/unit/processors/test_project_security_settings_processor.py
amimas: Yeah, thanks for thinking of unit tests. In this case there aren't much to test via unit test, I think.

gitlabform/processors/single_entity_processor.py
thomascizeron: https://github.com/gitlabform/gitlabform/pull/1164

gitlabform/processors/project/project_security_settings.py
thomascizeron: Ensure that the ID is url-encoded so that it can be safely used in a URL 

cf. https://python-gitlab.readthedocs.io/en/stable/api/gitlab.html#gitlab.base.RESTObject.encoded_id

I think it is safer to use encoded_id when dealing with URLs

gitlabform/processors/project/project_security_settings.py
thomascizeron: My bad thought the debug info were not needed. I've put the debug directly in `_update_project_security_settings` in latest commits

gitlabform/processors/project/project_security_settings.py
```diff
@@ -1,12 +1,68 @@
...
-            edit_method_name="put_project_security_settings",
+        super().__init__("project_security_settings", gitlab)
+
+    def _process_configuration(self, project_name: str, configuration: dict) -> None:
+        project: Project = self.gl.get_project_by_path_cached(project_name)
+
+        security_settings_in_config = configuration.get("project_security_settings", {})
+        security_settings_in_gitlab = self.get_project_security_settings(project)
+        debug(security_settings_in_gitlab)
+        debug("project_security_settings BEFORE: ^^^")
```
amimas: Didn't think of it earlier. I'm thinking it's probably best to move this logic of printing before-after, inside the `_update_project_security_settings` function. Otherwise, the "BEFORE: ^^^" will always print but sometimes may not print the "AFTER: ^^^". Does that make sense?

What do you think?

gitlabform/processors/project/project_security_settings.py
amimas: Yup, agree. But, `project.id` gives you the integer - no need to use the URL in the requests. Let me know if I missed something.

gitlabform/processors/project/project_security_settings.py
thomascizeron: Make sense !

gitlabform/processors/project/project_security_settings.py
thomascizeron: Mmmh i see

amimas:
Thanks for all the updates. Mostly looks good. Could you please address the linter reported issue? I think an expected variable was not passed into a function when it is being called.

PR #1154, opened 2026-01-05
feat: pass max_retries, retry_transient_errors and backoff_factor to legacy Gitlab API client

thomascizeron:
Solves #1025 

In addition to passing arguments to the python-gitlab api I've added configurable `max_retries`, `retry_transient_errors` and `backoff_factor` for the legacy client.

Also added 408, 425, 429 error code too the list of http_status that need retry if `retry_transient_errors` is set to True (default behavior)

### DIFF
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -58,7 +58,7 @@ gitlab:
   #
   # Some parameters are also passed to our custom GitLab core API client (for requests not yet migrated to python-gitlab):
   # max_retries: 3        # number of times to retry failed REST API requests
-  # retry_transient_errors: false # (defaults to `true`) retry requests that fail due to transient errors (408, 425, 429, 5xx)
+  # retry_transient_errors: false # (defaults to `true`) retry requests that fail due to transient errors (429, 5xx)
   # backoff_factor: 0.25 # factor for exponential backoff between retries
 
 # Configuration to apply to GitLab projects, groups and subgroups


gitlabform/gitlab/core.py
```diff
@@ -33,10 +33,14 @@ def __init__(self, config_path=None, config_string=None):
...
+        retries_status_forcelist = []
+        if self.gitlab_config.get("retry_transient_errors", True):
+            retries_status_forcelist = [408, 425, 429, 500, 502, 503, 504] + list(range(520, 531))
+
         retries = Retry(
-            total=3,
-            backoff_factor=0.25,
-            status_forcelist=[500, 502, 503, 504] + list(range(520, 531)),
+            total=self.gitlab_config.get("max_retries", 3),
+            backoff_factor=gitlab_config_from_file.get("backoff_factor", 0.25),
```
amimas: Can you think of any reason why `total` and `backoff_factor` is not in the `default_gitlab_config`? It seems right now we have 2 different place where default value is configured. If there isn't a valid reason, I think we should set all defaults in one place.

Also, both should retrieve value from `gitlab_config` to be consistent, I think.

gitlabform/gitlab/core.py
```diff
@@ -33,10 +33,14 @@ def __init__(self, config_path=None, config_string=None):
 
         self.session = requests.Session()
 
+        retries_status_forcelist = []
+        if self.gitlab_config.get("retry_transient_errors", True):
+            retries_status_forcelist = [408, 425, 429, 500, 502, 503, 504] + list(range(520, 531))
```
amimas: Looking at existing code, it seems error codes 4xx wasn't included in the retries. I'm looking at following docs from gitlab and don't see error codes `408` and `425`.

https://docs.gitlab.com/api/rest/troubleshooting/#status-codes

I can see error code `429` to be in this list since this refers to rate limit. But in that case, we should perform retry based on `Retry_After` header sent by GitLab's API. Doesn't make sense to blindly retry using backoff when we know we've hit rate limits.

Not sure if GitLab's API response actually includes the above mentioned header or a different header. Apparently this is what python-gitlab does as mentioned in #1025. Should confirm this but I couldn't find any reference in GitLab's docs on what their API response contains when clients receive `429 too many requests` as response.

To keep things simple, maybe let's use same retry status list as before. At least this PR allows configurability for users. If you can, open a separate PR that includes retry enhancement for error code 429.

What do you think?

docs/reference/index.md
amimas, line 1: Just a thought on this doc - not related to this PR. Personally I don't like how the sample config code block is becoming a giant documentation of various options :). I feel this page needs a bit of enhancement, but that's a separate discussion/PR.

amimas (commented):
Thank you for the PR and trying to complete #1025 as much as possible. Left a few comments.

Are you able to check if there's any unit test for the logic you're updating? Assuming it's not, how easy/difficult would it be to add some test?

gitlabform/gitlab/core.py
thomascizeron: Totally right on `408` and `425`, I did'nt check Gitlab return codes and assumed those were transient errors

Yes I think that handling the `Retry_After` logic would be cleaner for `429` but instead of re implementing it would it be simpler to just migrate the needed API calls to `python-gitlab` ?

I agree that at least this PR provides configurability, I can add a comment to explain that 429 should be handled with the `Retry_After` logic.

docs/reference/index.md
thomascizeron: Yes I agree, I made a proposition in this PR https://github.com/gitlabform/gitlabform/pull/1156 it is draft as it is based on this branch.

gitlabform/gitlab/core.py
amimas: That's my thought as well. Given the amount of resource we have and our future goal is to migrate fully to `python-gitlab`, it's probably best to focus on that effort. A lot of processors have already been migrated and new processors are being implemented using `python-gitlab`.

docs/reference/index.md
amimas: You're very kind. Thanks for all your time and energy lately for gitlabform. Let's close this topic here for now.

docs/reference/index.md
```diff
@@ -55,6 +55,11 @@ gitlab:
   # GraphQL API parameters (passed to python-gitlab's GraphQL client):
   # max_retries: 10       # number of times to retry failed GraphQL requests
   # obey_rate_limit: true # automatically wait and retry if rate limit is hit
+  #
+  # Some parameters are also passed to our custom GitLab core API client (for requests not yet migrated to python-gitlab):
+  # max_retries: 3        # number of times to retry failed REST API requests
+  # retry_transient_errors: false # (defaults to `true`) retry requests that fail due to transient errors (408, 425, 429, 5xx)
```
amimas: In latest code, `408` and `425` is not included. Let's update the doc for accuracy.

gitlabform/gitlab/core.py
amimas: Thank you for addressing this.

docs/reference/index.md
amimas:
```suggestion
  # retry_transient_errors: false # (defaults to `true`) retry requests that fail due to transient errors (429, 5xx)
```

PR #1155, opened 2026-01-05
refactor: Move username/group transformations from config to processors

thomascizeron:
Removed redundant `UserTransformer` and `GroupTransformer` classes, moving username→user_id and groupname→group_id transformations directly into individual processors for better error handling and architectural consistency.

**Changes:**
- Removed `UserTransformer` and `GroupTransformer` from configuration transformers
- Added transformation logic to `ProtectedEnvironmentsProcessor` and `MergeRequestsApprovalRules` processors
- Updated error messages to provide context about which configuration section failed
- Deleted 3 test files (test_user_transformer.py, test_group_transformer.py, replaced test_user_and_group_transformers.py with initialization test)


gitlabform/configuration/transform.py
```diff
@@ -72,7 +72,7 @@ def convert_to_simple_types(configuration: Configuration):
 
 class UserTransformer(ConfigurationTransformer):
```
amimas: I haven't actually explored this area of the project properly yet. It seems in this class, we have logic to transform username to user id. This is being done for configs under `protected_environment.*.deploy_access_levels.user` and `merge_requests_approval_rules.*.users`.

But, the project supports using username instead of user id in other/all areas of gitlabform config. So, not sure why this class is dealing with only the above 2 areas. Wondering if this class should be removed entirely and update whatever else is handling transformation of username to user id.

Do you have time to investigate this?

thomascizeron: Yes it seems redundant as others processors handle it without using `UserTransformer`, I can look to remove it and use the same pattern everywhere

thomascizeron: Note: I tried something more like it and edited the title and description of PR

amimas: Thanks for looking into it. I'm actually thinking we should centralize this logic. The main logic of transforming username to user id or group name to group id will be same in every processor. It doesn't make sense to build this logic everywhere. The existing setup probably wasn't best either.

Thinking out loud here... GitLabForm generates an "effective configuration" and then it's passed on to the processors. At least I believe that's what is happening. I feel the generating of effective configuration should take care of these transformations so that processors don't need to manage it. Not sure if I'm missing an issue with that approach. Maybe in some cases the transformation shouldn't happen?

Anyways, in general I'm a little on the fence right now about the changes. For example: couple of items below are about supporting username natively using GitLab's API instead of transforming to user id.

- https://github.com/gitlabform/gitlabform/issues/791
- https://github.com/gitlabform/gitlabform/pull/1017


@TimKnight01  - Would be good to brainstorm this together. Don't want to rush making changes in this area but definitely should look into this, in my opinion.

TimKnight-Opencast: @amimas @thomascizeron - Thanks both. I think for the sake of consistency our end, we should always transform to user_id, so we don't have to think "Does this Gitlab API support Username?"

As for centralised translation, I guess Integration Tests would cover this, if it's already being done up front, or in Abstract Classes. One place which does user_name to Id transformation is the ideal in terms of maintainability. It depends on whether the "older" non-python gitlab mappers can share that logic with the newer ones?

amimas: Thanks Tim. I agree that if we always transform to user/group id, we don't need to worry about managing individual processors.

If I'm not mistaken, I think we are in agreement that ideally the logic should be centralized. I think that's what the existing code tried to do but don't think it was maintaineable. That might be why some of the other processors has the logic repeated.

I'm not sure yet what the ideal solution will look like. There are a few things unknown to me. @thomascizeron  - are you able to continue investigate this? If not, that's totally fine. I understand this will require time to investigate. Maybe we'll come back to this later.

thomascizeron: I'll try to take the time to find a good solution, don't know when but I agree that doing the conversion at one point only is the more consistent

thomascizeron: I think the simplest solution to have the transformation done at one point only is to have a global ID transformer that performs before processors and swap all users/groups to their ids for a fixed list of paths in the config that need the mapping.

it can be tedious to maintain the list of path that need the covnersion but it is simpler that maintaining all the individual conversions in the processors logic.

I can probably draft a PR with this implemented for reference.

thomascizeron: I was thinking something like this: https://github.com/gitlabform/gitlabform/pull/1185

Some acceptance tests don't pass, if the general idea is validated I will try to make it work

I've started from this branch but as the refactoring was growing I felt a new PR/git history was necessary

amimas: Thank you @thomascizeron for looking into this. What's the next step?

Should we close this PR and re-open #1185? That PR was closed when the `v5` branch was merged to `main` and deleted.

I briefly skimmed through the change. I think overall it's headed in the right direction, but haven't spent enough time with it yet. @TimKnight01 - what's your thought?

TimKnight01: Agreed - I've got a bit of time sitll @thomascizeron @amimas so I can pick up from #1185 against new `main`

amimas:
Closing this in favor of #1250

PR #1169, opened 2026-01-30
test: add python 3.14 to test and fix snyk issues in tests.dockerfile

TimKnight01:


amimas:
@TimKnight01 - Is there any value in running tests using both Python versions 3.12 and 3.14?

I'm just thinking of tests failing on v3.14 but passing on v3.12. That can prevent PRs from merging even though release happens using v3.12. It's probably not likely to happen but not sure we need both. I was thinking we upgrade to Python v3.14 as part of the v5 release.

What do you think?

TimKnight01:
@amimas agreed - unfortunately GitHub in our setup, unlike GitLab won't run changes to actions until they're merged into Main, so I have no way of knowing 3.14 works til we merge it in. Once it (hopefully) just goes green we can update to use 3.14 as the sole version, and maybe keep smoke tests only on 3.12 too

.github/workflows/tests-standard.yml
amimas: Not sure if it was intentional. Any reason not including 3.14 for the security test?

TimKnight01:
@amimas my initial thought was "We've never before run multiple versions on the Ultimate/Premium", but I can add them now if you'd like 👍

PR #1173, opened 2026-02-08
docs: Add instructions for running gitlabform with locally trusted certificates

### LINKED ISSUE #1135: Unable to get past certificate verification for locally trusted CA certificates
**Describe the bug**

Refer to the following screenshot:

Image

I have configured a custom CA certificate to be trusted by my system (as shown in the curl command's output in the screenshot), however GitLabForm still reports error regarding certificate verification failure.

**GitLabForm version**

```text
🏗 GitLabForm version: 4.6.1 = the latest stable 😊
```

**GitLab version**

17.11-ce

brlin-tw:
This pull request adds instructions to allow making use of the system trust store for HTTPS verification.

<details>
<summary>Original content of this PR</summary>
By default the Requests library uses certificates from the certifi package and is unable to honour the certificates trusted by the system.  This patch modifies the behavior to use the CA bundle file from the OS whenever possible.

Fixes #1135.

This patch is the result of the Amazon Q AI agent session([transcript](https://hackmd.io/7u9b8yzASROgMG4aYc9pdA?view)), I did:

* Tested that the patch works
* Ran the unit tests
* Ran `mypy .`
* Ran `black .`

but please review.

## References

* [SSLContext.set_default_verify_paths() — ssl — TLS/SSL wrapper for socket objects — Python 3.14.3 documentation](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_default_verify_paths)
* [CA Certificates — Advanced Usage — Requests 2.32.5 documentation](https://requests.readthedocs.io/en/latest/user/advanced/#ca-certificates)
</details>

### DIFF
--- a/tests/acceptance/standard/test_files.py
+++ b/tests/acceptance/standard/test_files.py
@@ -328,11 +328,6 @@ class TestFiles:
         assert push_access_user_ids == []
         assert merge_access_user_ids == []
 
-        # In previous test, we set unprotect_access_level to MAINTAINER, so it should remain unchanged after the file update
-        # But, this assertion is only valid for GitLab EE, as unprotect_access_level is not available in GitLab CE
-        if is_enterprise_edition:
-            assert unprotect_access_level is AccessLevel.MAINTAINER.value
-
     def test__set_file_with_chinese_characters(self, project, branch):
         set_file_chinese_characters = f"""
         projects_and_groups:


brlin-tw:
Revision 515c5f9 is entirely generated by the AI agent as I don't have the expertise to do it properly, please review.

brlin-tw:
@TimKnight01 

> the formatting job is failing, so can you fix that first please?

I've fixed the formatting.

> What is the specific problem you were finding that has prompted this change?

GitLabForm won't trust the CA configured in my system's trust store, see #1135.

gitlabform/gitlab/core.py
```diff
@@ -54,7 +56,12 @@ def __init__(self, config_path=None, config_string=None):
         self.session.mount("https://", HTTPAdapter(max_retries=retries))
 
         self.session.verify = self.gitlab_config["ssl_verify"]
-        if not self.gitlab_config["ssl_verify"]:
+        if self.gitlab_config["ssl_verify"] is True:
+            # Use system CA bundle instead of certifi
```
TimKnight01: In which case @brlin-tw I think I would like an option in the config to allow users to opt-in to using system CA bundle. `"ssl_system_ca"` true/false -> defaulting to False

amimas: Hi @TimKnight01  - what would the config look like? Is it something like below?

```yaml
gitlab:
  url: https://gitlab.yourcompany.com
  token: "<private token OR an OAuth2 access token>"
  ssl_verify: true
  ssl_system_ca: true
```

Not sure why we need a separate config/option. Any reason why the existing option (`ssl_verify`) won't handle local cert bundle the way it's proposed so far?

TimKnight01: @amimas 
Previously if you did ssl_verify: true, it would do no special sauce

Now it faffs around with the Root file path to get CA Certs, which maybe a red-line for some security reasons, it certainly makes me uncomfortable as the default, I'd be concerned this could open up attack-vectors like injecting a CA Cert somehow.

If it's under it's own toggle you have to opt-in to using GLF to find Root CA Certs rather than letting python figure out Certs at a language level.

amimas: Thanks for explaining. I understand where you're coming from. What makes sense to you in terms of how it should ideally work? Ignoring this change for a moment, would you expect a separate option when you want to enable SSL verification? I didn't expect that. I guess to me the config `ssl_verify` is a global/generic option - doesn't really say whether it's through `certifi` or locally trusted certs.

Your reasoning is good though. We can/should probably consider this as a breaking change. But, would you still expect locally trusted CA to be used using a separate config option?

Thoughts @TimKnight01 ?

TimKnight-Opencast: @amimas to be honest I don't really think GitLabForm should be doing anything with Certs or Certifi explicitly. The issue trying to be resolved here feels like something that could be fixed in the Runner or by building a Docker Container from gitlab form and doing the certs in there.

Software shouldn't mess around with CaCerts thats an environment thing

TimKnight-Opencast: (aaand I forgot to change my GH account to my personal one again when replying here 😂 )

amimas: Hi @TimKnight01 - Sorry I'm still not able to follow what you're saying. GitLabForm is simply performing SSL verification via the `ssl_verify` flag so that all communications with GitLab API is secured. You're not saying GitLabForm should not do this, right?

I think the question you're raising is whether GitLabForm should use/rely on certs that are trusted locally at the OS level. Is that right? If yes, could you please share any some reference or industry standards? I'm not a security expert, so want to understand this a bit better.

Let's take `curl` as an example. Take a look at the following documentation.

https://curl.se/docs/sslcerts.html

> Use the native store
> In several environments, in particular on Microsoft and Apple operating systems, you can ask curl to use the system's native CA store when verifying the certificate. Depending on how curl was built, this may already be the default.

TimKnight01: @amimas I'll have to dig into standards and look at say cURLs codebase.

I think it is fine for GLF to say "Is SSL Enabled?" when the verify flag is on

I don't feel comfortable with GitLabForm having hard-coded routes to look at the OS level Certificate directory.

For me personally if my CA Cert was causing an issue connecting to an instance I would do something like
```dockerfile
from gitlabform

#Setup my CA Certs as default on this image

#Setup Python to be using those Certs - therefore GLF will use them when it calls via the relevant Python back-end APIs
```

amimas: Yeah, let us know what you find out. It's good to be cautious.

To me, I don't have any concern with the current change. I also don't think GitLabForm is hard coding any routes. It's using python's built-in `ssl` module to find the default cert on the OS.

Having said that, I just realized we haven't talk about python-gitlab 😃 . The change proposed in this PR is only for legacy API client created by GitLabForm. Since we're migrating away from it and moving towards `python-gitlab` library, this change won't actually help. Currently `python-gitlab` also relies on `certifi` package just like we do. For system CA store, their recommendation is to set the environment variable `REQUESTS_CA_BUNDLE` with path to the system's CA.

https://python-gitlab.readthedocs.io/en/stable/api-usage-advanced.html#ssl-certificate-verification

I think we should go that route too. In that case, no change should be required. Just documenting how to run gitlabform by setting the above mentioned environment variable.

@brlin-tw  - could you please try the above mentioned steps and share your findings? It should work for you given you've set that environment variable correctly.

@TimKnight01 - FYI.

TimKnight01: `REQUESTS_CA_BUNDLE` ah yes! that's what I was getting towards and have seen it done before.

amimas: Sounds good.

@brlin-tw  - Please try the above suggestions. After you try that out, it'd be good if you can convert this PR to simply document the steps you needed to use system stored CA. Thank you!

amimas: Thanks for the discussion @TimKnight01 .

amimas: @brlin-tw  - Did you get a chance to try the suggestion above? Would love to get the documentation updated with those steps.

brlin-tw: @amimas

> could you please try the above mentioned steps and share your findings? It should work for you given you've set that environment variable correctly.

Apologies for the ignorance, I can verify that the SSL verification works properly after specifying the `REQUESTS_CA_BUNDLE` environment variable.

> convert this PR to simply document the steps you needed to use system stored CA

Where in the documentation would you like to have the steps documented?

(Specifically, which section in [🏃 Running - GitLabForm](https://gitlabform.github.io/gitlabform/running/)?)

brlin-tw: I've updated the pull request, please review.

docs/running.md
amimas, line 1: Could you please add a reference to this section about how to run with custom CA cert from the following config documentation?

https://github.com/gitlabform/gitlabform/blob/main/docs/reference/index.md#optional-parameters

In the above section, it lists `ssl_verify` config and a link back to this would be useful, I think.

gitlabform/gitlab/core.py
amimas: Thank you for testing and confirming @brlin-tw . Really appreciate it. I've updated the PR title and also left a comment on the latest change.

Please take a look and also update the PR description so that it's relevant to the current change being proposed.

docs/running.md
amimas: Hi @brlin-tw  - will you have time to look into this in the next few days? Would love to get those merged.

brlin-tw: @amimas 

Apologies for the ignorance, I've applied the requested change.

amimas: Thanks for looking into this @brlin-tw . Update looks good.

PR #1177, opened 2026-02-09
feat: upgrade to python 3.14

TimKnight01:


amimas (commented):
I think project.toml also has config for python version. That should be updated too, right? Any other files need to be updated? Can you please do a search for 3.12?

PR #1185, opened 2026-02-16
refactor: Centralize user/group id fetching in a transformer that uses python-gitlab

thomascizeron:
Centralize principal name→ID resolution in `PrincipalIdsTransformer`.

Based on the [following discussion](https://github.com/gitlabform/gitlabform/pull/1155#discussion_r2739557366) 

- Move all username/groupname lookups into the transformer and add dict-key handling for `members`/`group_members`.
- Remove duplicate runtime lookups from processors so they consume pre-resolved `*_id`/`*_ids`.
- Tests: Add unit tests for the new transformer

### DIFF
--- a/gitlabform/configuration/transform.py
+++ b/gitlabform/configuration/transform.py
@@ -1,11 +1,10 @@
-from logging import debug
+from logging import debug, warning, error
 from abc import ABC, abstractmethod
 from ez_yaml import ez_yaml
 from ruamel.yaml import YAML
 from types import SimpleNamespace
 from gitlab import GitlabGetError
 
-from cli_ui import fatal, warning
 from ruamel.yaml.comments import CommentedMap
 from yamlpath import Processor
 from yamlpath.exceptions import YAMLPathException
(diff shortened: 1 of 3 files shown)


gitlabform/configuration/transform.py
```diff
@@ -257,3 +181,258 @@ def _do_transform(self, configuration: Configuration):
...
+        if user_id is None:
+            raise GitlabGetError(
+                f"No users found when searching for username '{username}' while transforming path '{path}'",
+                404,
+            )
+        return user_id
+
+    def _get_group_id(self, groupname: str, path: str) -> int:
+        try:
+            return self.gitlab.get_group_id(groupname)
```
TimKnight01, line 348: Is this cached using lruCache like get_user_id_cached is?

gitlabform/processors/project/members_processor.py
```diff
@@ -1,6 +1,6 @@
-from cli_ui import debug as verbose, warning, info, error
+from cli_ui import debug as verbose, info, error
```
TimKnight01: Can you replace cli_ui with `logging` here please - I recall we decided to move to logging for consistency

gitlabform/processors/project/tags_processor.py
```diff
@@ -1,5 +1,5 @@
 from logging import debug
-from cli_ui import warning, fatal, error
+from cli_ui import warning, fatal
```
TimKnight01: as before, replace cli_ui with logging

TimKnight01 (changes requested):
Basic concept looks good @thomascizeron - would definetly want to see transform.py covered in good unit tests, as the code, whilst readable could be hard to maintain without unit tests acting as Documentation 👍

gitlabform/configuration/transform.py
thomascizeron: yes it is also cached, it calls get_group_by_path_cached that uses lru cache decorators

PR #1239, opened 2026-04-08
fix(acceptance): use safe_load in test running

EhteshamSid:
`yaml.load()` in `tests/acceptance/standard/test_running.py` doesn't pass a safe Loader. This can deserialize arbitrary Python objects and is an RCE risk if the YAML comes from user input or the network. Switched to `yaml.safe_load()`.

### DIFF
--- a/tests/acceptance/standard/test_running.py
+++ b/tests/acceptance/standard/test_running.py
@@ -42,7 +42,7 @@ class TestRunning:
 
         path = Path("output.yml")
         yaml = YAML(typ="safe")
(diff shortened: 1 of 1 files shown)


EhteshamSid:
Closing - the patch is incorrect. `yaml` here is a `ruamel.yaml.YAML(typ="safe")` instance, which does not have a `.safe_load()` method. The original `yaml.load(path)` was already safe since `typ="safe"` uses the safe loader. Apologies for the noise.

PR #1240, opened 2026-04-10
feat: add support for not including projects that are scheduled for deletion

rickbrouwer:
Add support for not including projects that are scheduled for deletion by using `--include-projects-scheduled-for-deletion` with as default `false`.

### Checklist

- [X] Tests have been added

Fixes #1216


gitlabform/lists/projects.py
```diff
@@ -116,20 +131,25 @@ def _get_projects(self, target: str, groups: Groups) -> Projects:
...
+                    scheduled_for_deletion.append(project_object["path_with_namespace"])
             except NotFoundException:
                 debug("Could not find '%s'", project)
                 if project_transfer_source := self._get_project_transfer_source_from_config(project):
                     source_project = self._find_project_transfer_source_in_gitlab(project_transfer_source)
                     if source_project:
                         if source_project["archived"]:
                             archived.append(project)
+                        if source_project.get("marked_for_deletion_on"):
+                            scheduled_for_deletion.append(project)
```
amimas: I realize we're following what was done for archived projects. In this case, should we actually check additionally whether the flag is enabled before adding the projects to corresponding lists?

Checking the flag is being done in the `_get_all_and_omitted_projects_from_groups`.

rickbrouwer: Yeah, that is possible. I will also apply the same fix for archived to keep the behavior consistent

rickbrouwer: Done

TimKnight01 (approved):

TimKnight01 (approved):

TimKnight01:
Thanks @rickbrouwer - releasing in 5.3.0

PR #1241, opened 2026-04-13
test: restore Premium & Ultimate test functionality in CI

TimKnight01:
- requires_license fixture will skip tests if license is expired anyway, as this could be hiding real errors
- fail acceptance tests if license has expired, this should give us quick feedback when we need to request a new one for tests in CI

### DIFF
--- a/.github/workflows/tests-standard.yml
+++ b/.github/workflows/tests-standard.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Start GitLab in Docker
         run: ./dev/run_gitlab_in_docker.sh
       - name: Run standard acceptance tests against Enterprise Edition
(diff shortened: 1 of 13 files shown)


TimKnight01:
I **suspect** this PR will fail for both Ultimate and Premium tests, I have already requested new Licenses from GitLab so when I get those we can get this Green and gain confidence in Test Suite

amimas (approved):
Thanks for looking into it @TimKnight01 . I agree we should enforce ultimate tests instead of silently skipping them. Otherwise we might end up with changes being merged that introduces issues for ultimate features.

Are the premium tests being enforced?

TimKnight01:
@amimas they are enforce, but if you check the Logs of the Action run, they all show as _Skipped_ which is interesting to me

TimKnight01:
I spotted this when looking at the logs on the cli_ui refactor and saw Premium all marked as skipped

PR #1247, opened 2026-04-15
feat (build): migrate gitlabform to newer dev tools and add a dev toolkit

amimas:
This PR migrates the project to be managed using `uv`. This is a newer tool that simplifies several aspects of project management. Additional features/changes have been introduced but they are all tied to development/packaging of gitlabform. Details are described below.

## Current state

- Project uses `setuptools` for building the python package
- There are no deterministic/reproducible build, which could be different  between CI vs local development
- Dependencies are not consistently managed. For example: the project uses tools like `black`, `bandit`, etc. in CI, but they are not available locally unless developers/contributors manually installs them
- Some dependencies (i.e. docs) are not pinned to any versions
- Development process requires running various different types commands - some are python commands (create and activate virtual env), ensure relevant python version is available, some are shell scripts, etc.

## Updates/changes being introduced

- Migrate to `uv` - some highlights of this are:
    - Install/update python version that the project requires (using `.python-version` file). No need to use `pyenv` or similar tools
    - Deterministic build by providing `uv.lock` file containing version details of all dependencies, including transitive dependencies
    - Automatically manage python virtual environment - no need to manually create a virtual environment and activate/deactivate it
 - Migrate to `hatchling`
     - Use `hatchling` to build the python package - no need for `setuptools`. This setup is common when using `uv`
- Pin all dependencies in `pyproject.toml`, including tools used for docs, linting, testing, etc. 
- GitLabForm development toolkit
     - A custom python CLI that uses `uv` under the hood when needed, but provides a consistent development experience. For example:
         - `uv run workspace setup` will prepare the dev environment using `uv` (install python env if needed, setup virtual environment, install/update dependencies in the virtual environment, configure pre-commit hook in the git repo, etc.)
         - `uv run qa lint` will run all the linters in parallel
         - more similar commands, which is also discoverable easily by running `uv run dev` or `uv run dev --help`. Usage for each command is also available by using the `--help` flag
    - Migrate pre-commit hook to `prek` tool
        - Also configured pre-commit hook to use the locally installed version of the tool (installed through `uv`) so that we have a single source of truth for tools and their versions (`pyproject.toml`)
    - Provide capability to locally build **and** verify python packages or docker image
- CI updates
    - Kept the CI update as minimal as possible. Since package manager is being switched in this PR, some updates were needed
    - Additional improvements to CI can be done in separate PR

closes #1217


amimas:
@TimKnight01  - Could you please review this PR?

Currently the CI is failing because it's not running the workflows from my branch. We will need a way to manually trigger the workflows from my branch.

For the review, I suggest starting with the `local_development.md` first and try it out or test the steps locally. Otherwise, following the order of files in this PR in GitHub might be difficult to connect the dots when reviewing.

Looking forward to your feedback.

TimKnight01:
@amimas - will review - we could temporarily add `pull_request` trigger back to the workflow on main 

Basically `pull_request_target` trigger let's Forks use our github-secrets for running Acceptance Tests with the downside that they **always** run the workflow defined on `main` meaning it's then really hard to test Workflow changes in advance when making changes to the CI, I might see if there's a way we can configure the CI to run `pull_request` trigger for the root repo, so if we want to make changes to CI, Maintainers raise branches in the root repo rather than on Forks

TimKnight01:
@amimas Figured out a way to get branches cut from the main repo to trigger CI defined on themselves using `pull_request` workflow trigger  -> this branch is rebased with that and now correctly trying to use `uv` to install packages in CI

And on my PR from a Fork -> [here](https://github.com/gitlabform/gitlabform/pull/1250), it uses the pull_request_target trigger so that it can access the Secret for Acceptance Testing

.github/workflows/prs.yml
```diff
@@ -61,16 +61,15 @@ jobs:
...
+        run: pip install uv==0.11.2
+      - name: Run acceptance tests that require premium license
         env:
           GITLAB_EE_LICENSE: ${{ secrets.GITLAB_EE_LICENSE }}
-        run: ./dev/run_gitlab_in_docker.sh
-      - name: Run acceptance tests that require premium license
-        run: pytest --cov=. --cov-report=xml --durations=0 --reruns 3 --reruns-delay 10 --log-cli-level=WARNING tests/acceptance/premium
+        run: |
+          uv sync --frozen --no-dev --group test
+          uv run --no-sync gitlab-local up
```
TimKnight01: Can we retain "Start in Docker" as a separate step? @amimas

.python-version
```diff
@@ -0,0 +1 @@
+3.12
```
TimKnight01, line 1: 3.14 ?

TimKnight01 (commented):
Lots to review so I'll take a full look when the Pipeline is green, but I really love the idea of the Development Kit and being able to use `uv` to run all the commands such as standing up Docker

.github/workflows/prs.yml
amimas: Yup. Makes sense. Pushed an update separating the logics into different steps.

.python-version
amimas: Should it be set to 3.14? I wasn't sure since we're still testing 3.12 in some portions of CI. I'm happy to move to 3.14 and address 3.12 related issue later if needed. What do you think?

.github/workflows/_releases.yml
```diff
@@ -86,18 +89,28 @@ jobs:
...
+        run: |
+          uv run --no-sync package build
+          uv run --no-sync package verify
+      - name: Publish to PyPI
         run: |
-          pip install wheel build
-          python -m build
-      - name: Publish distribution 📦 to PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1
+          uv run --no-sync package publish
```
amimas, line 103: I have not tested this step. Let me know if you have any suggestions for how to test it. Didn't dig into it yet.

.python-version
TimKnight01: @amimas - I _think_ we should develop in 3.14 but still in CI/Build test in 3.12, now if that is not how .python-version works with UV and it'd break our 3.12 let's leave as is 👍

.github/workflows/_releases.yml
TimKnight01: @amimas `releases` runs when _any_ tag in format v*.*.* is pushed, so you could push v0.0.50~alpha or some completely unused thing, meaning downstream consumers won't accidentally pick it up when it pushes out to pypi.

When we're happy with it we can probably delete it out of the various registries. But I think that's the best way to test in isolation without breaking our ability to release other versions?

.github/workflows/_releases.yml
```diff
@@ -39,21 +39,24 @@ jobs:
       - tests-standard
       - tests-premium
     runs-on: ubuntu-latest
+    env:
+      UV_PYTHON: 3.14
```
TimKnight01: Can we set a global environment variable for these rather than per-job?

.python-version
TimKnight01: I _think_ we can achieve what I'm asking by having the .python-version file being the version we're developing in e.g 3.14

Actually when I look at the CI logs I can see it setting up python 3.14 venv, so .python-version should just control local dev?

TimKnight01:
@amimas been trying out the local_dev instructions off the branch and it's really clean! Thank you for the work in this one, it makes everything so much more consistent.

TimKnight01:
CLI output from `gitlab-local up` needs updating to reflect`uv` usage:
image

TimKnight01:
Not sure if this is just a github runner thing, but comparing `uv` to previous way of standing up tests, it's about 4-5m quicker using `uv` which if it is consistently that will be a nice improvement on run time

.python-version
amimas: Yup. We can make version "3.14" default for development in `.python-version`. Right now I have set `UV_PYTHON` environment variable in the CI where we want to test different python versions, which `uv` will use over `.python-version` file.

I'll set the default to "3.14" soon.

.github/workflows/_releases.yml
amimas: Thanks for that suggestion. I like that approach. Will test it out and post here.

.github/workflows/_releases.yml
amimas: Pushed an update. I've setup a "composite action" that can be used across all the jobs. It contains the uv installation logic and also sets up necessary environment variable for any subsequent steps.

.github/workflows/_releases.yml
amimas: @TimKnight01  - I just tried that approach. Pushed an update to the release workflow with some tweaks I needed to make sure the workflow runs. You can see the workflow run here:

https://github.com/gitlabform/gitlabform/actions/runs/24642006237

The above was run when I pushed a tag named [v0.0.540a1](https://github.com/gitlabform/gitlabform/tree/refs/tags/v0.0.540a1). And you can find this [here in pypi](https://pypi.org/project/gitlabform/0.0.540a1/)

Some observation  about how it worked: In my first attempt, I had set the version in `pyproject.toml` file as `0.0.540-alpha` (https://github.com/gitlabform/gitlabform/blob/v0.0.540-alpha-1/pyproject.toml#L13). When the release workflow executed `uv run --no-sync package publish`, it actually published it as [0.0.540a0](https://pypi.org/project/gitlabform/0.0.540a0/). Apparently that is the versioning convention according to [PEP440](https://peps.python.org/pep-0440/#pre-releases) for pre-release versioning.

We should follow that versioning convention when we manually set the version number during release. Good thing is `uv` is automatically applying the convention, I guess but it can be confusing if `pyproject.toml` doesn't match with final release number. I plan to automate this in the next round of updates so that we don't need to manually enter a version number for cutting a release - let the tool automatically set version number based on conventional commit convention since last release tag.

Please review and let me know if you see anything that needs updates for this PR.

mfriedenhagen:
Hi, just one remark: if you create dummy releases, some people get an email 😊. I firstly was afraid something bad happens within the project, maybe add a release comment like "fooling around while introducing uv". I appreciate the move to a proper build tool 👍

.github/workflows/_releases.yml
TimKnight01: @amimas sounds good, happy to align with PEP440 pre-release versioning

TimKnight01:
@amimas this looks great, I am happy!

Going to merge in and then rebase my open PR onto it, so we can validate `main` still works and everything

amimas:
Thank you for the shout @mfriedenhagen . Will be cleaning up the dummy/test releases soon but that's a good call out. Might have to look at using a separate test repo for these kind of tests. If you have any other suggestions, please let us know.

PR #1250, opened 2026-04-16
refactor: Centralize user/group id fetching in a transformer that uses python-gitlab

TimKnight01:
Centralize principal name→ID resolution in PrincipalIdsTransformer.

Based on the https://github.com/gitlabform/gitlabform/pull/1155#discussion_r2739557366

- Move all username/groupname lookups into the transformer and add dict-key handling for members/group_members.
- Remove duplicate runtime lookups from processors so they consume pre-resolved *_id/*_ids.
- Tests: Add unit tests for the new transformer


TimKnight01:
Follows on from: https://github.com/gitlabform/gitlabform/pull/1155 and discussions therein

gitlabform/configuration/transform.py
```diff
@@ -257,3 +179,183 @@ def _do_transform(self, configuration: Configuration):
...
+            "projects_and_groups.*.tags.*.allowed_to_create.*.user",
+            "projects_and_groups.*.branches.*.allowed_to_push.*.user",
+            "projects_and_groups.*.branches.*.allowed_to_merge.*.user",
+            "projects_and_groups.*.branches.*.allowed_to_unprotect.*.user",
+        ],
+        "dict_keys": [
+            "projects_and_groups.*.members.users.*",
+            "projects_and_groups.*.group_members.users.*",
+        ],
+    }
```
amimas: I haven't had a chance to think about an alternative yet. So, don't have any suggestions.

Ideally we don't want to maintain hard-coded paths like this. GitLabForm follows "raw parameter passing" design so that it's not tied with specific release/version of GitLab's API. That way users can start to apply new/updated attributes without needed a release from GitLabForm.

I'm thinking of following same pattern. Can we detect user name or group name anywhere within the config and simply apply the transformation? Any downside to this approach?

I think we can safely assume the names to be available anywhere within the sections (i.e. `project_and_groups.*.*`)? In this example: anywhere within `branches`, `members`, `schedules`, `tags`, etc, assuming it will detect keys from all the nested levels from there.

amimas (commented):
Didn't get a chance to fully review it. Just one one comment so far if you could please investigate.

gitlabform/processors/group/group_members_processor.py
amimas, line 59: When the processors run, the user/group names from the config should already be transformed to IDs, right? In that case, is this logic still needed? Looks like here we're still trying to do things based on username. Let me know if I misunderstood.

amimas, line 164: Same comment as above about having logics based on username.

gitlabform/processors/group/group_members_processor.py
```diff
@@ -191,11 +192,8 @@ def _process_users(
...
-                    user_id = self.gl.get_user_id_cached(user)
-                    if user_id is None:
-                        message = f"Could not find User '{user}' on the Instance"
-                        error(message)
-                        raise GitlabGetError(message, 404)
+                    # user_id is pre-resolved by PrincipalIdsTransformer
+                    user_id = users_to_set_by_username[user]["user_id"]
 
                     if common_username in users_before:
                         group_member: GroupMember = group.members.get(user_id)
```
amimas, line 199: Same comment as above. Do we still need `common_username`?

gitlabform/processors/project/members_processor.py
amimas, line 113: Similar comment as above. Since the config will have transformed the config already when the processor starts, do we still need this logic?

amimas (commented):
Some comments based on my understanding that processors execute after gitlabform generates the effective config, which would include username/group name transformed to IDs. In that case, we shouldn't need any checks/logic in the processors, right?

Also, wouldn't the branch protection processor also be refactorable now? I recall there are some logic in the branches processor where usernames are being trasnformed.

Another thing we can do to keep this PR smaller and easier to review is leave the processors as-is and update them separately in future PRs.

Let me know what you think.

TimKnight01:
@amimas thanks for the review, I'll go through and see what can be applied 👍

TimKnight01:
I agree with keeping other Processors out of this PR and raising an issue to tidy up "user logic"

gitlabform/configuration/transform.py
TimKnight01: @amimas pushed an update for this

gitlabform/processors/group/group_members_processor.py
TimKnight01: @amimas, working through the logic; for `dict` keys like:
```
 users:
     timknight01:
            access_level: {AccessLevel.OWNER.value}
```

The transformation _doesn't_ currently replace the dictname `timknight01` with a user_id, but it does appenf the user_id as a key under the dict 

I'll do some investigations later about changing this around, because I'd like to never have usernames in the config post-transformation as that's clearer.

TimKnight01: In `_transform_dict_keys_to_ids` in `transform.py` we'd need to mutate the raw Config to swap the `name` to `id` 🤔

gitlabform/configuration/transform.py
```diff
@@ -257,3 +179,164 @@ def _do_transform(self, configuration: Configuration):
...
+    # Factorized mapping path lists to keep transformation paths in one place
+    group_mapping_paths = {
+        "values": [
+            "projects_and_groups.*.**.group",
+            "projects_and_groups.**.groups",
+        ],
+        "dict_keys": [
+            "projects_and_groups.**.groups.*",
+        ],
+    }
```
amimas, line 201: This seems to be unused variable. Let's remove it.

gitlabform/configuration/transform.py
```diff
@@ -257,3 +179,164 @@ def _do_transform(self, configuration: Configuration):
...
+                if node_coordinate.parentref != from_key:
+                    continue
+
+                parent = node_coordinate.parent
+                node = node_coordinate.node
+
+                # if it's a dict_key, do nothing here, we'll transform it in _transform_dict_keys_to_ids
+                # TODO: REFACTOR?
+                if isinstance(node, CommentedMap):
+                    continue
```
amimas, line 275: What refactor is this referring to?

gitlabform/configuration/transform.py
```diff
@@ -257,3 +179,164 @@ def _do_transform(self, configuration: Configuration):
...
+            path="projects_and_groups.*.**.users.*",
+            id_key="user_id",
+            lookup=self._get_user_id,
+        )
+
+    @staticmethod
+    def _dedupe(items: list[int]) -> list[int]:
+        return list(dict.fromkeys(items))
+
+    def _transform_principal_to_ids(self, processor, path: str, from_key: str, to_key: str, get_id_from_name_function):
```
amimas, line 263: Let's set the type for the callable function.


```suggestion
    def _transform_principal_to_ids(self, processor, path: str, from_key: str, to_key: str, get_id_from_name_function: Callable[[str, str], int]):
```

gitlabform/configuration/transform.py
```diff
@@ -257,3 +179,164 @@ def _do_transform(self, configuration: Configuration):
...
+                node = node_coordinate.node
+                if not isinstance(node, (dict, CommentedMap)):
+                    continue
+                if id_key in node:
+                    continue
+                key = str(node_coordinate.parentref)
+                node[id_key] = lookup(key, path)
+        except YAMLPathException as e:
+            debug(f"YAMLPathException while transforming dict keys to IDs for path '{path}': {e}")
+            pass
```
amimas, line 324: Could we add a bit more documentation for this? I'm not sure if it's clear what is being injected. Maybe we give a input config and what this method will transform that into?

gitlabform/configuration/transform.py
```diff
@@ -257,3 +179,164 @@ def _do_transform(self, configuration: Configuration):
...
+
+    def _get_user_id(self, username: str, path: str) -> int:
+        user_id = self.gitlab.get_user_id_cached(username)
+        if user_id is None:
+            raise GitlabGetError(
+                f"No users found when searching for username '{username}' while transforming path '{path}'",
+                404,
+            )
+        return user_id
```
amimas, line 333: Any reason the `get_user_id_cached` method doesn't handle raising exception? Just wondering whether we need to define yet another method that is supposed to return user id, which already exists. I imagine the exception would be applicable in all scenarios? But if a processor or other logic should capture the exception and handle it differently, it should do that in there. Does it make sense?

gitlabform/configuration/transform.py
```diff
@@ -257,3 +179,164 @@ def _do_transform(self, configuration: Configuration):
...
+
+    def _get_group_id(self, groupname: str, path: str) -> int:
+        try:
+            return self.gitlab.get_group_id(groupname)
+        except GitlabGetError as error:
+            raise GitlabGetError(
+                f"No groups found when searching for group '{groupname}' while transforming path '{path}'",
+                404,
+            ) from error
```
amimas, line 342: Similar comment as above about possibly avoid defining another method for retrieving group id.

tests/acceptance/standard/test_group_members_users.py
```diff
@@ -246,14 +246,15 @@ def test__add_user_with_access_level_name(self, group, users, one_owner_and_two_
...
-                access_level: developer
+              users:
+                {users[0].username}:
+                  access_level: owner
+                {users[1].username}:
+                  access_level: developer
+                {users[2].username}:
+                  access_level: developer
+                {user_to_add}: # new user 1
+                  access_level: developer
```
amimas, line 257: It seems the test config is currently incorrect and you're fixing it now. In that case, how is this test case passing? Is it validating properly or is there a hidden bug?

tests/unit/configuration/transform/test_implicit_name_transformer.py
```diff
@@ -48,3 +48,52 @@ def test__transform_for_protected_environments_sanity_check():
...
+    """Test ImplicitNameTransformer skips non-dict/CommentedMap nodes."""
+    config_yaml = """
+    projects_and_groups:
+      "foo/bar":
+        protected_environments:
+          production:
+            deploy_access_levels:
+              - access_level: maintainer
+          staging: some_string_value
+          testing: 12345
```
amimas, line 63: Could you clarify what exactly we're testing here? For a `protected_environment` section, I think this is invalid config because a dictionary would be expected. I don't think we're testing the protected environment processor here. So, not really sure what is being tested.

tests/unit/configuration/transform/test_principal_ids_transformer.py
```diff
@@ -0,0 +1,257 @@
...
+    config_yaml = """
+    projects_and_groups:
+      "my-group":
+        group_members:
+          users:
+            user1:
+              access_level: 50
+            user2:
+              access_level: 30
+    """
```
amimas, line 196: Not sure if I missed it... is there a test case where the config provided already in the format of what would've been injected? I think that would test whether gitlabform supports usage of ID instead of name, which is a helper from gitlabform. Or, maybe that's a test for processors 🤔 ?

gitlabform/configuration/transform.py
amimas: Thanks for the refactor @TimKnight01 . I think that looks good.

Going to close this thread so that we can focus on other threads or open new ones with appropriate context.

tests/acceptance/standard/test_group_members_users.py
TimKnight01: @amimas I think there was a hidden bug in our previous transformation logic and processing of group_members, where it didn't correctly look up the `users` header in the config.

tests/unit/configuration/transform/test_implicit_name_transformer.py
TimKnight01: @amimas to be honest I do need to go back through these, I just grabbed them from the previous PR and should improve the test naming

tests/unit/configuration/transform/test_principal_ids_transformer.py
TimKnight01: e.g. providing 
``` 
users:
    123:
        access_level: 50
```

TimKnight01: According to the `Group Members` .md file:

* `users` - to add single users. Here the key names are usernames and the values are as described in the [add member to a group or project endpoint of the Group and project members API](https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project),

gitlabform/configuration/transform.py
TimKnight01: @amimas done - I'll have another try of refactoring this so that it turns
```
   group_members:
           users:
              some-user:
                 access_level: 40
```
into
```
   group_members:
           users:
              123:
                 access_level: 40
```

TimKnight01: As I think that'll allow us to remove more of the Name -> Id logic out of the other Processors

tests/unit/configuration/transform/test_implicit_name_transformer.py
TimKnight01: updated

tests/acceptance/standard/test_group_members_users.py
amimas: Thanks @TimKnight01 . Have you figured out yet what that existing bug is? Wondering if this will resolve an existing open issue. I would've expected an issue to be raised already.

gitlabform/configuration/transform.py
```diff
@@ -257,3 +181,184 @@ def _do_transform(self, configuration: Configuration):
...
+            to_key="group_ids",
+            get_id_from_name_function=self.gitlab.get_group_id,
+        )
+
+        self._transform_dict_keys_to_ids(
+            processor,
+            path="projects_and_groups.*.**.groups.*",
+            id_key="group_id",
+            lookup=self.gitlab.get_group_id,
+        )
```
amimas, line 264: For user id lookup, the `get_user_id_cached` function is called. Any reason the cached function is not called for looking up the group ID? I see `self.gitlab.get_group_id` being called here.

gitlabform/configuration/transform.py
amimas, line 1: @TimKnight01  - Been a little busy lately and haven't been able to review it clearly, especially since I'm not fully familiar with this part of the code base. Used an AI review and one item that stood out was following.

```md
Performance (Multiple Passes): Each transformer (Users, Groups, Access Levels) performs multiple passes over the entire configuration tree using `processor.get_nodes`.

Feedback: For very large GitLabForm configurations, this might introduce a slight delay. While likely acceptable for a CLI tool, a more optimized approach would be a single walk of the tree that identifies keys to be transformed based on a registry of paths.
```

Not sure how accurate this feedback is. Do you have time to look into this?

amimas (commented):
Would it be possible to exclude the changes made in the `processors/` directory or are they required for this pass of the PR?

Also, is it correct to assume the original config (i.e.`user`) will be available to the processors and the transform will add the `user_id`?

tests/acceptance/standard/test_group_members_users.py
TimKnight01: I don't see any issue, it's a hidden bug whereby something works which shouldn't

Currently I could do
```
group_members:
  user-name:
     access-level: developer
```

but it _should_ be
```
group_members:
 users:
     user-name:
        access-level: developer
```

Otherwise the declarative YAML is unclear and GLF is assuming the entries are users

gitlabform/configuration/transform.py
TimKnight01: @amimas as a heads up I've gone round and round on this solution of trying to update the keys in this scenario.

I can get the YAML updated using YAMLPath set_value but it ends up wrapping the new value as a String rather than a Dict, whatever I do, which would massively impact downstream parsing. So I think the current implementation where we add an "id" field is the only viable solution

e.g.
```
group_members:
           users:
              some-user:
                 access_level: 40
```

becomes
```
group_members:
           users:
              some-user:
                 access_level: 40
                 user_id: 123
```

The conversion is done up front we just have to access it differently in later processors - relating to your comments about logic still existing 

I am going to put a pin in this, if you'd like to create a separate issue to look at it later - let's do that, but right now trying to mutate this bit of YAML is too much

gitlabform/configuration/transform.py
TimKnight01: @amimas `get_group_id` is using a cached call, the function just hasn't been remaned.

One for a separate tidy up, that is just renaming two functions?
image

gitlabform/configuration/transform.py
TimKnight01: @amimas yeah I've also been busy - will think about how we could make an optimisation pass 👍

TimKnight01: ah it's not actually going over the whole tree each time, because it's using regex matching on `get_nodes`, so will only each be going over a subset of the tree

It's _possible_ we could refactor it to just trawl the tree from top to bottom in one go, but I feel like this might make the code more complex - and if we hear of performance concerns from people we can revisit?

TimKnight01:
@amimas I am going to draw us back to the Problem Statement the original refactor I jumped off of was trying to address. As I feel this is quite a big change, and I'm unsure how much actual benefits it'll bring

**Current situation**
A person may declare users and groups by name or id in declarative YAML
Our Processors must make an API call to get the full entity from Gitlab (whose API supports search by name or id) and grab it's ID off there
This could cause Latency by adding multiple redundant API calls for the same entities - however in all implementations of get_group_id or get_user_id, we implement an lru_cache

**Problem**
Our code is a bit messy because each processor must understand that a group or user could be an id or name in the YAML
We have a transformation process which runs over the YAML document prior to processing, so could turn all names into ids up front via API calls

**Refactor**
It is hard to mutate the YAML document upfront to ensure every "name" entry is now an Id
Potentially we open a can of worms by allowing the code to mutate the raw YAML that a user provided us with
It does not necessarily remove the need for Processors to understand that in some cases a user or group will still be referred to by name, in fact it's potentially more confusing as you need to know that it's only a "name" if it's a Key e.g.
```
users:
  - user1:
        - some-field: true
        - user-id: 123
```


I don't think it will meaningfully reduce the amount of API calls made in total, or significantly reduces code-complexity, and it adds overhead to refactor all existing Processors (including those not already using python-gitlab implementations) to pull the data correctly. And my gut feeling is that this refactor has grown enough complexity that it could cause knock-on impacts.

I'd therefore suggest we may want to consider dropping this PR all together - pulling out some of the Documentation improvements and Integration Test fixes (especially for that potential hidden bug) into their own PRs/Issues

amimas:
Hey @TimKnight01  - Thanks!! Sometimes it's good to go back to the drawing board. I do agree with you that it's getting a little complicated. I like your suggestions for putting this aside for now.

I agree with you that the performance gain will likely be minimal since we already have lru_cache for various functions including retrieval of group/user id.  I feel some refactoring might still be good in terms of reducing code/logic duplications within the processors - but this PR has been a good learning experience 😅 . Maybe we can revisit this later.

Thinking out loud here: I think we probably don't need to do the transformation of "dict" - I'm thinking the transformation probably should be limited to something like `user: jdoe` to `user_id: 123` and make it additive config rather than replacing.

Please close this PR if there aren't anything else we need to discuss. Might be good to spend our limited time/effort into other areas for now.

TimKnight01:
@amimas agreed - there are nicer ways to achieve what's being intended here. I'll close it and raise an issue to look at de-duplicating/tidying up the get_group and get_user bits

PR #1251, opened 2026-04-19
feat: add support for group level branch protection settings

### LINKED ISSUE #1049: Does gitlabform support setting group-level protected branches?
gitlabform supports project-level protected branches but I couldn't work out how to configure group-level protected branches. Is this supported?
It looks like gitlabform supports the group settings [documented here](https://docs.gitlab.com/user/group/), but the group-level protected branches is at [separate API endpoint](https://docs.gitlab.com/api/group_protected_branches/).

rickbrouwer:
Add support for group level branch protection settings.

### Checklist

- [X] Tests have been added

Fixes #1049


gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,161 @@
...
+
+        for branch in sorted(configuration["group_branches"]):
+            branch_configuration: dict = self.convert_user_and_group_names_to_ids(
+                configuration["group_branches"][branch]
+            )
+
+            self.process_branch_protection(gitlab_group, branch, branch_configuration)
+
+    def convert_user_and_group_names_to_ids(self, branch_config: dict) -> dict:
+        debug("Transforming User and Group names in group branch configuration to Ids")
```
amimas: This shouldn't be needed once #1250 is merged. But even without that, does group level branch protection allow configuring users or groups?

GitLab's API doc seems conflicting. I see there are attributes for creating the branch protection with attributes for `allowed_to_merge`, `allowed_to_push`, `allowed_to_unprotect`, where user/group can be specified.

https://docs.gitlab.com/api/group_protected_branches/#protect-repository-branches

But the API doc also calls out that specifying users/groups is not supported.
Have you had a chance to try out the API and see what exactly works?

Image

gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,161 @@
...
+
+class GroupBranchesProcessor(AbstractProcessor):
+
+    def __init__(self, gitlab: GitLab, strict: bool):
+        super().__init__("group_branches", gitlab)
+        self.strict = strict
+
+        self.custom_diff_analyzers["merge_access_levels"] = BranchesProcessor.naive_access_level_diff_analyzer
+        self.custom_diff_analyzers["push_access_levels"] = BranchesProcessor.naive_access_level_diff_analyzer
+        self.custom_diff_analyzers["unprotect_access_levels"] = BranchesProcessor.naive_access_level_diff_analyzer
```
amimas: The `BranchesProcessor` here is really from the project level branch processor. Using logic from project's branch processor here feels.... probably not the right approach.

If this is a common logic, I think they should be extracted into a separate util and then used where needed.

amimas (commented):
Thank you for the PR and looking into this. Left couple of comments/observations.

I see project branch protection logic/class is being reused/referenced here. Those logics have evolved with GitLab's API. It feels a bit weird to call that class here. Maybe I'm wrong. Haven't really thought about it. But I'm mostly wondering how this will be maintained when group vs project API for branch protection starts to diverge in terms of how they work. Maybe that's a future gitlab issue, if any?

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: Yeah, you are right. I just checked our Gitlab and indeed, I can't select users or groups there (weird, but oh well...). I removed that part. With that, I also immediately get rid of that transforming "code" (and I don't have to wait for #1250 either)

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: Done. I have extracted the shared logic.

docs/reference/protected_branches.md
```diff
@@ -2,7 +2,42 @@
...
+
+    This section requires GitLab Premium (paid). (This is a GitLab's limitation, not GitLabForm's.)
+
+!!! warning
+
+    Protected branch settings for groups are restricted to top-level groups only.
+
+This section purpose is to manage the [group-level protected branches](https://docs.gitlab.com/ee/api/group_protected_branches.html).
+
+Note: group protected branches only support access levels. Individual users and groups cannot be specified (this is a GitLab API limitation).
```
amimas: Suggesting minor tweak and highlight GitLab API limitation.


```suggestion
This section purpose is to manage the [group-level protected branches](https://docs.gitlab.com/ee/api/group_protected_branches.html).

!!! info

    This section requires GitLab Premium (paid). (This is a GitLab's limitation, not GitLabForm's.)

!!! warning

    Protected branch settings for groups are restricted to top-level groups only. (This is a GitLab's limitation, not GitLabForm's.)

    Group protected branches only support access levels. Individual users and groups cannot be specified (this is a GitLab API limitation).
```

docs/reference/protected_branches.md
```diff
@@ -2,7 +2,42 @@
...
+
+The supported keys are: `protected`, `push_access_level`, `merge_access_level`, `unprotect_access_level`, `allow_force_push`, and `code_owner_approval_required`.
```
amimas: I suggest we remove this or rephrase it differently so that we're not explicitly listing supported keys.

GitLabForm's users are expected to be familiar with GitLab's API to know what attributes are supported. Since one of the architecture/design of GitLabForm is "raw parameter passing", any attributes supported by GitLab's API in future will be supported automatically, assuming the processor is implemented correctly and no unexpected surprises from GitLab's API :)

Hope that makes sense.

gitlabform/processors/group/group_branches_processor.py
amimas: Thank you for checking @rickbrouwer .  Did you happen to try if GitLab's API accepts users/groups? I'm just curious if this is actually supported (mistakenly) but not exposed in the UI. I doubt that's the case, though.

gitlabform/processors/group/group_branches_processor.py
amimas: Thank you so much for the refactor. Really appreciate it.

amimas:
@rickbrouwer - Thank you so much for looking into the feedbacks so far. Need some time to review the code and have been a bit busy with work lately. Appreciate your patience.

rickbrouwer:
No problem at all. I will also review and process your feedback later. For me, that is mainly in the evenings and on weekends.

Regarding your question:

> Did you happen to try if GitLab's API accepts users/groups?

Not yet; I might just check if I can do that. I don't run an Enterprise version myself. But where I work, we do. Previously, with my other PRs (like the avatar PR), I first built my own version and tested it against our Enterprise version. Later, I wasn't able to do this anymore; let's just say, in summary, that has something to do with policies on our end ;)). However, I do want to spend some time looking into what is possible in that regard.
It remains interesting, whether that might be possible via the API; after all, via the GUI you first see a very short spinner, making it seem as if it can't load something that might actually be possible. Who knows.

So, I'm going to check that and get back to you later. I will leave that request as a comment on unresolved for now so that I can post my answer there later.

Further; in conclusion, because it is worth saying, I greatly appreciate your reviews. Reviewing takes a tremendous amount of time; I know that too. You examine everything thoroughly and think carefully about where things can be improved. That makes your reviews valuable, and your contributions make this project of higher quality. Thank you for that.

docs/reference/protected_branches.md
rickbrouwer: yeah, I will remove that

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: In any case, it does not work with this current code:
```
[15:18:02] WARNING  Connected as non-admin. You may encounter         core.py:93
                    permission issues.                                          
           CRITICAL Configuration string '<redacted>' is not one  transform.py:221
                    of the valid access levels: no access,                      
                    minimal, guest, planner, reporter,                          
                    developer, maintainer, owner, admin   
```

rickbrouwer:
but it doesn't seem to be going well at all anyway.

If I use this config:

```
---
config_version: 4

gitlab:
  url: https://gitlab
  ssl_verify: true

projects_and_groups:
  my-first-group/*:
    group_branches:
      main:
        protected: true
        push_access_level: no access
        merge_access_level: maintainer
```

Then it seems like he wants to add it to all subgroups as well, which logically doesn't work...

```                              
@ ( 1/17) Processing group: my-first-group
@ ( 2/17) Processing group: my-first-group/my-subgroup-1
[15:16:52] ERROR    Protecting branch 'main' at   group_branches_processor.py:117
                    group level failed! Error                                   
                    '['cannot be associated with                                
                    a subgroup']                                                
@ ( 3/17) Processing group: my-first-group/my-subgroup-2
           ERROR    Protecting branch 'main' at   group_branches_processor.py:117
                    group level failed! Error                                   
                    '['cannot be associated with                                
                    a subgroup']                    
```

/edit:

Fixed this.

gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,142 @@
...
+
+            self.process_branch_protection(gitlab_group, branch, branch_configuration)
+
+    def process_branch_protection(self, group: Group, branch_name: str, branch_config: dict):
+        protected_branch: Optional[GroupProtectedBranch] = None
+
+        try:
+            protected_branch = group.protectedbranches.get(branch_name)
+        except GitlabGetError:
+            debug(f"The branch '{branch_name}' is not protected at group level!")
```
amimas: In project branch protection processor, this is logged at `info` level. Let's keep it consistent.

gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,142 @@
...
+            special_list_keys = [
+                "merge_access_levels",
+                "push_access_levels",
+                "unprotect_access_levels",
+            ]
+            for key, value in transformed_branch_config.items():
+                if key not in special_list_keys:
+                    existing_value = getattr(protected_branch, key, None)
+                    if existing_value != value:
+                        debug(f"Creating data to update {key} as necessary")
```
amimas: Same comment as above.

gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,142 @@
...
+                        protected_branch_api_patch_data[key] = value
+
+            debug("Creating data to update merge_access_levels as necessary")
```
amimas: Same comment as above.

gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,142 @@
...
+            merge_access_items_patch_data = BranchProtection.build_patch_request_data(
+                transformed_access_levels=transformed_branch_config.get("merge_access_levels"),
+                existing_records=tuple(self._get_list_attribute(protected_branch, "merge_access_levels")),
+            )
+            if len(merge_access_items_patch_data) > 0:
+                protected_branch_api_patch_data["allowed_to_merge"] = merge_access_items_patch_data
+
+            debug("Creating data to update push_access_levels as necessary")
```
amimas: Same comment as above.

gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,142 @@
...
+            push_access_items_patch_data = BranchProtection.build_patch_request_data(
+                transformed_access_levels=transformed_branch_config.get("push_access_levels"),
+                existing_records=tuple(self._get_list_attribute(protected_branch, "push_access_levels")),
+            )
+            if len(push_access_items_patch_data) > 0:
+                protected_branch_api_patch_data["allowed_to_push"] = push_access_items_patch_data
+
+            debug("Creating data to update unprotect_access_levels as necessary")
```
amimas: Same comment as above.

gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,142 @@
...
+            else:
+                warning(message)
+
+    @staticmethod
+    def _get_list_attribute(protected_branch: GroupProtectedBranch, attribute_name: str) -> list[Any]:
+        existing_list_value: list[Any] = []
+        existing_attr = protected_branch.attributes.get(attribute_name)
+        if existing_attr is not None:
+            existing_list_value = existing_attr
+        return existing_list_value
```
amimas: Could this be part of the `branch_protection` utility? The logic in both project vs group is same. Might have to play with type definition for `protected_branch` parameter a little since that's the only difference stood out to me.

gitlabform/processors/project/branches_processor.py
```diff
@@ -265,229 +266,15 @@ def convert_user_and_group_names_to_ids(self, branch_config: dict):
...
-                        mapped_list.append(mapped_item)
-                    new_branch_config[target_key] = mapped_list
-                    new_branch_config.pop(key)
-
-        # this key is not present in
-        # protected_branch.attributes, so _needs_update() would always
-        # return True with this key present.
-        new_branch_config.pop("protected")
-        return new_branch_config
+        return BranchProtection.map_config_to_protected_branch_get_data(our_branch_config)
```
amimas: I might be missing something. Why do we need this static method in this processor's class? Couldn't the lines where this is being called directly call the method from `BranchProtection`?

amimas (commented):
Thanks @rickbrouwer for the updates. Haven't looked at the tests yet but overall logic looks good. Some minor comments about log level and other clarifications.

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: yes, that would be possible

gitlabform/processors/project/branches_processor.py
rickbrouwer: Yeah, true. This could be called in the BranchProtection directly. Will change that.

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: updated

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: updated

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: updated

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: updated

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: updated

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: updated

gitlabform/processors/project/branches_processor.py
rickbrouwer: updated

tests/acceptance/premium/test_group_protected_branches.py
```diff
@@ -0,0 +1,116 @@
...
+                protected: true
+                push_access_level: {AccessLevel.NO_ACCESS.value}
+                merge_access_level: {AccessLevel.MAINTAINER.value}
+        """
+
+        run_gitlabform(config, group)
+        run_gitlabform(config, group)
+
+        protected_branch = group.protectedbranches.get("main")
+        assert protected_branch is not None
```
amimas, line 148: I feel this doesn't necessarily test what is intended. Maybe after the first run, validate expected result and then do the same after the second run? Is there another way idempotency can be tested?

tests/acceptance/premium/test_group_protected_branches.py
amimas, line 1: Can we add couple more test scenarios?

- what happens if the config includes `allowed_to_*` type attributes since they are in GitLab's documentation but apparently not supported?
- what happens if subgroups are included since only top-level group is supported by GitLab's API?

Ideally in both case there'll be warnings but not crash. Or, let me know if expectation is different.

gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,136 @@
...
+            if branch_config.get("protected") is None:
+                critical(f"The Protected key is mandatory in group_branches configuration, fix {branch} YAML config")
+                sys.exit(EXIT_INVALID_INPUT)
+
+        return True
+
+    def _process_configuration(self, group: str, configuration: dict):
+        if "/" in group:
+            debug(f"Skipping group_branches for '{group}' - only supported for top-level groups")
+            return
```
amimas: I'm thinking we move this logic to the beginning in `_can_proceed` function. That way it might exit early, although I haven't checked the flow, but that seems to be the right home for this. What do you think?

docs/reference/protected_branches.md
```diff
@@ -2,7 +2,40 @@
...
+
+This section purpose is to manage the [group-level protected branches](https://docs.gitlab.com/ee/api/group_protected_branches.html).
+
+!!! info
+
+    This section requires GitLab Premium (paid). (This is a GitLab limitation, not GitLabForm's.)
+
+!!! warning
+
+    Protected branch settings for groups are restricted to top-level groups only. (This is a GitLab limitation, not GitLabForm's.)
```
amimas: I have been thinking about this. I think we might need a separate section for this topic. Here's my thought:

Long before GitLab added the capability to configure branch protection at the group level, GitLabForm's config syntax (i.e. `my-group/*` pattern) allowed managing  branch protection for all projects under that group from a single config. So, if we look at "project level branch" protection, but configured/managed in one place like group level setting, this is already achievable without group level branch protection and there are no limits in terms of top-level group only.

I hope I was able to explain what I mean above. Definitely needs to be explained clearly, but I'm wondering maybe this should be clearly explained in the doc.

What do you think?

rickbrouwer: I understand what you mean. We need to explain the difference clearly, and also indicate that branch protection on projects is possible in all variants of GitLab.

tests/acceptance/premium/test_group_protected_branches.py
rickbrouwer: Done

tests/acceptance/premium/test_group_protected_branches.py
rickbrouwer: Done

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: Done

docs/reference/protected_branches.md
rickbrouwer: Added extra explanation at top-level

rickbrouwer: Do you want to leave the "old" text where it is, so including the top-level text? Or is the top-level text sufficient, and can I use that to simplify the text below? Feel free to offer code suggestions if you want to remove something or change the text.

TimKnight01: @rickbrouwer - sorry if this is explained elsewhere; I'm just working my way down from the top of the PR.

What happens if I mix patterns in my GitLabForm config e.g.

```
top-group:
   group_branches:
       main:
           protected: true
           push_access_level: maintainer
top-group/sub-group/*:
  branches:
      main:
           protected: false
           push_access_level: maintainer
```
or:
```
top-group:
   group_branches:
       main:
           protected: true
           push_access_level: maintainer
top-group/sub-group/*:
  branches:
      main:
           protected: true
           push_access_level: no access
```

gitlabform/processors/group/group_branches_processor.py
```diff
@@ -0,0 +1,148 @@
...
+            branch_config = configuration["group_branches"][branch]
+            if branch_config.get("protected") is None:
+                critical(f"The Protected key is mandatory in group_branches configuration, fix {branch} YAML config")
+                sys.exit(EXIT_INVALID_INPUT)
+
+            for key in ("allowed_to_push", "allowed_to_merge", "allowed_to_unprotect"):
+                if key in branch_config:
+                    for item in branch_config[key]:
+                        if isinstance(item, dict) and (
+                            "user" in item or "user_id" in item or "group" in item or "group_id" in item
```
TimKnight01: can we extract this into a well named function, something like `branch_protection_config_contains_user_or_group`?

TimKnight01 (commented):
only one additional comment from me @rickbrouwer 

Looks good, thanks for the work refactoring this and adding new functionality

rickbrouwer:
> Looks good, thanks for the work refactoring this and adding new functionality

Thanks! Your feedback is greatly appreciated. Regarding your question about mixing config, I will build this PR on our end to run a test against a GitLab version to test this.

rickbrouwer:
Okay, @TimKnight01, because of your comment about mixing, I overlooked something fundamental.

I made a build and I see that it wants to apply the settings to the subgroups, but never to the top-level.
That isn't possible either, because get_groups() only looked for keys ending in /*.

That means an adjustment needs to be made in 'gitlabform/configuration/groups.py' to allow top-level support.

I have made that adjustment and will add it to a separate commit shortly so we can take a closer look at whether we want to do this.

I will also test this extensively myself later, so I need to take a good look at everything first.

gitlabform/processors/group/group_branches_processor.py
rickbrouwer: yes, done

TimKnight01:
Sounds good @rickbrouwer I'll wait out for a re-review request 👍

rickbrouwer:
I'm a bit stuck on the naming (`group_branches`). Is `group_protected_branches` better?

There is 'initial default branch protection', in addition to 'protected branches', both at the group level.
I will continue developing for a little while longer, but I will occasionally throw my notes in here as well so that the community/maintainers can contribute their thoughts regarding preferences.
I occasionally commit in the branch with separate commits for traceability. I also create my own builds to double-test things.

Screenshot:

image

docs/reference/protected_branches.md
rickbrouwer: The project will then have protection with a lock on it:

image

rickbrouwer:
@TimKnight01 

Back in review 🙂 
Switched the config back to the standard `group_1/*:` syntax. The processor's `_can_proceed` enforces top-level-only. Also renamed the section to `group_protected_branches`. The renaming is to avoid confusion with the (possible future) default_branch_protection group setting (I haven't figured out where this exists within the API.). Should there be a different opinion on that, renaming is done in little time..

TimKnight01:
@rickbrouwer I agree with that naming scheme, it makes sense and aligns with the GitLab naming scheme

docs/reference/protected_branches.md
TimKnight01: @rickbrouwer cool - that's kind of what I expected to happen, top-level takes priority, preventing users only with access to sub-groups overriding some security practice enforced at the top level.

Could you add a note to the .md file that this is expected behaviour please?

TimKnight01:
Looks good to me - if you could make the final change to the MD file, then I'm happy to approve and merge, thanks!

docs/reference/protected_branches.md
rickbrouwer: Well... from another look, you can override it.

I think Gitlab is saying something like: "Your own (project) setting isn't necessary, because it already exists at the group level... but if you want something different yourself, that's fine."

Do you think that's worth documenting, or is it more something where if a user sets one at the group level ánd one at the project level, that person figures out for themselves how GitLab handles it?

image

TimKnight01: Ah got you

I feel like it's worth a quick note on how GitLab functions, we do that elsewhere in the Docs and link out to GitLab documentation where we can, just to help our end-users navigate the settings based on our learnings. As often GL don't make it clear themselves!

rickbrouwer: added a note https://github.com/gitlabform/gitlabform/pull/1251/changes/a8e0c8b36e7fc06e0658ab82d47e75cb221c6d10

TimKnight01 (approved):

TimKnight01:
@amimas @rickbrouwer - Approved, will leave open for a day or so in case you have any comments / further changes you realise need to be made :+1:

rickbrouwer:
Thanks! And no hurry! 🙂

rickbrouwer:
Thank you all for the good feedback, reviews, and collaboration.

TimKnight01:
Thanks for the contribution @rickbrouwer - appreciate the commentary and explanation around it, made it easy to unpick 👍

PR #1282, opened 2026-05-11
fix: allow editing and creation of files on wildcard protected branches 

### LINKED ISSUE #1274: Wildcard Protected Branches causing 400: You can only create or edit files when you are on a branch
I'm seeing the error **400: You can only create or edit files when you are on a branch** whenever I try to enforce the existence and contents of a file on protected branches while also using wildcards in protected branch paths.

For example, with this configuration:
```
---
config_version: 4

gitlab:
  url: https://my.gitlab.instance.example

protected branch: &protected_branch
  allow_force_push: false
  code_owner_approval_required: true
  merge_access_level: developer
  protected: true
  push_access_level: maintainer

code settings: &code_owner_settings
  "CODEOWNERS":
    branches: protected
    commit_message: "GitlabForm: updating CODEOWNERS file"
    content: |
      [SomeGroup] @someuser
      .gitlab-ci.yml
      CODEOWNERS
    overwrite: true
    skip_ci: true
    template: false

projects_and_groups:
  mygroup/myproject:
    branches:
      main: *protected_branch
      bugfix/*: *protected_branch
    files: *code_owner_settings
```
    
I'm observing the error **400: You can only create or edit files when you are on a branch**

If I omit the `bugfix/*` line (and there is no wildcard protected branch configured in `mygroup/myproject`, e.g. due to prior run with config above), then there is no error.

Searching online (and querying Gemini, fwiw) I could not find helpful information. Searching through the issues on github.com/gitlabform/gitlabform/issues, I keep ending up at https://github.com/gitlabform/gitlabform/issues/999 but I have no idea whether that issue report really is related.

I'm running with GitlabForm v5.3.0 and GitLab v18.11.

TimKnight01:
- resolves #1274
- where user has specified a ProtectedBranch rule using wildcard, we find all matching branches with that rule and iterate through them to add/edit Files according to the config

### DIFF
--- a/gitlabform/processors/project/files_processor.py
+++ b/gitlabform/processors/project/files_processor.py
@@ -4,11 +4,10 @@ import sys
 from pathlib import Path
 
 from logging import debug, info, warning, critical
-from typing import List, Optional
 
 from jinja2 import Environment, FileSystemLoader
 from gitlab import GitlabGetError, GitlabUpdateError
-from gitlab.v4.objects import Project, ProjectFile, ProjectBranch, ProjectProtectedBranch
+from gitlab.v4.objects import Project, ProjectFile, ProjectBranch
 from gitlab.base import RESTObject
 
 from gitlabform.constants import EXIT_INVALID_INPUT, EXIT_PROCESSING_ERROR
@@ -35,34 +34,20 @@ class FilesProcessor(AbstractProcessor):
                 continue
 
             config_target_ref = configuration["files"][file]["branches"]
-            branches_to_update: List[ProjectBranch | ProjectProtectedBranch] = []
+            branches_to_update: list[ProjectBranch] = []
 
-            if isinstance(config_target_ref, str):
-                # Target ref could be either 'all' or 'protected'.
-                # Get a list of branches that should be updated.
-                if config_target_ref == "all":
-                    branches_to_update.extend(project.branches.list(get_all=True, lazy=True))
-                elif config_target_ref == "protected":
-                    branches_to_update.extend(project.protectedbranches.list(get_all=True, lazy=True))
+            if config_target_ref == "all":
+                # Get all Branches in the Project
+                branches_to_update.extend(project.branches.list(get_all=True, lazy=True))
+            elif config_target_ref == "protected":
+                # Get all Protected Branch definitions -> this could return wildcard branch names
+                protected_branches = project.protectedbranches.list(get_all=True, lazy=True)
+                for protected_branch in protected_branches:
+                    self.append_branches_matching_ref(branches_to_update, file, project, protected_branch.name)
             elif isinstance(config_target_ref, list):
-                # Get a list of branches from the config that should be updated.
+                # List of specific branch refs to update -> this could contain wildcard branch names
                 for ref in config_target_ref:
-                    try:
-                        if "*" in ref:
-                            wildcard_branches = project.branches.list(get_all=True, regex=ref)
-                            for wildcard_branch in wildcard_branches:
-                                project_branch = project.branches.get(wildcard_branch.get_id())
-                                branches_to_update.append(project_branch)
-                        else:
-                            branches_to_update.append(project.branches.get(ref))
-
-                    except GitlabGetError:
-                        message = f"! Branch '{ref}' not found, not processing file '{file}' in it"
-                        if self.strict:
-                            critical(message)
-                            sys.exit(EXIT_INVALID_INPUT)
-                        else:
-                            warning(message)
+                    self.append_branches_matching_ref(branches_to_update, file, project, ref)
 
             debug(
                 "File '%s' to be updated in '%s' branche(s)",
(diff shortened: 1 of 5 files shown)


TimKnight01:
@rickbrouwer - are you able to give this a review please? I think the other maintainers are quite busy on work projects 🙂

rickbrouwer:
> @rickbrouwer - are you able to give this a review please? I think the other maintainers are quite busy on work projects 🙂

Sure. I will try to get back to you quickly.

gitlabform/processors/project/files_processor.py
```diff
@@ -64,104 +64,127 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
             )
 
             for branch in branches_to_update:
-                info(f"Processing file '{file}' in branch '{branch.name}'")
+                process_only_first: bool = configuration.get("files|" + file + "|only_first_branch", False)
+
+                branch_name: str = branch.name
+                if branch_name.endswith("/*"):
```
rickbrouwer: I think we need to support more here:

The following is in GitLab:

image

gitlabform/processors/project/files_processor.py
```diff
@@ -46,11 +46,11 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
                     branches_to_update.extend(project.protectedbranches.list(get_all=True, lazy=True))
             elif isinstance(config_target_ref, list):
                 # Get a list of branches from the config that should be updated.
-                for branch_name in config_target_ref:
+                for ref in config_target_ref:
                     try:
-                        branches_to_update.append(project.branches.get(branch_name))
+                        branches_to_update.append(project.branches.get(ref))
```
rickbrouwer: Does a wildcard like `bugfix/*` ever get past this `branches.get(ref)` call? I am a little afraid that this endpoint is an exact match, so the new wildcard handling in the loop below wouldn't trigger for list-style configs. Or am I missing something?

gitlabform/processors/project/files_processor.py
TimKnight01: Will look into better wildcard handling 👍

gitlabform/processors/project/files_processor.py
TimKnight01: @rickbrouwer updated and test added 👍

rickbrouwer:
I see that some tests are failing and there is a coverage issue. I guess you are still working on it, but here is a little bit of input from me.

Of course, there is still some documentation to go, but I'm sure that will be fine :)

Furthermore, wildcards now get expanded in two places (the list loop and the per-branch loop). For list configs the second expansion never triggers since the names are already concrete. So, it's effectively only used for the `protected`/`all` path. Might it is cleaner to consolidate this into one spot?

And I see an `Optional` import, but I don't think that is necessary (right?)

TimKnight01:
@rickbrouwer - thanks, refactored to remove the duplication in wildcard logic, fixed the typing (and the acceptance test failure was unrelated)

rickbrouwer:
Great! Looks good.
Maybe some documentation if necessary?

In any case, I think this line needs to be adjusted:

https://github.com/gitlabform/gitlabform/blob/a42548ac9d198c72cd7fd2d4274af941d84f751d/docs/reference/files.md?plain=1#L11

TimKnight01:
@rickbrouwer thanks - I'll get onto Documentation shortly, we had a long weekend in the UK and now I'm just prepping to start new things at work. But hope to get this PR wrapped up this week 👍

rickbrouwer:
Great Tim!

I just took a quick look at the documentation for the branches API. It seems that GitLab's branches API expects a re2 regular expression (see https://docs.gitlab.com/api/branches/, at the `regex` parameter "Return list of branches with names matching a re2 regular expression"). I think we must translate the glob to an anchored regex so the documented patterns work, like `regex_ref = fnmatch.translate(ref)`, but this also needs an import (`import fnmatch`).
As a concrete example: the `*gitlab*` pattern from the docs example isn't a valid re2 expression (a quantifier at the start with nothing to repeat), so that case would fail against the API.

TimKnight-Opencast:
@rickbrouwer good catch in the documentation, I'd run into that when writing my tests but failed to check again when doing the docs

gitlabform/processors/project/files_processor.py
```diff
@@ -35,133 +35,160 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
+        try:
+            if "*" in ref:
+                # convert ref to re2 regular expression expected by Gitlab API https://docs.gitlab.com/api/branches/
+                regex_ref = fnmatch.translate(ref)
+                wildcard_branches = project.branches.list(get_all=True, regex=ref)
+                for wildcard_branch in wildcard_branches:
+                    branch_id = wildcard_branch.get_id()
+                    if branch_id:
+                        project_branch = project.branches.get(branch_id)
+                        branches_to_update.append(project_branch)
```
rickbrouwer: I think we already got the full ProjectBranch objects, so the extra branches.get(branch_id) call per branch is redundant. So, I think we could use the objects from the list directly to save a request per branch. Here is a code suggestion, but we do need to test if what I think is correct.

```suggestion
                wildcard_branches = project.branches.list(get_all=True, regex=regex_ref)
                for wildcard_branch in wildcard_branches:
                    branches_to_update.append(wildcard_branch)
```

rickbrouwer (changes requested):

gitlabform/processors/project/files_processor.py
TimKnight01: I'll debug through, if I recall list() does not return the full ProjectBranch object required for later steps

TimKnight01: Ah, I think the change is fine, it was a hangover from when the list was
project.protectedbranches.list() and I wanted to ensure branches_to_update only contained ProjectBranch objects.

rickbrouwer:
It looks like the regex_ref = fnmatch.translate(ref) line was dropped in the last commit. Is that what is meant?

TimKnight01:
@rickbrouwer - it wasn't being used and when I did implement it looked like the GL list() api rejected the requests

TimKnight01:
image
is valid re2 (the golang flavour uses google's re2) - but the list() api call blows up after converting the string using the package above :(

Not really sure why but removed because it works without, I tweaked the try-except so that we can isolate errors from List call

rickbrouwer:
I'd suggest accepting the glob syntax in the YAML (consistent with how users already write `bugfix/*` elsewhere in the same config) and converting it to RE2 internally before passing it to the API.

I'm curious how you view that and if it works out. If you need any help with that, just let me know.

TimKnight01:
@rickbrouwer - I'm a bit stuck with that suggestion been round the houses with `re`, `glob` and `fnmatch` but none of them end up constructing regexes which then function as expected with the list() endpoint

rickbrouwer:
I'd like to take a look at it too. I'll fork your branch.

rickbrouwer:
Converted wildcard patterns to RE2. So users can now write `*` wildcards. And updated the docs and tests to match. I also ran the test and it is working well now.

My changes:

https://github.com/gitlabform/gitlabform/commit/b0ba389a738c98f1ae639f23b7e64793f62eb207

I haven't looked at the unit tests and their coverage yet. If you like, I can do that as well.

TimKnight01:
Great thanks, I don't have much time today, so if you're on a roll, can you
look at the unit tests? See if there is an easy win?

TimKnight01:
Your commit looks good 👍

rickbrouwer:
And the unit tests:

https://github.com/gitlabform/gitlabform/commit/9c78dfeabcbc5318a5e5b64905f4fa05647889da

You are welcome to take over everything, I can commit it to your branch (if it is not protected), or I can offer my entire branch as a PR with a reference to this PR.

I am fine with all options.

TimKnight01:
I think you should be able to push to my branch, and then it'll show as
co-authored in the git history :)

Thanks for the help on this one

rickbrouwer:
That's perfectly fine.

Pusht my commit :)

PR #1301, opened 2026-06-02
feat: allow groups to stop section propagation to descendants

sentoz:
# Summary

Add support for parent-side inheritance blocking with `propagate: false`.

This introduces a new configuration control that lets a group or common `*`
section apply configuration at the current level while preventing that same
section from flowing further down to descendant subgroups and projects.

## What Changed

* Added internal propagation-aware configuration state with:
  * `effective_config`
  * `propagatable_config`
* Implemented `propagate: false` handling in the configuration layer for:
  * common `*` config
  * groups and subgroups
  * projects
* Preserved existing `inherit: false` behavior
* Added validation for invalid `propagate` values
* Added unit coverage for:
  * group to subgroup/project blocking
  * subgroup subtree blocking
  * local redefinition reopening propagation
  * coexistence of `inherit: false` and `propagate: false`
  * `skip: true` not reopening propagation by itself
  * invalid `propagate` values
* Added documentation for `propagate: false`
* Added acceptance scenarios for project variables propagation behavior

## Behavior

Example:

```yaml
projects_and_groups:
  team_a/*:
    variables:
      propagate: false
      team_secret:
        key: TEAM_SECRET
        value: team-value
```

With this configuration:

* `team_a` gets `variables`
* `team_a/subgroup_1` does not inherit `variables`
* `team_a/project_1` does not inherit `variables`

If a deeper level defines the same section locally, that local definition
becomes the new propagation source for its subtree unless it also sets
`propagate: false`.

## Implementation Notes

* Public configuration APIs were kept unchanged
* The propagation logic is isolated to the new configuration-state flow
* Existing inheritance-break logic remains in place and is still validated by
  the previous test suite

## Risks / Review Focus

* Review the semantics of subtree-wide blocking when a descendant redefines the
  same section locally;
* Review the interaction between `propagate: false` and existing `skip: true`
  behavior
* Review the common `*` behavior carefully, since it now acts as a propagation
  source that can also be globally blocked

### DIFF
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -162,7 +162,7 @@ projects_and_groups:
   group_1/*:
     project_settings:
       visibility: private # <-- different value!
-``` 
+```
 With this configuration, for a project `group_1/project_1` the effective configuration will be like:
 ```yaml
 project_settings:
@@ -181,7 +181,7 @@ projects_and_groups:
         key: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDB2QKx6BPzL...
         title: global_key
         can_push: false
-  
+
   group_1/*:
     deploy_keys:
       key_b: # <-- another key under deploy_keys!
@@ -279,6 +279,57 @@ deploy_keys:
 !!! important
 
     `inherit: false` can be placed at ANY place in the configuration (if it makes sense).
+
+### Stopping propagation to descendants
+
+You can also apply a section at the current level while preventing that same section from flowing further down the hierarchy by placing `propagate: false` under that section.
+
+Example:
+
+```yaml
+projects_and_groups:
+  team_a/*:
+    variables:
+      propagate: false
+      team_secret:
+        key: TEAM_SECRET
+        value: team-value
+```
+
+For the above configuration:
+
+- group `team_a` gets the `variables` section,
+- subgroup `team_a/subgroup_1` does not inherit that `variables` section,
+- project `team_a/project_1` does not inherit that `variables` section either.
+
+A deeper level can define the same section locally to start a new inheritance source for its own subtree:
+
+```yaml
+projects_and_groups:
+  team_a/*:
+    variables:
+      propagate: false
+      team_secret:
+        key: TEAM_SECRET
+        value: team-value
+
+  team_a/subgroup_1/*:
+    variables:
+      subgroup_secret:
+        key: SUBGROUP_SECRET
+        value: subgroup-value
+```
+
+With this configuration, `team_a/subgroup_1` and projects below it inherit only `subgroup_secret`.
+
+!!! important
+
+    `propagate: false` is supported in the same configuration locations where `inherit: false` is supported.
+
+Comparison:
+
+- `inherit: false` means: "this level does not inherit the section from above"
+- `propagate: false` means: "this level applies the section here, but does not pass it further down"
 
 ### Skipping sections
 
--- a/gitlabform/configuration/common.py
+++ b/gitlabform/configuration/common.py
@@ -1,6 +1,6 @@
 from abc import ABC
 
-from gitlabform.configuration.core import ConfigurationCore
+from gitlabform.configuration.core import ConfigurationCore, ConfigurationState
 
 
 class ConfigurationCommon(ConfigurationCore, ABC):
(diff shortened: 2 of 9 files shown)


sentoz:
The failing CI test does not appear to indicate a regression in the newly implemented `propagate: false` feature. Instead, the new acceptance coverage introduced for this feature exposed an older issue in project selection when GitLabForm is executed against a group with subgroup recursion enabled.

The test that reveals this problem is `TestVariablesPropagation.test__common_level_propagation_break__subgroup_local_redefinition_flows_to_projects_below` in `tests/acceptance/standard/test_project_variables.py (line 553)`. It fails at `tests/acceptance/standard/test_project_variables.py (line 580)`, where the test expects one variable to be created on a project inside a subgroup, but finds none.
The underlying issue is in `ProjectsProvider.get_projects()` in `gitlabform/lists/projects.py (line 42)`. When the target is a group and recursive subgroup traversal returns more than one group, the code falls back to `_get_single_project(target)` instead of collecting projects from the full group hierarchy. As a result, the subgroup project is never processed, GitLabForm reports `Nothing to do`., and the test fails with `assert 0 == 1`.

TimKnight01:
Hi @sentoz 

Thanks for the PR - what is the use-case you're trying to solve with this change?

I am initially not hugely in favour of a "propagate" change, it sounds contradictorary to how GLF should work. If you define something in `*` it should apply to everything. If you need some specific areas to opt out of that, then the rules should be applied at the appropriate levels.

sentoz:
> Hi @sentoz
> 
> Thanks for the PR - what is the use-case you're trying to solve with this change?
> 
> I am initially not hugely in favour of a "propagate" change, it sounds contradictorary to how GLF should work. If you define something in `*` it should apply to everything. If you need some specific areas to opt out of that, then the rules should be applied at the appropriate levels.

I'm implementing full GitLab management through GitLabForm. In my case, all access rights are assigned only through groups. Direct access remains only for the root user.

There's a requirement to revoke direct permissions for all nested projects and groups, and assign them only to root. All other permissions should be assigned only through groups. Additionally, the configuration specifies a nesting level of 3, while in GitLab itself the nesting goes up to 6–7 levels. When recursion is enabled to revoke permissions, access rights are also assigned to nested groups/projects.

```
projects_and_groups:
  group/*:
    group_settings:
      visibility: private
      auto_devops_enabled: false
      wiki_access_level: disabled
    group_members:
      users:
        root:
          access_level: owner
        gitlab-worker:
          access_level: developer
      enforce: true
      keep_bots: true
      groups:
        access/managers:
          group_access: reporter
        access/automations:
          group_access: reporter
        access/study:
          group_access: reporter
    members:
        root:
          access_level: owner
    project_settings:
      visibility: private
      analytics_access_level: disabled

  group/subgroup/*:
    group_members:
      users:
        root:
          access_level: owner
      enforce: true
      keep_bots: true
      inherit: false
      groups:
        access/study:
          group_access: maintainer
```

So, for subgroups with 3–7 levels of nesting, direct permission assignments are applied `access/study.group_access: maintainer`, which should not happen, because GitLab already provides permission inheritance.

sentoz:
There may already be a handle, and I just didn't understand how to use it correctly, but without this fix, I see two paths:

1. First, reset all permissions, and then assign them without including recursion — i.e., two runs. However, in that case, issues may arise for other automations when they intersect.
1. Describe all the nested groups below, but that would result in a very large YAML file, and you'd also have to constantly monitor for the appearance of new subgroups.

TimKnight01:
Hmmm direct permission assignment to subgroups should not happen as you say, GitLab should propagate that down. This to me feels like a bug that needs fixing _rather_ that something we solve with a new field @sentoz

sentoz:
If this is a bug, that changes everything. I'll write tests and a fix a bit later.

TimKnight01:
There are two things to think about:

1. I set `Dev` permission for user-group 1 on Group A. user-group 1 gets `Dev` permissions on all sub-groups of A by dint of Gitlab's permission propagation. **GLF** does not need to set `Dev` permission for user-group 1 on all sub-groups of A
2. I set `Dev` permission for user-group 1 on Group A. I set `Maintainer` permission for user-group 1 on Group A/sub-group-b/sub-group-c. **GLF** must set `maintainer` permission on sub-group-c for user-group 1, but otherwise doesn't need to set permissions as Gitlab will propagate `Dev` permission down

PR #1303, opened 2026-06-03
fix: collect subgroup projects for recursive group targets

### LINKED ISSUE #1302: Projects in descendant subgroups are skipped when running against a group target with recursion enabled
**Describe the bug**

When GitLabForm is executed with a group target and subgroup recursion enabled, projects inside descendant subgroups may be skipped entirely.

This issue was discovered while working on PR #1301 ,which added acceptance coverage for the new `propagate: false` behavior. The new tests did not expose a regression in that feature itself; instead, they revealed an older bug in project discovery.

The problem appears to be in `ProjectsProvider.get_projects()` in `gitlabform/lists/projects.py`. When the target is a group and recursive subgroup traversal returns more than one group, the code falls back to treating the target as a single project instead of collecting projects from the full group hierarchy. As a result, subgroup projects are not processed and GitLabForm prints `Nothing to do.` even though matching projects exist.

This was exposed by the acceptance test:
`TestVariablesPropagation.test__common_level_propagation_break__subgroup_local_redefinition_flows_to_projects_below`
from `tests/acceptance/standard/test_project_variables.py`.

Expected behavior:
A project located in a subgroup under the target group should be discovered and processed.

Actual behavior:
The subgroup project is skipped, no processing happens, and the test fails because the expected configuration is never applied.

This issue was discovered in the context of PR #1301 

**GitLabForm version**

`5.5.4`

sentoz:
This PR fixes project discovery for group targets when subgroup recursion is enabled.

Previously, `ProjectsProvider.get_projects()` used `len(groups.get_effective()) == 1` to decide whether the target should be treated as a group or as a single project. That worked for a group without subgroups, but broke as soon as recursive traversal returned more than one effective group. In that case, GitLabForm incorrectly fell back to `_get_single_project(target)` and skipped projects in descendant subgroups.

The fix changes this decision point to use whether any effective groups were found at all. If `get_groups(target)` returns groups, the target is treated as a group and project collection proceeds through `_get_projects(...)`. Only when no groups are found does GitLabForm fall back to `_get_single_project(target)`.

This PR also adds a regression test covering a group target with recursive subgroup discovery, ensuring subgroup projects are collected through the group path rather than being treated as a single-project target.

Closes #1302

### DIFF
--- a/gitlabform/lists/projects.py
+++ b/gitlabform/lists/projects.py
@@ -40,7 +40,7 @@ class ProjectsProvider(GroupsProvider):
         groups = self.get_groups(target)
 
         if target not in ["ALL", "ALL_DEFINED"]:
-            if len(groups.get_effective()) == 1:
+            if groups.get_effective():
                 projects = self._get_projects(target, groups)
             else:
                 projects = self._get_single_project(target)
--- a/tests/acceptance/standard/test_project_variables.py
+++ b/tests/acceptance/standard/test_project_variables.py
@@ -531,7 +531,10 @@ class TestVariablesRegression:
         group_for_function,
         subgroup_for_function,
     ):
-        project = create_project(subgroup_for_function, "descendant_subgroup_project")
+        project = create_project(
+            subgroup_for_function,
+            "descendant_subgroup_project",
+        )
         try:
             config = f"""
             projects_and_groups:
--- a/tests/acceptance/standard/test_project_variables.py
+++ b/tests/acceptance/standard/test_project_variables.py
@@ -5,7 +5,7 @@ from logging import info  # for wraps
 from unittest.mock import patch
 from gitlab.exceptions import GitlabListError
 
-from tests.acceptance import run_gitlabform
+from tests.acceptance import create_project, run_gitlabform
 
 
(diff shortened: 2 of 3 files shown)


TimKnight01:
@sentoz - thanks for the unit test with the change, are you able to create an Acceptance Test which would demonstrate the regression against an instance of GitLab?

sentoz:
> @sentoz - thanks for the unit test with the change, are you able to create an Acceptance Test which would demonstrate the regression against an instance of GitLab?

Added

sentoz:
our regression test passes; the remaining failure is unrelated

TimKnight01:
@sentoz looks good to me, can you fix the lint issues and then we can proceed

TimKnight01 (approved):

rickbrouwer:
About 

https://github.com/sentoz/gitlabform/blob/263b26d240637610bb4eb25983347bc5483ec709/gitlabform/lists/projects.py#L118

Is leaving the `len(groups.get_effective()) == 1` check here intentional (since with recursion it also yields multiple effective groups and skips filtering config projects outside the target group)?

PR #1308, opened 2026-06-07
ci: restructure github actions workflow and updated release process

amimas:
Currently, there are separate GitHub Actions workflows for running acceptance tests on GitLab premium vs ultimate (i.e. `.github/workflows/tests-premium.yml`, `.github/workflows/tests-ultimate.yml`). There's another workflow with similar name (`.github/workflows/tests-standard.yml`) that runs acceptance test on GitLab standard/community edition but it also runs various other types of tests. This overloaded what that workflow file implies or maybe not intuitive what kind of jobs belong in which workflows.

Re-organized the workflows using domain based jobs (i.e. build, test, static checks, etc.) to make it a bit more easier to follow.

**Overall change summary**:

- A single re-usable workflow for Acceptance test, which runs all kinds of acceptance tests
- A single re-usable workflow that takes care of builds - this includes packaging and verifying the package, building docs, running smoke test on the built package
- A single re-usable workflow for all static checks. Right now security test using `bandit` is part of this workflow, but we can look into splitting that into a separate security focused workflow later
- Workflow file rename and add `name` property to jobs, workflows. This helps to shorten how "checks" are displayed in the PR UI.
- Release workflow updates:
    - Removed redundant jobs (i.e. linting, acceptance test, etc.) from release workflow since they already run on `main` branch workflow
    - Added a job that checks if `main` branch workflow for the commit was successful so that release can proceed - included option allowing maintainers to manually trigger the release. Either way the release is triggered, it will only proceed if the selected version tag's commit has a successful completion of main branch workflow that performed build, and tests.
    - Docs site deployment job consumes artifacts from build workflow, instead of rebuilding it again - would like to adopt this in future for pypi and container image so that build-once-and-deploy method is used
- Take codecov token from repository secret

**Notes**:

Somethings to keep in mind for before/after merging:

- Add codecov token into repository secret
    - We should look at updating this job because I believe we're on older version right now and with newer version, there's no need to provide a token.
- Since job/workflow names are being updated, it'd affect required checks. They should be updated accordingly in the repository setting after this is merged


amimas:
@TimKnight01 - Could you please review this?

Initially I was hoping to add more checks/jobs into the CI so that we can move towards automated release process and also update the process so that build artifacts are used for release instead of "rebuilding". But, ended up restructuring the workflows first. Hope the update makes sense.

You can check [my fork](https://github.com/amimas/gitlabform) where I've tested this. Here's a [test PR](https://github.com/amimas/gitlabform/pull/12), [main branch workflow](https://github.com/amimas/gitlabform/actions/runs/27101348278), and a [release workflow](https://github.com/amimas/gitlabform/actions/runs/27101954312), which ofcourse fails at publishing steps since I haven't targeted separate pypi or container registry. Unfortunately, I couldn't test how the "PR from forked repo" will work.

Please let me know if you have any questions.

TimKnight01 (approved):
@amimas this looks good to me, I like the tidy up of releases, so that we don't re-run Acceptance Tests when we don't need to, as they will have run on `main` already

.github/workflows/_releases.yml
```diff
@@ -4,71 +4,90 @@ on:
...
+              per_page: 50,
+            });
+            const mainRun = runs.data.workflow_runs.find(
+              run => run.head_sha === context.sha && run.name === 'Main branch'
+            );
+            if (!mainRun) {
+              throw new Error('No completed main workflow was found for this commit.');
+            }
+            if (mainRun.status !== 'completed' || mainRun.conclusion !== 'success') {
+              throw new Error(`Main workflow did not pass. status=${mainRun.status} conclusion=${mainRun.conclusion}`);
```
TimKnight01: The only downside here is that when we push a Tag, it'll immediately trigger this job, which will obviously fail cos the `main` job still running for the SHA will have also only just started.

Is there anyway we can "sleep" this job or similar?

TimKnight01 (changes requested):

.github/workflows/_releases.yml
amimas: Doh! 🤦 . You're right! I forgot about that. I'm sure we can add some logic to wait if the main workflow is in-progress. Does the approach make sense though? I'm not a fan of the inline script, to be honest.

TimKnight01: @amimas yes in general I like the split of the Build pipeline as you have it, and the idea that on a release we don't have to re-run all the tests, just check if the relevant SHA pipeline on Main completed successfully e.g. had Green Tests

docs/contrib/releases.md
```diff
@@ -20,4 +20,6 @@ We try to follow the [PEP 440](https://peps.python.org/pep-0440/) versioning sch
     - Upload new version of gitlabform to [pypi package registry under gitlabform](https://pypi.org/project/gitlabform/).
     - A corresponding [GitHub release](https://github.com/gitlabform/gitlabform/releases) will be created that references the new tag.
 
+    The release workflow also verifies that the commit on `main` passed the normal main-branch checks before publishing. If a release must proceed despite issues, a maintainer can manually trigger the `Releases` workflow and enable `force: true` to bypass that verification.
```
amimas: Need to update the doc to reflect latest state.

.github/workflows/_releases.yml
amimas: @TimKnight01 - I ended up addressing this slightly differently, which I think is more consistent with rest of the setup. Here's what changed in latest update as of now:

Dropped that yaml-inline nodejs based script. I don't like scripts inside a yaml config - it's not testable and can be hard to maintain. Plus it introduced nodejs when everything else is based on python. Didn't need add any waiting logic due to re-work of the github workflows - more details below.

Incorporated the release validation logic into our existing dev toolkit (`uv run release`). This added a bit more logic into `dev/release.py` than the original nodejs based approach but it does cover all of our checks, I think. Because of the github workflow re-work (see below), we can now check whether release workflow is run from the same commit as the main workflow (successful means build and tests passed) and has a corresponding version tag.

While adding the release validation logic into our `uv run release` package/command, I realized existing structure had a bit problem. For example: python build & verify, docker build & verify, pypi publish, ghcr publish, were all bundled inside `dev/release.py`. It was a bit overloaded. So, I've moved the python build & verify into `dev/package.py` (they were already available as `uv run package`) and moved docker build & verify into `dev/docker.py` (they were already available as `uv run docker`).

Reworked the github workflows. Now main branch workflow **triggers** the release workflow. Optionally we can manually trigger the release workflow. If triggered manually, we have to provide the release tag and the **run id** of the main workflow. The release validation logic added into the toolkit (mentioned above) will ensure release is triggered on a version tag corresponding to main workflow. If it passes, rest of release jobs will proceed. Whether pass or fail, a summary/note will be available in the workflow run itself.

image

With those re-work of the workflow, the release workflow will only trigger when needed and proceed if allowed.

Hopefully I've covered everything when testing it. Here's my [fork repo workflows](https://github.com/amimas/gitlabform/actions) if you want to browse.

Please let me know if you have any questions.

TimKnight01: @amimas thanks for the description, I've used triggers before in both Github and GitLab and they work very well for this sort of thing

docs/contrib/releases.md
amimas: Updated.

TimKnight01:
Once this is in we'll need to update the Repo branch protection settings to remove the old job names from "status checks" and add the new ones:
image

Which is fine, it's a few mins of click-ops to get the repo correct

TimKnight01 (approved):

.github/workflows/_releases.yml
amimas: Thanks for the review. I'll merge it shortly and update the PR check settings in the repo.

PR #1309, opened 2026-06-08
feat!: upgrade versions of ruamel.yaml and yamlpath

TimKnight01:
BREAKING CHANGE:
- Drops default support for YAML 1.1
- Adds upgrade documentation
- Drops DeepDiff as assert equals handles this natively now
- Drops ez_yaml due to it's slow update cadence, resulting in dependency management mismatch between native ruamel.yaml and yamlpath. We are able to provide a simple interface for dumping ruamel.yaml to a string, reducing codebase dependencies and makes managing/configuring ruamel.yaml easier to avoid regressions


docs/upgrade.md
```diff
@@ -2,6 +2,25 @@
...
+In order to retain `yaml 1.1` configuration you **must** add the `%YAML` directive to the top of your configuration file, for example:
+
+```yaml
+%yaml 1.1
+---
+ projects_and_groups:
+   some_group/*:
+     project_settings:
+       foo: yes
+```
```
amimas, line 20: I think in yaml, `---` indicates multi-document file? Is that intentional or required?

docs/upgrade.md
```diff
@@ -2,6 +2,25 @@
...
+
+Alternatively update your YAML to be compatible with `yaml 1.2` specification, for example by replacing `yes` with `true` in the above example, and any instances of `no` with `false`.
```
amimas, line 22: Would be good to check if we have examples anywhere else in the docs that uses `yes`, `no` type syntax. If we do, they should be updated.

gitlabform/configuration/transform.py
```diff
@@ -58,9 +59,9 @@ def convert_to_simple_types(configuration: Configuration):
         # we needed complex ruamel.yaml's types like ordereddict and CommentedSeq
         # for transformations, but at the end convert them to simple dict and lists
         # for easier to understand debug output and tests
+        config_yaml_string = util.yaml_config_to_string(configuration.config)
 
-        config_yaml_string = ez_yaml.to_string(obj=configuration.config, options={})
-        simple_yaml_loader = YAML(typ="safe", pure=True)
+        simple_yaml_loader = util.configure_ruamel_yaml_loader(typ="safe", pure=True)
         configuration.config = simple_yaml_loader.load(config_yaml_string)
```
amimas, line 65: Does the code comment need updates?

gitlabform/processors/project/project_variables_processor.py
```diff
@@ -65,12 +67,14 @@ def _print_diff(
...
-                    ez_yaml.to_string(variables_list),
-                    "  ",
+            if self.log_level == INFO:
+                yaml = util.yaml_config_to_string(variables_list)
+                info(
+                    textwrap.indent(
+                        yaml,
+                        "  ",
+                    )
                 )
```
amimas, line 77: Is there a reason why we need to print config of variables like this? I don't think we do this for any other processors/config entities. Maybe we follow what other processors are doing, unless what they are doing is wrong. Since this PR is marked as breaking change anyways, it should be okay to change in this area.

Ideally I'd love to implement diff logic centrally so that we don't need to do this for individual processors. But that'd be a separate feature and is out of scope of this change.

tests/acceptance/standard/test_files_templates.py
```diff
@@ -81,14 +81,15 @@ def test__disabled_templating(self, project):
...
-                template: no
+                template: false
                 content: |
                   This is a text containing literals: {{{{ group }}}}/{{{{ project }}}}
         """
 
         run_gitlabform(config, project.path_with_namespace)
 
         project_file = project.files.get(ref="main", file_path="README.md")
+        expectation = "This is a text containing literals: {{ group }}/{{ project }}\n"
```
amimas: Was this supposed to be used in assertions? Also, why is this change needed here?

tests/unit/configuration/test_yaml_version.py
```diff
@@ -28,9 +41,23 @@ def configuration_with_yes_annotated():
     return Configuration(config_string=config_yaml)
 
 
-def test__default_yaml_11(configuration_with_yes):
-    assert configuration_with_yes.get("projects_and_groups|some_group/*|project_settings") == {"foo": True}
+def test__unannotated_yaml_11(yaml11_configuration_with_yes):
+    """
+    By default, in v7.+ we use YAML 1.2, so the "yes" value is not interpreted as a boolean.
```
amimas: What is v7.+ referring to?

pyproject.toml
```diff
@@ -43,11 +41,8 @@ dependencies = [
     "python-gitlab[graphql]==8.4.0",
     "requests==2.34.2",
     "rich==15.0.0",
-    # yamlpath and ruamel and tightly coupled, currently 0.19.1 of ruamel and 3.9.0 of yamlpath break support for
-    # yaml 1.1 "yes" -> True conversion. If we want to upgrade we'd need to make a breaking-change and drop support for
-    # unannotated yaml 1.1
-    "ruamel.yaml==0.17.21",
-    "yamlpath==3.8.2",
+    "ruamel.yaml==0.19.1",
+    "yamlpath==3.9.1",
```
amimas, line 45: Not sure if I missed it. I didn't see `yamlpath` being imported in any of the files that were changed in this PR. How is this related to ruamel - or is it?

amimas (commented):
Thank you so much for looking into it @TimKnight01 . The `deepdiff` seems unrelated to yaml dependency. That could be a separate PR without a major relesae, right? Or did I miss some details?

TimKnight01:
@amimas I could rollback the deepdiff change if you'd like, but it significantly simplifies and improves those unit test readability, whilst removing another dependency. I would prefer if we're making a breaking change and dropping dependencies, we might as well do it all in one go. I can separate into another PR but would still prefer to roll it into this breaking change release

docs/upgrade.md
TimKnight01: @amimas yeah it's required when supplying the %YAML directive

docs/upgrade.md
TimKnight01: No other references to `yes` or `no` in the Docs or .py files

gitlabform/configuration/transform.py
TimKnight01: @amimas no, the comment still counts, the util method performs the same process as ez_yaml.to_string did in v 1.1 (and in fact v2.0.0 which also worked, but in 2.2 they've massively changed up the logic when I looked at the codebase - the code in yaml_config_to_string is in fact copied from ez_yaml v2.0.0).

gitlabform/processors/project/project_variables_processor.py
TimKnight01: @amimas honestly I have no idea why it prints out the variables like this! But I thought it made sense to introduce logic which means it won't try convert the yaml to string if the `info()` function won't actually do anything (e.g. logging level is set to WARNING or higher). As that reduces the time to complete and chances of random errors in code that would never actually do anything useful.

I think another Issue/Pr to clean this up centrally would make sense to me

tests/acceptance/standard/test_files_templates.py
TimKnight01: Hold over from me debugging, will remove

tests/unit/configuration/test_yaml_version.py
TimKnight01: Me forgetting how to count 😂 

We're on v5 right now, so post this change we should be on v6, somehow I convinced myself we would be on v7 after this PR

pyproject.toml
TimKnight01: @amimas - `yamlpath` has a tight coupling to `ruamel.yaml` -> yamlpath imports ruamel.path and has constraints within it's install script e.g.
```
  install_requires=[
        "ruamel.yaml>0.17.5,!=0.17.18,<=0.17.21",
        "python-dateutil<=3"
    ],
``` 
-> from v3.8.0 of yamlpath

so in order to get our package installing I needed to bump them both to the most recent versions

amimas:
@TimKnight01 - thanks for the replies on the inquiries above.

Is there any reason why we still need to support yaml 1.1 style syntax? Regardless of which approach we take, this is a breaking change anyways. I'm thinking we simply drop yaml 1.1 support.

What do you think?

TimKnight01:
@amimas to be honest we don't have to do anything special to support YAML 1.1 the current version of the parsers handle it all for us when people provide the version header. And 1.1 hasn't entered end of life, so there's little reason to explicitly drop it imho

gitlabform/processors/project/project_variables_processor.py
amimas: Separate issue/PR sounds good to me. Let's open the issue so we remember to revisit this and then close this comment.

pyproject.toml
amimas: Thanks for explaining.

amimas:
Thanks for looking into this @TimKnight01 .

amimas (approved):

PR #1317, opened 2026-06-17
fix: validate  key in files config to avoid cryptic KeyError

### LINKED ISSUE #1312: Error occurred while processing project: exception: 'branches'
**Describe the bug**

I'm trying to update the branch settings in my projects but gets the error:
```
[17:30:58] WARNING  Error occurred while processing project      __init__.py:520
                    infra/dns-infrastructure, exception: 'branches'                                                  
``` 

**GitLabForm version**

5.5.4

**GitLab version**

[v19.0.1-ee](https://gitlab.com/gitlab-org/gitlab/-/tags/v19.0.1-ee)

This is the configuration:

```yaml
config_version: 4

projects_and_groups:
  infra/*:
    project_settings:
      default_branch: main
      only_allow_merge_if_pipeline_succeeds: true
      only_allow_merge_if_all_discussions_are_resolved: true
      remove_source_branch_after_merge: true

    branches:
      main:
        protected: true
        push_access_level: maintainer
        merge_access_level: maintainer
        allow_force_push: false

    tags:
      "*":
        protected: true
        create_access_level: maintainer
    files:
      ".gitignore":
        skip_ci: true
        overwrite: true
        file: ".gitignore"
        commit_message: "chore:\ sync .gitignore with central template"
```
Thanks in advance for the help.

rickbrouwer:
The files processor accessed `configuration["files"][file]["branches"]` directly, raising `KeyError: 'branches'` when users omitted the key. The misleading error pointed users at their top-level `branches:` section instead of the missing sub-key under the file entry.

Now exits with EXIT_INVALID_INPUT and a clear message listing the accepted values (`all`, `protected`, or a list of branch names).

Fixes #1312


gitlabform/processors/project/files_processor.py
```diff
@@ -33,7 +33,16 @@ def _process_configuration(self, project_and_group: str, configuration: dict):
...
+            file_config = configuration["files"][file]
+            if "branches" not in file_config:
+                critical(
+                    f"File '{file}' is missing the required 'branches' key."
+                    f" Specify which branches to apply the file to using one of:"
+                    f" 'all', 'protected', or a list of branch names (e.g. ['main'])."
+                )
+                sys.exit(EXIT_INVALID_INPUT)
+
+            config_target_ref = file_config["branches"]
```
amimas: I _believe_ `branches` is a special key introduced by gitlabform for convenience - GitLab's API expects `branch`, which is set based on the input provided in `branches`.

https://docs.gitlab.com/api/repository_files/#update-a-file-in-a-repository

Not sure if we're doing it consistently but in following 2 processors, the `_can_proceed` function is used for validating mandatory key like this one.

https://github.com/gitlabform/gitlabform/blob/8c7f75a9e285e2a2f64b524b335a2f8654c1ffaf/gitlabform/processors/project/merge_requests_approvals.py#L39-L48

https://github.com/gitlabform/gitlabform/blob/8c7f75a9e285e2a2f64b524b335a2f8654c1ffaf/gitlabform/processors/project/branches_processor.py#L41-L48

amimas (commented):
Thanks for looking into this @rickbrouwer . In general, looks good. Just have a question/suggestion on the approach.

gitlabform/processors/project/files_processor.py
rickbrouwer: Yeah. If have aligned it to `_can_proceed`.
One small addition: skipped files (`skip: true`) bypass the validation, since requiring a `branches:` key on a file you've explicitly opted out of would be confusing. I think that is better. Or would you prefer it strict?

amimas: Thanks @rickbrouwer for moving the validation into `_can_proceed`.


Regarding the `skip: true`, here's my thought. If we do include this update, we should add some log message to indicate something is not being processed. But, I'm leaning towards not adding this logic. It's a very tiny change but I feel it's a new feature. Currently the `abstract_processor.py` has the logic that uses `skip: true` for top-level config sections (e.g. `project_settings`, `branches`, `schedules`, `members`, `files`, etc.). Your logic is adding this option to a level deeper (e.g.  `files.<file-name>.skip`).

https://github.com/gitlabform/gitlabform/blob/f2ec72bb2d301a9833ec0bd85d3a41cda0e146b0/gitlabform/processors/abstract_processor.py#L34-L36

Not saying no to this, but I think that should be a separate PR as it's a new feature, if the community needs it. In that case, would be good to assess whether we need to provide this option for `files` only or for others too. My personal opinion is that I'm not a big fan of the `skip` option in general. My thought is that gitlabform provides config-as-code workflow. So, the config file should represent how the target system should behave. Having `skip` option adds noise to this configuration. If something is not desired, it shouldn't exist in the config. But, this is my personal opinion.

rickbrouwer: I dropped the skip check.

As I saw it, this piece already existed:

https://github.com/gitlabform/gitlabform/blob/main/gitlabform/processors/project/files_processor.py#L32-L34

So, it isn't really new, only how the can_proceed interacts with it. But I removed it.

amimas (approved):

gitlabform/processors/project/files_processor.py
amimas: Ahh... thanks for that pointer. Looks like this feature already existed for `files` processor since the beginning.

TimKnight01 (approved):

PR #1318, opened 2026-06-17
fix: don't re-validate inherited  in subgroup lookup

### LINKED ISSUE #1316: inherit: false on a project-only key (e.g. merge_requests_approval_rules) aborts when a wildcard also matches a subgroup
**Describe the bug**

When a group/* wildcard defines a project-level key (merge_requests_approval_rules) with inherit: false, and that wildcard also matches a subgroup under group/, gitlabform aborts at config-read with a CRITICAL:

```
CRITICAL The inheritance-break flag set in "group/subgroup" under key
         "merge_requests_approval_rules" is invalid because it has no
         higher level setting to inherit from.                          core.py:161
```
The break is perfectly valid for the projects the wildcard matches (they inherit the global * rule to break from). It only fails for the subgroup, because the global *'s merge_requests_approval_rules is a project-scoped setting that isn't considered applicable to a group — so there's "nothing to inherit from" at the group level, and the inherit-break validation rejects it.

gitlabform already dispatches keys by entity type at apply time (it applies group_members only to groups, project_settings only to projects). The gap is that the read-time inherit-break validation runs before/independent of that dispatch and is entity-type-blind — it validates a project-only key against a group, where it can never be valid.

## Minimal reproduction

Group layout in GitLab:

```
acme/                      (group)
  acme/widget              (project)      <- matched by "acme/*"
  acme/tools/              (subgroup)     <- matched by "acme/*"  ← triggers the abort
    acme/tools/helper      (project)
```

`config.yml`:

```yaml
config_version: 4
projects_and_groups:
  "*":
    merge_requests_approval_rules:
      standard:
        approvals_required: 1
        name: "All eligible users"
      enforce: true
  "acme/*":
    group_members:
      groups:
        some-team:
          group_access: maintainer
      enforce: true
    merge_requests_approval_rules:
      strict-main:
        approvals_required: 1
        name: "Strict - main"
        protected_branches: ["main"]
      enforce: true
      inherit: false          # valid for acme/widget (a project); orphaned for acme/tools (a subgroup)
```

**GitLabForm version**

gitlabform 5.5.4 (latest stable)

**GitLab version**

GitLab 19.0.1-ee

rickbrouwer:
Fixes the `inherit: false` on a wildcard ancestor (e.g. `acme/*`) caused a false abort when resolving a subgroup (e.g. `acme/tools`) that had no own config layer. `_get_effective_subgroup_config` re-validated the ancestor's config under the subgroup's name, without the common config in scope. Each subgroup layer is now validated only when no parent exists above it (no common config and no non-empty accumulated ancestor).

Fixes #1316


TimKnight01 (approved):

TimKnight01:
@rickbrouwer I think this will need a Rebase to bring it in line with the refactored CI config, otherwise the required checks will continue showing as not completed

gitlabform/configuration/groups.py
amimas: For clarity, could we set the variable type?


```suggestion
        common_config: dict = self.get_common_config()
```

tests/unit/configuration/test_inheritance_break_subgroups.py
```diff
@@ -134,3 +134,30 @@ def test__inheritance_break__flag_set_at_project_level__second_subgroup_project_
...
+          acme/*:
+            merge_requests_approval_rules:
+              inherit: false
+              override_rule:
+                approvals_required: 2
+        """
+
+        configuration = Configuration(config_string=config_yaml)
+
+        effective_config = configuration.get_effective_config_for_group("acme/tools")
```
amimas, line 157: Do we need to add a comment that `acme/tools` is a project, not a subgroup? I haven't looked at these tests before. So, not sure if it's needed.

tests/unit/configuration/test_inheritance_break_subgroups.py
```diff
@@ -134,3 +134,30 @@ def test__inheritance_break__flag_set_at_project_level__second_subgroup_project_
...
+          "*":
+            merge_requests_approval_rules:
+              some_rule:
+                approvals_required: 1
+          acme/*:
+            merge_requests_approval_rules:
+              inherit: false
+              override_rule:
+                approvals_required: 2
+        """
```
amimas, line 153: Do we need another test case to confirm that inheritance is working for subgroups (i.e. `acme/subgroup` is still inheriting `some_rule`  and also adding a new rule? Or, is that already covered by existing tests?

amimas (commented):
Thanks for looking into this @rickbrouwer . Left a few comments, mostly for my own understanding around the tests.

tests/unit/configuration/test_inheritance_break_subgroups.py
amimas: Just realized that `merge_requests_approval_rules` is applicable for projects only setting/config. So, I guess this test  also validates that gitlabform doesn't fail when it comes across `acme/subgroup`, right?

rickbrouwer: Yes. The test confirms that `get_effective_config_for_group("acme/tools")` no longer aborts on the inherited `inherit: false`, and that the resulting config is the correct replacement (only `override_rule`, common `*` dropped).

tests/unit/configuration/test_inheritance_break_subgroups.py
rickbrouwer: `acme/tools` is treated as a subgroup here (the call is `get_effective_config_for_group`, not `get_effective_config_for_project`). So no extra comment needed in my opinion.

gitlabform/configuration/groups.py
rickbrouwer: applied

tests/unit/configuration/test_inheritance_break_subgroups.py
amimas: You're right! I missed that.

tests/unit/configuration/test_inheritance_break_subgroups.py
amimas: Thanks @rickbrouwer for explaining. If I understood the issue, the breaking inheritance has 2 paths:

- the section is applicable for projects only
- the section is applicable for groups only

Whichever is applicable and if parent config for that section does not specify the same section, then `inherit: false` should fail. Do we have tests covering both paths?

I also wonder if `inherit: false` should even cause failure in the first place. It could be a warning. If there's nothing to inherit from, there's no harm anyways to the section being processed? 🤔

rickbrouwer: Validation is entity-type agnostic. `_validate_break_inheritance_flag` only checks for a parent layer, not whether the section is project- or group-only. Existing tests in `test_inheritance_break_validation.py` cover all scopes (`common`, `group`, `subgroup`, `project`) with various sections.

On warn vs fail: that's a behaviour change outside this fix. So out of the scope of this PR. If we want to do that, it would be better in another PR, wouldn't it?

amimas: Thanks again for clarifying that the validation is not based on entity type. On warn vs fail, I agree it'd be better in another PR. I think the change in this PR would still be applicable in that case?

amimas (approved):

PR #1330, opened 2026-06-26
chore(github): Update bug report template to include used config and contribution question

rickbrouwer:
Added a section to inquire about contributing a fix.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -19,7 +19,13 @@ Output of `gitlabform -V`
 
 F.e. `13.12.4-ee`
 
+**GitLabForm configuration**
+
+The relevant part of your gitlabform config (optionally anonymized).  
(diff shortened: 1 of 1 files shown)


amimas (commented):
I like the update. Thanks @rickbrouwer .

Since you're updating this, could we add instructions for sharing gitlabform config being used? Sometimes it's even helpful to have before-and-after config. Issues often misses this info and sometimes it's not easy to tell what's happening from description.

rickbrouwer:
@amimas 

Yes, that seems like a very good idea to me. What do you think is the right way to phrase the questions here?
Something like:
```
**GitLabForm configuration**

The relevant part of your gitlabform config (optionally anonymized).
If applicable, share both the before and after config.
```

amimas:
I'm good with your suggested content/question. We can update it later if needed.

rickbrouwer:
> I'm good with your suggested content/question. We can update it later if needed.

Done

PR #1331, opened 2026-06-26
fix(group_saml_links): stop delete-and-recreate churn on unchanged links

### LINKED ISSUE #1328: Unchanged configuration for group_saml_link is deleted and recreated
**Describe the bug**

Users are intermittently reporting being dropped out of groups configured using group_saml_links.

When defining the following:

teams/test/*:
  group_saml_links:
    maintainers:
      saml_group_name: gitlab-test-maintainer
      access_level: maintainer
    develpers:
      saml_group_name: gitlab-test-developer
      access_level: developer

I've noticed that upon every run of GitLab Form, when I check the audit logs for the group, I can see that the configured group_saml_links are first removed, then re-added. The behavior we'd expect to see is that if the group_saml_link is defined in the config, configured on the group, and there were no changes, there should be no changes applied.

The same behavior happens with `enforce: true` configured as well. Except, any group not defined in the config is removed & not re-added. The behavior we'd expect to see is that only groups that have changes are processed.

**GitLabForm version**

6.0.0

**GitLab version**

18.11.5-ee

rickbrouwer:
The processor compared the configured field `saml_group_name` against the GitLab API field `name` via the generic `_needs_update` helper.

Because `saml_group_name` was always "only in configuration", the helper returned True on every run, deleting and recreating each link.

Compare via the API field instead, drop the misplaced `group.save()` calls, and filter the `enforce` control flag out before iterating so it cannot be processed as a link.

Fixes #1328

### DIFF
--- a/gitlabform/processors/group/group_saml_links_processor.py
+++ b/gitlabform/processors/group/group_saml_links_processor.py
@@ -29,7 +29,13 @@ class GroupSAMLLinksProcessor(AbstractProcessor):
 
             if existing_link is None:
                 group.saml_group_links.create(link_configuration)
-            elif self._link_differs(existing_link, link_configuration):
+                continue
+
+            # GitLab API uses 'name' where the config uses 'saml_group_name'.
+            expected = {**link_configuration, "name": saml_group_name}
+            expected.pop("saml_group_name", None)
+
+            if self._needs_update(existing_link.asdict(), expected):
                 # GitLab API has no update endpoint for SAML links; delete and recreate instead.
                 debug(f"Updating SAML link: {saml_group_name} with {link_configuration}")
                 existing_link.delete()
@@ -37,15 +43,6 @@ class GroupSAMLLinksProcessor(AbstractProcessor):
 
         if enforce_links:
             self._delete_extra_links(group, existing_links, configured_links)
-
-    def _link_differs(self, existing: GroupSAMLGroupLink, configured: dict) -> bool:
-        # 'saml_group_name' was already used to look up the existing link, so the names match.
-        for key, value in configured.items():
-            if key == "saml_group_name":
-                continue
-            if getattr(existing, key, None) != value:
-                return True
-        return False
 
     def _delete_extra_links(
         self,
--- a/gitlabform/processors/group/group_saml_links_processor.py
+++ b/gitlabform/processors/group/group_saml_links_processor.py
@@ -31,7 +31,9 @@ class GroupSAMLLinksProcessor(AbstractProcessor):
                 group.saml_group_links.create(link_configuration)
                 continue
 
-            # GitLab API uses 'name' where the config uses 'saml_group_name'.
+            # GitLab API's request and response attributes differs when SAML link is created vs retrieved.
+            # On creation or retrieving a link, attribute is called `saml_group_name` but and returned response is 'name' instead.
+            # To compare existing link (retrieved from GitLab) with configured link, link configuration need to replace `saml_group_name` with `name`
             expected = {**link_configuration, "name": saml_group_name}
             expected.pop("saml_group_name", None)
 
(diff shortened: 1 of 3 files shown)


rickbrouwer:
tests are flaky. trying to solve that with https://github.com/gitlabform/gitlabform/pull/1332

gitlabform/processors/group/group_saml_links_processor.py
```diff
@@ -11,53 +11,50 @@ class GroupSAMLLinksProcessor(AbstractProcessor):
     def __init__(self, gitlab: GitLab):
         super().__init__("group_saml_links", gitlab)
 
-    def _process_configuration(self, group_path: str, configuration: dict) -> None:
+    def _process_configuration(self, group_path_and_name: str, configuration: dict) -> None:
         """Process the SAML links configuration for a group."""
 
-        configured_links = configuration.get("group_saml_links", {})
+        configured_section = configuration.get("group_saml_links", {}) or {}
```
amimas, line 17: For my learning, why do we need `or {}`? Doesn't the existing code already do that?

gitlabform/processors/group/group_saml_links_processor.py
```diff
@@ -11,53 +11,50 @@ class GroupSAMLLinksProcessor(AbstractProcessor):
...
-        # Process each configured SAML link
         for _, link_configuration in configured_links.items():
             saml_group_name = link_configuration.get("saml_group_name")
+            existing_link = existing_by_name.get(saml_group_name)
 
-            if saml_group_name not in existing_link_names:
-                # Create the saml link as it does not already exist
+            if existing_link is None:
+                group.saml_group_links.create(link_configuration)
+            elif self._link_differs(existing_link, link_configuration):
```
amimas: I might be missing some details. Any reason the `_needs_update()` function can't be used for checking if the existing link and configuration has drifted?

amimas (commented):
Thanks @rickbrouwer for looking into it. It's good that the PR includes unit tests. For processors, our current practice relies on acceptance tests using a live gitlab instance. Could you please update existing acceptance tests? I think at least one new test case can be added to check if unchanged config is not causing a removal and re-add flow.

gitlabform/processors/group/group_saml_links_processor.py
rickbrouwer: Correct, the previous version didn't have `or {}`. In practice `AbstractProcessor` usually skips empty sections so it rarely matters, but `or {}` makes this method safe on its own. It is a bit defensive.

gitlabform/processors/group/group_saml_links_processor.py
rickbrouwer: Fair, a small remap saves the helper. I'll switch.

gitlabform/processors/group/group_saml_links_processor.py
```diff
@@ -11,40 +11,36 @@ class GroupSAMLLinksProcessor(AbstractProcessor):
...
             saml_group_name = link_configuration.get("saml_group_name")
+            existing_link = existing_by_name.get(saml_group_name)
+
+            if existing_link is None:
+                group.saml_group_links.create(link_configuration)
+                continue
+
+            # GitLab API uses 'name' where the config uses 'saml_group_name'.
+            expected = {**link_configuration, "name": saml_group_name}
+            expected.pop("saml_group_name", None)
```
amimas: Suggesting a bit more explanation for clarity, if this makes sense to you. In my view, it's a GitLab API consistency issue than GitLabForm's config issue.


```suggestion
            # GitLab API's request and response attributes differs when SAML link is created vs retrieved.
            # On creation or retrieving a link, attribute is called `saml_group_name` but and returned response is 'name' instead.
            # To compare existing link (retrieved from GitLab) with configured link, link configuration need to replace `saml_group_name` with `name`
            expected = {**link_configuration, "name": saml_group_name}
            expected.pop("saml_group_name", None)
```

amimas (approved):
Looks good to me. Suggested a code comment for clarity.

PR #1332, opened 2026-06-26
test(remote-mirrors): stabilize flaky mirror sync test

rickbrouwer:
Sidekiq's libcurl resolves localhost to ::1 inside the GitLab container, but nginx only listened on IPv4. That caused intermittent ECONNREFUSED in test_remote_mirror_http_password_auth_sync.

Also poll the mirror's update_status on the source project so future regressions surface GitLab's last_error instead of a generic timeout.


amimas (commented):
Thanks for looking into this @rickbrouwer . Really glad if this makes the remote-mirror test less flaky - ideally not flaky at all.

Did you get a chance to run this test repeatedly locally? Curious if the change actually runs the tests more smoothly.

rickbrouwer:
Ran it locally a few times; they all went well this time.

Also ran it locally a few times with the "old" code; not a single one succeeded.

amimas (approved):

PR #1343, opened 2026-07-02
fix: diff log should not require verbose

### LINKED ISSUE #1315: `--diff-only-changed` should show diffs without requiring `--verbose`
### Problem statement:

`--diff-only-changed` suggests the tool will show only the settings that would change. However, without `--verbose`, no diffs are shown at all — the output only lists the processed projects with no indication of what would change:
```
* (49/64) Processing project: asdf
* (50/64) Processing project: fdsa
```

Adding `--verbose` makes the diffs appear, but also floods the output with section processing messages that `--diff-only-changed` should arguably suppress:

```
* (49/64) Processing project: asdf
           INFO     Skipping section 'project' - not in config.
           INFO     Processing section 'project_settings' in dry-run mode.
           INFO     project_settings changes:
                    visibility: "private" => "public"
           INFO     Skipping section 'project_security_settings' - not in config.
           INFO     Processing section 'project_push_rules' in dry-run mode.
           INFO     Diffing for section 'project_push_rules' is not supported yet
           ...20 more lines of noise...
```

### Expected behavior:

`--noop --diff-only-changed` should show the diffs at the default log level, without requiring `--verbose`. Ideally, the output for a project with one change would look something like:
```
* (49/64) Processing project: asdf
project_settings changes:
  visibility: "private" => "public"
```

Projects without changes could either be omitted or shown without additional detail.

Current workaround:

We use `--verbose` and pipe through grep, which also requires setting `COLUMNS` to prevent Rich from wrapping log lines:
```
gitlabform --noop --verbose --diff-only-changed ALL_DEFINED 2>&1 | grep -Ev "Skipping section|Diffing for section|Processing section"
```

Environment:

GitLabForm 5.5.4
Python 3.12

TimKnight01:
- adds custom log levels to support two custom log types
- resolves #1315


gitlabform/constants.py
```diff
@@ -5,3 +5,10 @@
 
 # legacy single approval rule name
 APPROVAL_RULE_NAME = "Approvers (configured using GitLabForm)"
+
+# ------ Logging constants for custom log levels -----
+# Log out Diff logs higher than WARNING (i.e. not requiring --verbose too) but without having to masquerade as an
+# error/unexpected by using logger.warning/error etc
+DIFF_LOG_LEVEL = 32
+# Log out notices about number of failed projects etc
+NOTICE_LOG_LEVEL = 33
```
amimas, line 16: How did we arrive at these numbers? Could we add some note about it?

amimas: Is the comment accurate? I think the `NOTICE_LOG_LEVEL` is not being used for failed projects only?

TimKnight01:
@amimas updated following comments, thank you

amimas (approved):
From a quick glance over the file changes, it looks good to me @TimKnight01 . But, I haven't had a chance to try it out locally. Do we have any existing test cases or quick way to add some tests? Maybe we can look at those in the CI. Otherwise, would be nice to add some samples execution in the PR description from your local test run.

TimKnight01:
@amimas - you can see some of the results in the CI -> more logging, I can see if we can add one in, we'd need to I think check if it got written out with caplog, should be doable 🤔

amimas:
Hey @TimKnight01 - Sorry, not sure what you were referring to as "in the CI -> more logging". Are you still looking into this? Just checking if we're good to merge this. The branch needs updating though.

TimKnight01:
@amimas - sorry I've started a new role at work so it's eaten a lot of my time, will try and get back to this this week if I can

amimas:
No worries. I completely understand. Hope you're enjoying the new role.

docs/contrib/coding_guidelines.md
```diff
@@ -6,6 +6,12 @@ Please do:
 * add tests along with the new code that prove that it works:
     * when adding/changing non-trivial logic please add unit tests,
     * for all bug fixes and new features using GitLab API please add acceptance tests
+* use info as the base log level, debug if you might expose sensitive information (such as raw API request/responses)
+    * [Logging levels information](https://docs.python.org/3/howto/logging.html#logging-levels)
+    * by default we log out WARNING and above issues, with INFO logged out when specifying the `--verbose` argument in CLI 
+    * use `logging` module for logging, do not use `print` statements
+    * we have a custom log level for "diff" messages, used only in the difference_logger
+    * we have a custom log level "notice" used only in gitlabfor __init__.py in order to output version information and the total numbers of Projects/Groups processed
```
rickbrouwer, line 14:
```suggestion
    * we have a custom log level "notice" used only in gitlabform __init__.py in order to output version information and the total numbers of Projects/Groups processed
```

rickbrouwer (changes requested):

docs/contrib/coding_guidelines.md
```diff
@@ -6,6 +6,12 @@ Please do:
 * add tests along with the new code that prove that it works:
     * when adding/changing non-trivial logic please add unit tests,
     * for all bug fixes and new features using GitLab API please add acceptance tests
+* use info as the base log level, debug if you might expose sensitive information (such as raw API request/responses)
+    * [Logging levels information](https://docs.python.org/3/howto/logging.html#logging-levels)
+    * by default we log out WARNING and above issues, with INFO logged out when specifying the `--verbose` argument in CLI
```
rickbrouwer, line 11:
```suggestion
    * by default we log out WARNING and above issues, with INFO logged out when specifying the `--verbose` argument in CLI
```

rickbrouwer (changes requested):

gitlabform/__init__.py
rickbrouwer:
```suggestion
from rich.logging import RichHandler
from rich.console import Console
from rich.markup import escape
from rich.theme import Theme
```

gitlabform/__init__.py
```diff
@@ -33,8 +38,12 @@
 from gitlabform.processors.group import GroupProcessors
 from gitlabform.processors.project import ProjectProcessors
 
+# Setup rich console to output nicely formatted version information
 console = Console()
```
rickbrouwer:
```suggestion
# Setup rich console to output nicely formatted version information
console = Console()

# Separate console for the logging handler, with styles for our custom log levels
log_console = Console(
    theme=Theme(
        {
            "logging.level.diff": "bold magenta",
            "logging.level.notice": "bold cyan",
        }
    )
)
```

rickbrouwer:
I think docs/running.md also needs an update (there is a suggestion about --verbose --diff-only-changed)

Also I think we loses the styling. `RichHandler` has markup off by default (so, `:sparkles:` now prints literally ":sparkles:" and the red/green/yellow are gone).
Passing an `extra={"markup": True}` on those specific calls brings it back. So, please add this an test if I'm correct. I'd avoid `markup=True` on the handler itself btw, since diff output can contain brackets and would then raise a `MarkupError`.

amimas:
Thanks for those feedbacks @rickbrouwer . I just pushed an update. When you get a chance, could you please review?

rickbrouwer:
@amimas 
Is it okay if I check this out and make some adjustments? I'll make a separate commit so we can review the whole thing.

rickbrouwer:
I’m just going to go ahead and do that (sorry 😉 ). If it’s not allowed or anything, I’m can drop the commit.

Personally, I think this commit does the trick, so I’m approving it now.
Feel free to take a look at it, and if you’re happy with it too, go ahead and merge it.

rickbrouwer (approved):

amimas:
Ahh... ofcourse it's okay! Thanks for those adjustments and the review. Looks good to me. Going to merge this now.

PR #1348, opened 2026-07-09
fix(ci): upload unit test coverage to codecov for fork PRs

rickbrouwer:


.github/workflows/tests-unit.yml
```diff
@@ -38,15 +33,18 @@ jobs:
...
         run: uv run --no-sync qa test tests/unit --cov=. --cov-report=xml --log-cli-level=WARNING
-      - name: Upload coverage to Codecov
-        if: ${{ env.CODECOV_TOKEN != '' }}
-        uses: Wandalen/wretry.action@v3
+      - name: Save PR metadata
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          printf '{"pr": %s, "sha": "%s"}\n' "$PR_NUMBER" "$PR_SHA" > pr-info.json
```
TimKnight01, line 41: @rickbrouwer can we use the same process for the Acceptance Tests, would need to pass in the relevant flags and append those to the report name? but would be good if we can be consistent in how we upload codecov?

Happy to merge this in, check it works and then refactor

PR #1351, opened 2026-07-10
fix: escape regex as raw string in core.py

rickbrouwer:
The following message appears in the log:

```
/app/.venv/lib/python3.14/site-packages/gitlabform/gitlab/core.py:120: SyntaxWarning: "\d" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\d"? A raw string is also an option.
  sem_ver_regex = "^\d*\.\d*\.\d*"
```

So, we use raw string for semver regex to silence SyntaxWarning on Python 3.14

### DIFF
--- a/gitlabform/gitlab/core.py
+++ b/gitlabform/gitlab/core.py
@@ -117,7 +117,7 @@ class GitLabCore:
         current_version = "0.0.0"
 
         if self.version != "unknown":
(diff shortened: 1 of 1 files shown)


amimas (approved):
Thank you for looking into it.

TimKnight01 (approved):

PR #1352, opened 2026-07-11
refactor(processors): migrate hooks to MultipleEntitiesProcessor

rickbrouwer:
Fold HooksProcessor and GroupHooksProcessor into the existing MultipleEntitiesProcessor and fix a bug where `delete: true` on an entity absent from GitLab would fall through to the create branch contrary to the existing "do nothing if marked as 'delete'" comment).

### DIFF
--- a/gitlabform/processors/group/group_hooks_processor.py
+++ b/gitlabform/processors/group/group_hooks_processor.py
@@ -1,75 +1,56 @@
-from logging import debug, warning
-from typing import Dict, Any, List
+from logging import warning
 
-from gitlab.base import RESTObject, RESTObjectList
 from gitlab.v4.objects import Group
-from gitlab.v4.objects import GroupHook
 
 from gitlabform.gitlab import GitLab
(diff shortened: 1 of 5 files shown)


amimas:
Hi @rickbrouwer - I'm not sure what bug is being addressed but let's not migrate towards the `MultipleEntitiesProcessor`.

Some background info that is probably not clearly documented yet. The `MultipleEntitiesProcessor` is an artifact of GitLabForm creating and maintaining GitLab API. We have been gradually migrating our code base towards python-gitlab library. See #73 and https://github.com/orgs/gitlabform/projects/1. There used to be another processor called `SingleEntityProcessor` and so far we were able to migrate all the processors that dependent on that.

It's possible some processors have shared/similar logic between group vs project based entity. In that case, we can follow what's being done with other similar processors like "variables processor".

https://github.com/gitlabform/gitlabform/tree/main/gitlabform/processors/util

So, let's not migrate back to legacy code :)

PR #1353, opened 2026-07-11
refactor(processors): centralize dry-run diff in AbstractProcessor

rickbrouwer:
Four processors carried near-identical `_print_diff` implementations (`project_settings`, `merge_requests_approvals`, `project_security_settings`, `project_variables`). This pulls the shared flow into `AbstractProcessor` and removes the TODO in `merge_requests_approvals.py`.


gitlabform/processors/abstract_processor.py
```diff
@@ -124,7 +127,22 @@ def _process_configuration(self, project_or_project_and_group: str, configuratio
         pass
 
     def _print_diff(self, project_or_project_and_group: str, entity_config, diff_only_changed: bool):
-        info(f"Diffing for section '{self.configuration_name}' is not supported yet")
+        if self.get_entity_in_gitlab is None:
+            info(f"Diffing for section '{self.configuration_name}' is not supported yet")
```
TimKnight01: arguably this should be `debug` level _or_ using the diff-log custom log level - not sure it adds much values for end-users who are doing "--verbose", but is useful for those explicitly using "--log-diff"

TimKnight01 (changes requested):

gitlabform/processors/abstract_processor.py
rickbrouwer: i will change this to `debug`

rickbrouwer: done

gitlabform/processors/abstract_processor.py
```diff
@@ -124,7 +127,22 @@ def _process_configuration(self, project_or_project_and_group: str, configuratio
         pass
 
     def _print_diff(self, project_or_project_and_group: str, entity_config, diff_only_changed: bool):
-        info(f"Diffing for section '{self.configuration_name}' is not supported yet")
+        if self.get_entity_in_gitlab is None:
+            debug(f"Diffing for section '{self.configuration_name}' is not supported yet")
+            return
+
+        entity_in_gitlab = self.get_entity_in_gitlab(project_or_project_and_group)
+        entity_in_gitlab, entity_config = self._prepare_entities_for_diff(entity_in_gitlab, entity_config)
```
amimas: Line 134 and 135 is a bit confusing for me. In line 134, we're getting the `entity_in_gitlab`, but that's being overwritten again in line 135? It's not clear why. Should the 2nd instance be named something else?

I can see in `_prepare_entities_for_diff` function below, we're simply returning what's in the function's parameter. Is that function intended to be modified by some subclass only? If yes, should that just set to `pass` and let the subclass define the logic?

amimas (commented):
Thanks for looking into this @rickbrouwer . I think overall direction/structure is fine. Not sure if it's just me, but is it possible to make the change/logic for this a bit more easy to follow in the abstract processor?

I'm also wondering whether the abstract processor's `__init__` function call the subclass function and save it as `entity_in_gitlab`. I think there are other areas where this can be used in future for reducing some of the boilerplates from the subclasses in future.

rickbrouwer:
> Thanks for looking into this @rickbrouwer . I think overall direction/structure is fine. Not sure if it's just me, but is it possible to make the change/logic for this a bit more easy to follow in the abstract processor?
> 
> I'm also wondering whether the abstract processor's `__init__` function call the subclass function and save it as `entity_in_gitlab`. I think there are other areas where this can be used in future for reducing some of the boilerplates from the subclasses in future.

Yeah. Will refactor the extension mechanism to the template method pattern.

gitlabform/processors/abstract_processor.py
rickbrouwer: Will consolidate this in two hooks

gitlabform/processors/abstract_processor.py
```diff
@@ -123,8 +124,30 @@ def _should_retry_processor(e: Exception) -> bool:
...
+    def _get_entities_for_diff(
+        self, project_or_project_and_group: str, entity_config: dict
+    ) -> Optional[tuple[dict, dict]]:
+        """Return (gitlab_side, config_side) ready to be compared, or None to opt out.
+
+        Override in subclasses to enable the centralized dry-run diff for their section.
+        Both sides must be keyed the same way so `log_diff` can align them; if the shapes
+        differ between GitLab and the user's config, normalize them here.
+        """
+        return None
```
amimas, line 133: Should we keep the "gitlab_side" and "config_side" as 2 separate items. Right now printing diff is the only logic in the abstract processor that needs both.

By keeping them separate, we can probably use them for other uses cases similar to "print diff", but in that case the function name can be misleading or, maybe only of these entity is needed in that use case.

What do you think?

rickbrouwer: I'd keep the tuple. The variables processor needs both sides. It renames config-side keys to match the GitLab-side ones, so dropping config_side would just push that normalization into a second hook, which is exactly the pattern we just removed. The name already scopes it to diffing, so if a future use case only needs the GitLab state, that can be its own hook rather than reshaping this one now.

amimas: Thanks @rickbrouwer . We don't have to worry about future use case for now.

But something still seems odd to me. It has to do with how the getter hook is structured: it forces subclasses to implement a method that just returns one of the parameters passed into it. For example: the `project_settings_processor` does this:

```py
    def _get_entities_for_diff(self, project_path: str, entity_config: dict) -> tuple[dict, dict]:
        return self.gl.get_project_by_path_cached(project_path).asdict(), entity_config
```

Here, it's returning `entity_config` back to `AbstractProcessor` - do not need to transform or do anything with it. This applies to most cases.

What I'm suggesting is something like this - just a pseudo code; needs adjusting:

```py
class AbstractProcessor:
    def _get_desired_state(self, configuration: dict) -> dict:
        """Default: returns full configuration. 

        Subclasses ONLY override this if they need to slice or transform 'configuration'.
        """
        return configuration

    @abstractmethod
    def _get_current_state(self, project_or_group) -> dict:
        """Subclasses MUST implement this to fetch current state from GitLab API."""
        pass

    def process(self, project_or_group, configuration):
        if self.dry_run:
            before = self._get_current_state(project_or_group)
            after = self._get_desired_state(configuration)
            
            self._print_diff(before=before, after=after)
            return

        # ...
```

I'm hoping this is clearer, but I could be wrong. Please let me know if I missed something.

rickbrouwer: Yeah, I understand. I'd go with your split, maybe with one small adjustment (on `_get_current_state`)

gitlabform/processors/project/project_security_settings.py
amimas: I think here we can now utilize the new methods (i.e. `_get_current_state`, `_get_desired_state`). Or, should we keep it as-is?

gitlabform/processors/project/project_variables_processor.py
```diff
@@ -45,56 +41,31 @@ def _can_proceed(self, project_or_group: str, configuration: Dict[str, Any]) ->
...
+    @staticmethod
+    def _variable_identity(var: Dict[str, Any]) -> str:
+        return f"{var.get('key')}@{var.get('environment_scope', '*')}"
+
+    @staticmethod
+    def _masked_variable(var: Dict[str, Any]) -> Dict[str, Any]:
+        masked = {k: v for k, v in var.items() if k not in {"id", "_links"}}
+        if "value" in masked:
+            masked["value"] = hide(str(masked["value"]))
+        return masked
```
amimas, line 76: Please add some brief docstring with explanations about why they are needed and/or how it works.

gitlabform/processors/abstract_processor.py
```diff
@@ -123,8 +124,33 @@ def _should_retry_processor(e: Exception) -> bool:
...
+        Override in subclasses to enable diffing for their section. Return None (the
+        default) to opt out.
+        """
+        return None
+
+    def _get_desired_state(self, entity_config: dict) -> dict:
+        """Return the config side of the diff. Override only when the config needs
+        slicing or normalizing so its keys line up with the current state.
+        """
+        return entity_config
```
amimas, line 139: GitLabForm's config adds some extra keys that are not in GitLab's API. These keys are for GitLabForm's feature and they are not necessarily available in every processor or maybe in nested configs. For example: `enforce`, `strict`, `protected`, etc. Should they be stripped out from here? If they are not stripped out and the processors don't do it, how does it affect the diff?

amimas (commented):
I think the latest update looks really good @rickbrouwer . Thanks for keep working on it. Left couple of minor comments.

Just one thing stands out I think that might need a bit more clarification. Does the raw responses from `python-gitlab` contain read-only API fields (id, created_at, project_id, _links, etc.). If `_get_current_state` returns raw API objects while `_get_desired_state` returns clean YAML data from gitlabform config, then `_print_diff` will show every extra API field as a "deletion" or "modification."

gitlabform/processors/project/project_variables_processor.py
rickbrouwer: Will add

gitlabform/processors/abstract_processor.py
rickbrouwer: This isn't new to this PR. The four old per-processor _print_diff implementations had the same behavior. I'd prefer to leave the cleanup as a per-section follow-up (I am not doing this follow-up, so for who is interested (I want to focus next on migrating remaining processors)).

gitlabform/processors/project/project_security_settings.py
rickbrouwer: Will do for for `_get_current_state`.

gitlabform/processors/abstract_processor.py
amimas: I agree with you. We don't have to make that enhancement in this PR. I'll open a separate issue for it.

amimas: Opened issue #1404 .

amimas (approved):
Thanks again for working on this @rickbrouwer . I feel this will make it really easy for other proecessors to consume this and show drift/diff in the dry-run.

PR #1357, opened 2026-07-12
fix(processors): skip delete:true entities absent from GitLab

rickbrouwer:
Previously, when a config entry had `delete: true` for an entity that wasn't in GitLab, `MultipleEntitiesProcessor` fell through to the create branch instead of skipping.


rickbrouwer:
Just leave this open for a little while longer. I don't think it's necessary because I want to see if I can phase out the MultipleEntitiesProcessor.

rickbrouwer:
Closing this one in favor of https://github.com/gitlabform/gitlabform/pull/1359

PR #1359, opened 2026-07-14
refactor: migrate remaining processors to python-gitlab

rickbrouwer:
Migrate 6 processors from the legacy `MultipleEntitiesProcessor`/`_make_requests_to_api` path to the python-gitlab library: project badges, group badges, deploy keys, MR approval rules, group LDAP links, and protected environments.

I think only `members_processor.py` and the helpers in `gitlab/core.py`, `projects.py`, `groups.py` still use some legacy code.

Okay...this PR is big. I could have done it piece by piece, but things might be intertwined, and I could have made a choice to do it piece by piece, PR...merge and move on. Or proceed all at once to avoid potential conflicts.


rickbrouwer:
A lot of work. I don't put this in draft mode anymore because otherwise it just sits there. I rely mainly on the unit tests and acceptance tests here.

amimas:
I love the end result of this PR @rickbrouwer 😍 . Thanks for putting all that work on it.

I feel it makes it equally difficult/time consuming to review it properly though. Worst case if we need to temporarily undo the changes, we have to git revert everything instead of just a single processor or maybe have a bigger blast radius. I recall from some of the earlier refactors, we actually had to add a lot more test cases because existing tests didn't cover all the features/scenarios. In some case, the processors were completely written from scratch.

I would prefer doing this change one processor refactor/update per PR. Since you've gone through the changes required, maybe we can get through the easiest one first. Could we do that please?

rickbrouwer:
Yes, let's do that. I'll do them one by one. I'll start by removing old code, which is possible anyway, and one for the badges. Once that's merged, I'll move on to the next one.

PR #1388, opened 2026-08-03
refactor(badges): migrate badges processor to python-gitlab

rickbrouwer:
Migrate the badges processor to python-gitlab's native `.badges` CRUDMixin, replacing the manual REST wrappers in `gitlab/group_badges.py` and `gitlab/project_badges.py`.

Since groups and projects expose the same `.badges` manager, project and group badge handling now share a single helper in `processors/util/`.

Number 1 from https://github.com/gitlabform/gitlabform/pull/1359


gitlabform/processors/util/badges_processor.py
```diff
@@ -0,0 +1,111 @@
...
+            name = badge.get("name")
+            if name is None:
+                continue
+            if name in seen:
+                critical(
+                    f"Entities {seen[name]} and {key} in {configuration_name} for {display_name}"
+                    f" are the same in terms of their defining key: 'name'"
+                )
+                sys.exit(EXIT_INVALID_INPUT)
+            seen[name] = key
```
amimas: I'm not sure how this is being done currently using the legacy code. Here I'm thinking we exit after the loop completes, if duplicates are found. This way users get to know of all the duplicates instead of one by one.

The per entry log can probably be changed from `critical` to `warning` and then use `critical` when we exit like we have here.


What do you think?

gitlabform/processors/util/badges_processor.py
```diff
@@ -0,0 +1,111 @@
...
+    ) -> List[Union[GroupBadge, ProjectBadge]]:
+        badges = group_or_project.badges.list(get_all=True)
+        # For projects, .badges.list() also returns badges inherited from the parent group,
+        # so we must filter to project-kind only. The group endpoint returns only group badges.
+        if isinstance(group_or_project, Project):
+            return [b for b in badges if b.kind == "project"]
+        return list(badges)
+
+    @staticmethod
+    def _find_duplicates(configuration_name: str, display_name: str, configured_badges: Dict) -> None:
```
amimas: To me the function name `_find_duplicates` implies it'll return some data to me, that may be either empty or have a list of duplicate items. But, that's not happening in the function. We're actually exiting the process, if any duplicates are found.

We could make it something like `_has_duplicate_badges` and make it a `boolean` return type. That way, caller has to decide what should happen if duplicates exist. Or maybe keep existing logic of exiting the proceess but rename it.

gitlabform/processors/util/badges_processor.py
```diff
@@ -0,0 +1,111 @@
...
+    @staticmethod
+    def _list_existing_badges(
+        group_or_project: Union[Group, Project],
+    ) -> List[Union[GroupBadge, ProjectBadge]]:
+        badges = group_or_project.badges.list(get_all=True)
+        # For projects, .badges.list() also returns badges inherited from the parent group,
+        # so we must filter to project-kind only. The group endpoint returns only group badges.
+        if isinstance(group_or_project, Project):
+            return [b for b in badges if b.kind == "project"]
+        return list(badges)
```
amimas, line 78: What if it's a subgroup? Will that contain badges from parent group similar to the projects?

gitlabform/processors/util/badges_processor.py
```diff
@@ -0,0 +1,111 @@
...
+    @staticmethod
+    def _validate_required(configuration_name: str, display_name: str, entity_name: str, badge_config: Dict) -> None:
+        missing = [key for key in ("name", "link_url", "image_url") if not badge_config.get(key)]
+        if missing:
+            critical(
+                f"Entity {entity_name} in {configuration_name} for {display_name}"
+                f" doesn't have some of its keys required to create or update:"
+                f" {', '.join(missing)}"
+            )
+            sys.exit(EXIT_INVALID_INPUT)
```
amimas: I don't recall the processors validating if the config contains attributes required by GitLab's API. This makes GitLabForm tied to GitLab's API. Any changes to their API around this will require changes in GitLabForm, which we try to avoid as much as possible.

Instead, I think we should catch the relevant create/update exceptions so that users have a friendly/clear message about why the config failed.

gitlabform/processors/util/badges_processor.py
```diff
@@ -0,0 +1,111 @@
...
+            self._validate_required(configuration_name, display_name, entity_name, badge_config)
+
+            if matching:
+                if needs_update(matching.asdict(), badge_config):
+                    info(f"Editing {entity_name} of {configuration_name} in {display_name}")
+                    for key, value in badge_config.items():
+                        setattr(matching, key, value)
+                    matching.save()
+                else:
+                    info(f" * {entity_name} of {configuration_name} in {display_name} doesn't need an update.")
```
amimas: Do we currently do this for badges or other entities - logging what doesn't need update?

amimas (commented):
Thanks @rickbrouwer for this PR. I love that no changes were needed in the acceptance tests. To me, that means there aren't any regression - unless we're missing test scenarios.

Left a few comments for discussion. Aside from that, can we change the PR title and description? At least the PR title should be a bit more generic and say something along the line of badges processor being migrated to python-gitlab. Sharing the project/group logic is implementation detail.

gitlabform/processors/util/badges_processor.py
rickbrouwer: It returns only that group's own badges (even for subgroups); only the project endpoint returns inherited ones, which we filter. If you want I can add a code comment?

gitlabform/processors/util/badges_processor.py
rickbrouwer: This is consistent with how it was. I can rename it if you prefer. Yeah, ok, I will do that.

gitlabform/processors/util/badges_processor.py
rickbrouwer: This preserves existing behaviour. If we drop this then that would be a behaviour change.
Also here, we can open a separate issue to move all processors toward catching API exceptions.

gitlabform/processors/util/badges_processor.py
rickbrouwer: That it's maybe nicer indeed, but the fail-fast pattern here mirrors multiple_entities_processor.py:130-139. I prefer to keep parity. If you want this, then we must address this consistently across all processors in a separate issue.

gitlabform/processors/util/badges_processor.py
rickbrouwer: This is existing behaviour. Changing this convention feels out of scope here.

gitlabform/processors/util/badges_processor.py
rickbrouwer: added comment

amimas: Thanks for confirming and adding it to the comment.

gitlabform/processors/util/badges_processor.py
amimas: That's a good point, although I'm not sure if we can/should mirror `MultipleEntitiesProcessor` 100%. I imagine the same topics will come up again for remaining processors that need to be migrated. So, we'll either need revisit every processors again or we can take care of it now.

Do you suggest we do this later?

Having said that, I looked at couple of other processors that would've used the `MultipleEntitiesProcessor` if we continued with it. One is `Variables` and the other one is `Labels`.

I don't see them doing exactly same logic as this. Neither one does a `sys.exit` today or doesn't try to find duplicates. They either raise exception where needed or returns boolean. Or create/update logic skips over "existing matching entity", which I think is similar to duplicate findings.

https://github.com/gitlabform/gitlabform/blob/9ab386c7a0e5a048bb79cb11262305639c88782b/gitlabform/processors/util/variables_processor.py

https://github.com/gitlabform/gitlabform/blob/9ab386c7a0e5a048bb79cb11262305639c88782b/gitlabform/processors/util/labels_processor.py

Even if we do change the behaviour later, what is the right approach for badges? That way we know what issue to create or set the pattern for other migrations that we still need to do.

gitlabform/processors/util/badges_processor.py
amimas: Similar to my earlier comment, I'm not sure we should preserve all the existing behaviour of `MultipleEntitiesProcessor`. It kind of bypasses GitLabForm's [raw parameter passing](https://gitlabform.github.io/gitlabform/reference/#raw-parameters-passing) design. Although it's not documented clearly, it keeps GitLabForm not tied with GitLab's API changes.

In this specific case, I feel removing this validation doesn't have an impact on end-users. They'll still get a failure when the required GitLab API parameters are not set. Only difference is how the error is presented to them. That's why I suggested capturing exceptions where needed. So, in the end, it's not exactly changing the behaviour.

Doing this now will save us extra work later, and it'll come up for other processors that still need to be migrated.

What do you think @rickbrouwer ?

amimas:
Hi @rickbrouwer - Just checking if you had a chance to look into this PR again. No rush.

rickbrouwer:
Not yet. I will continue with it next week.

gitlabform/processors/util/badges_processor.py
rickbrouwer: Will do, will match the pattern of the other util processors.

gitlabform/processors/util/badges_processor.py
rickbrouwer: Agree

amimas (approved):
Looks good to me. Thanks for taking care of this @rickbrouwer .

PR #1395, opened 2026-08-07
docs: use renamed secret_push_protection_enabled setting

### LINKED ISSUE #1394: Fix (pre_receive_secret_detection_enabled): Stop unchanged settings from being applied on every run
**Describe the bug**

In our top-level group config we have the following set:

    project_security_settings:
      pre_receive_secret_detection_enabled: true

Upon every run of GitLab Form, we can see in the project audit events "updated `secret_push_protection_enabled` to true" despite the fact it is already enabled / set to true.

We'd expect to see is that GitLab Form only changes settings that don't match, instead of updating every run.

**GitLabForm version**

v6.2.1

**GitLab version**

v18.11.7-ee

**GitLabForm configuration**
  teams/*:
    project_security_settings:
      pre_receive_secret_detection_enabled: true

rickbrouwer:
The `pre_receive_secret_detection_enabled` setting was renamed to `secret_push_protection_enabled` in GitLab 17.11. Update the example and document that configs using the old name trigger a redundant update on every run, since the API only reports the new name.

Also fix a broken API reference link that pointed to an LDAP anchor.

Fixes #1394

### DIFF
--- a/docs/reference/project_security_settings.md
+++ b/docs/reference/project_security_settings.md
@@ -11,12 +11,6 @@ protection](https://docs.gitlab.com/ee/user/application_security/secret_detectio
 
 Values are documented at the [Project security settings API docs](https://docs.gitlab.com/api/project_security_settings/#update-the-secret_push_protection_enabled-setting).
 
-!!! note
-
-    This setting was renamed from `pre_receive_secret_detection_enabled` to `secret_push_protection_enabled` in GitLab 17.11.
-    Use the new name `secret_push_protection_enabled`. Configs still using the old name will trigger a redundant
-    update on every run, because the GitLab API only reports the setting under its new name.
-
 ## Example
 
 ```yaml
--- a/docs/reference/project_security_settings.md
+++ b/docs/reference/project_security_settings.md
@@ -9,7 +9,13 @@ This section purpose is to manage project security settings, especially [secret 
 On Gitlab Dedicated and Self-managed instances, you must [allow secret push
 protection](https://docs.gitlab.com/ee/user/application_security/secret_detection/secret_push_protection/#allow-the-use-of-secret-push-protection-in-your-gitlab-instance) before you can enable it in a project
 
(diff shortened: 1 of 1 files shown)


docs/reference/project_security_settings.md
```diff
@@ -9,13 +9,19 @@ This section purpose is to manage project security settings, especially [secret
...
 protection](https://docs.gitlab.com/ee/user/application_security/secret_detection/secret_push_protection/#allow-the-use-of-secret-push-protection-in-your-gitlab-instance) before you can enable it in a project
 
-Values are documented at [LDAP Group Links section of the Groups API docs](https://docs.gitlab.com/ee/api/project_security_settings.html#update-pre_receive_secret_detection_enabled-setting).
+Values are documented at the [Project security settings API docs](https://docs.gitlab.com/api/project_security_settings/#update-the-secret_push_protection_enabled-setting).
+
+!!! note
+
+    This setting was renamed from `pre_receive_secret_detection_enabled` to `secret_push_protection_enabled` in GitLab 17.11.
+    Use the new name `secret_push_protection_enabled`. Configs still using the old name will trigger a redundant
+    update on every run, because the GitLab API only reports the setting under its new name.
```
amimas: @rickbrouwer - Is there a better way to communicate this? I mean, this is applicable to almost all the processors. Users need to be familiar with GitLab's API in terms of what parameters/attributes they can use. I think the example update below is fine, but this note probably belongs somewhere else as a more generic guidance, if it's not there already.

rickbrouwer: Shall we just close the PR then?

Because you are right. And what we want to avoid is documenting every rename action later on.

Should any future issues arise, we will point them in the right direction. Wdyt?

rickbrouwer: ow no, the rest must of course remain.

I am only removing the note

amimas: I'm good with that. Thank you for taking care of this.

amimas (approved):

PR #1409, opened 2026-08-29
build(dev): replace gitlabform development toolkit shortcuts with one entrypoint

amimas:
Currently we have a separate package that's used as a development toolkit. This is intended for development of gitlabform itself. The toolkit (cli) gives domain based options for various tasks such as: buiding and verifying python package, running linters, starting up a local gitalb instance, running tests, etc. Almost all of these commands are also used in the CI process. This keeps dev and CI process in sync.

When this toolkit was setup, they were added in `pyproject.toml` under the `[project.scripts]` configuration section. I found recently that this results in an expected behaviour. Each items becomes installed globally. For example: within the gitlabform repo, we can run `uv run dev -h`, but it also possible to run `dev -h` from anywhere, even outside the repository.

Each of the items from the dev toolkit that were added there was inteded as a shortcut. For example: instead of running `uv run dev workspace setup`, we can run `uv run workspace setup`. Because these are installed/available globally, it has a bigger consequence when there are naming collision. For example, the toolkit provides command `uv run dev docker build` or `uv run docker build`. Since the `docker` command from this toolkit becomes available globally, it collides with standard docker-cli.

This PR fixes the above issue in following ways:

- Rename the dev toolkit's entrypoint from `dev` to `glf-dev`. This way it's clear what this is referring to while keeping the name short for convenience.
- Removed all the shortcut entrypoints that were added before. Now all commands has to go through `glf-dev`. For example: `uv run glf-dev <command> <option>`
- Since `glf-dev` becomes available globally, added a check so that running `glf-dev` outside the gitlabform repo will not run or execute because it won't function as intended outside of the repo anyways.

### DIFF
--- a/.github/workflows/tests-acceptance.yml
+++ b/.github/workflows/tests-acceptance.yml
@@ -39,9 +39,9 @@ jobs:
       - name: Install dependencies
         run: uv sync --frozen --no-dev --group test
       - name: Start GitLab (${{ matrix.flavor }}) in docker
-        run: uv run --no-sync gitlab-local up --gitlab-flavor ${{ matrix.flavor }}
+        run: uv run --no-sync glf-dev gitlab-local up --gitlab-flavor ${{ matrix.flavor }}
       - name: Run Standard acceptance tests for ${{ matrix.flavor }} flavor
-        run: uv run --no-sync qa test tests/acceptance/standard --cov=. --cov-report=xml --durations=0 --reruns 3 --reruns-delay 10 --log-cli-level=WARNING
+        run: uv run --no-sync glf-dev qa test tests/acceptance/standard --cov=. --cov-report=xml --durations=0 --reruns 3 --reruns-delay 10 --log-cli-level=WARNING
       - name: Save PR metadata
         env:
           PR_NUMBER: ${{ github.event.pull_request.number }}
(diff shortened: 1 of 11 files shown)


.github/workflows/tests-acceptance.yml
```diff
@@ -39,9 +39,9 @@ jobs:
       - name: Install dependencies
         run: uv sync --frozen --no-dev --group test
       - name: Start GitLab (${{ matrix.flavor }}) in docker
-        run: uv run --no-sync gitlab-local up --gitlab-flavor ${{ matrix.flavor }}
+        run: uv run --no-sync glf-dev gitlab-local up --gitlab-flavor ${{ matrix.flavor }}
       - name: Run Standard acceptance tests for ${{ matrix.flavor }} flavor
-        run: uv run --no-sync qa test tests/acceptance/standard --cov=. --cov-report=xml --durations=0 --reruns 3 --reruns-delay 10 --log-cli-level=WARNING
+        run: uv run --no-sync glf-dev qa test tests/acceptance/standard --cov=. --cov-report=xml --durations=0 --reruns 3 --reruns-delay 10 --log-cli-level=WARNING
```
amimas, line 44: I believe the acceptance tests are failing because they are run with `pull_request_target` for the workflow. So, they are not using the workflow updates from this branch. That's why the CI is still triggering `uv run --no-sync gitlab-local` (notice it's missing `glf-dev`).

There were no changes to gitlabform code. So, it should be okay to merge this as is.

rickbrouwer (approved):
lgtm!

PR #1421, opened 2026-09-07
Implement group merge request approval settings

szEvEz:
Previously discussed [here](https://github.com/gitlabform/gitlabform/pull/914)
and Implements https://github.com/gitlabform/gitlabform/issues/701

### DIFF
--- a/gitlabform/processors/group/group_merge_requests_approvals_processor.py
+++ b/gitlabform/processors/group/group_merge_requests_approvals_processor.py
@@ -1,0 +1,55 @@
+from logging import debug
+from typing import Any, cast
+
+from gitlab.v4.objects.groups import Group
+
+from gitlabform.gitlab import GitLab
+from gitlabform.processors.abstract_processor import AbstractProcessor
+
+
+class GroupMergeRequestsApprovalsProcessor(AbstractProcessor):
+    def __init__(self, gitlab: GitLab):
+        super().__init__("group_merge_requests_approvals", gitlab)
+
+    def _get_current_state(self, group: str) -> dict:
+        gitlab_group: Group = self.gl.get_group_by_path_cached(group)
+        return self._get_group_merge_requests_approvals(gitlab_group)
+
+    def _process_configuration(self, group: str, configuration: dict) -> None:
+        settings_in_config: dict = configuration.get("group_merge_requests_approvals", {})
+
+        gitlab_group: Group = self.gl.get_group_by_path_cached(group)
+        settings_in_gitlab = self._get_group_merge_requests_approvals(gitlab_group)
+
+        debug(settings_in_gitlab)
+        debug("group_merge_requests_approvals BEFORE: ^^^")
+
+        if self._needs_update(settings_in_gitlab, settings_in_config):
+            debug(f"Updating group merge requests approvals settings for group {group}")
+            self._update_group_merge_requests_approvals(gitlab_group, settings_in_config)
+        else:
+            debug("No update needed for group merge requests approvals settings")
+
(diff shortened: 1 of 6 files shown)


gitlabform/processors/group/group_merge_requests_approvals_processor.py
```diff
@@ -0,0 +1,55 @@
...
+            info(f"Updating group merge requests approvals settings for group {group}")
+            self._update_group_merge_requests_approvals(gitlab_group, settings_in_config)
+        else:
+            info("No update needed for group merge requests approvals settings")
+
+    def _get_group_merge_requests_approvals(self, gitlab_group: Group) -> dict:
+        # TODO: python-gitlab has no manager for the /groups/:id/merge_request_approval_setting
+        # endpoint yet
+        settings = cast(dict[str, Any], self.gl.http_get(self._api_path(gitlab_group)))
+        return self._unwrap_values(settings)
```
amimas, line 37: I'm thinking this function should return the api response from gitlab as-is. The calling function should call `_unwrap_values`, if it needs.

gitlabform/processors/group/group_merge_requests_approvals_processor.py
```diff
@@ -0,0 +1,55 @@
...
+    def _api_path(gitlab_group: Group) -> str:
+        return f"/groups/{gitlab_group.id}/merge_request_approval_setting"
+
+    @staticmethod
+    def _unwrap_values(settings: dict) -> dict:
+        """Flatten GitLab's response to plain values. This API returns every setting as an object
+        with the `value`, `locked` and `inherited_from` keys (both on GET and on PUT), while in the
+        config the values are set directly. Without this nothing would ever compare as equal, so we
+        would re-apply the settings on every run and print the raw objects in the dry-run diff."""
+        return {key: setting["value"] for key, setting in settings.items()}
```
amimas, line 55: Is there a better name for this function? I haven't had a chance to think this through yet, so don't have a suggestion at the moment.

gitlabform/processors/group/group_merge_requests_approvals_processor.py
```diff
@@ -0,0 +1,55 @@
...
+        if self._needs_update(settings_in_gitlab, settings_in_config):
+            info(f"Updating group merge requests approvals settings for group {group}")
+            self._update_group_merge_requests_approvals(gitlab_group, settings_in_config)
+        else:
+            info("No update needed for group merge requests approvals settings")
+
+    def _get_group_merge_requests_approvals(self, gitlab_group: Group) -> dict:
+        # TODO: python-gitlab has no manager for the /groups/:id/merge_request_approval_setting
+        # endpoint yet
+        settings = cast(dict[str, Any], self.gl.http_get(self._api_path(gitlab_group)))
```
amimas, line 36: Let's wrap the logic in `try...except` for handling exceptions.

gitlabform/processors/group/group_merge_requests_approvals_processor.py
```diff
@@ -0,0 +1,55 @@
...
+        return self._unwrap_values(settings)
+
+    def _update_group_merge_requests_approvals(self, gitlab_group: Group, settings: dict) -> None:
+        updated_settings = cast(dict[str, Any], self.gl.http_put(self._api_path(gitlab_group), post_data=settings))
```
amimas, line 40: Let's wrap the logic in `try...except` for handling exceptions.

amimas (changes requested):
Thanks for working on this PR @szEvEz . Left a few comments for review.

Do we need the unit test? Some of the unit test cases aren't really testing the logic of the processor itself - for example the diff rendering logic. I feel the acceptance test should be good or add more test case there, if needed.

PR #1422, opened 2026-09-07
refactor(deploy-keys): migrate deploy-keys processor to python-gitlab

rickbrouwer:
Migrate deploy-keys processor to python-gitlab

Number 2 from https://github.com/gitlabform/gitlabform/pull/1359.


gitlabform/processors/project/deploy_keys_processor.py
```diff
@@ -1,18 +1,99 @@
...
-            # but NOT the value (according to https://docs.gitlab.com/ee/api/deploy_keys.html#update-deploy-key)
-        )
+        super().__init__("deploy_keys", gitlab)
+
+    def _process_configuration(self, project_and_group: str, configuration: Dict):
+        configured_keys: Dict = configuration.get("deploy_keys", {})
+        enforce = configured_keys.pop("enforce", False)
+
+        project: Project = self.gl.get_project_by_path_cached(project_and_group)
+        existing_keys: List[ProjectKey] = list(project.keys.list(get_all=True))
```
amimas: Doesn't `project.keys.list()` return a `list` already? Why do we ned to cast it here?

gitlabform/processors/project/deploy_keys_processor.py
```diff
@@ -1,18 +1,99 @@
...
+
+        # All deletions happen before any addition, because a deploy key's value is unique
+        # within the instance: adding a key back under another title only succeeds once the
+        # key under the old title is gone.
```
amimas: Is there a reference to GitLab's docs for this that can be added here?

gitlabform/processors/project/deploy_keys_processor.py
```diff
@@ -1,18 +1,99 @@
...
+        if enforce:
+            configured_titles = {key_config["title"] for key_config in configured_keys.values()}
+            for existing in list(existing_keys):
```
amimas: Just for my learning, why do we need to do `list(existing_keys)` instead of `existing_keys`?

gitlabform/processors/project/deploy_keys_processor.py
```diff
@@ -1,18 +1,99 @@
...
+
+        for entity_name, key_config in configured_keys.items():
+            if not key_config.get("delete", False):
+                continue
+
+            matching = next((k for k in existing_keys if k.title == key_config["title"]), None)
+            if matching:
+                info(f"Deleting {entity_name} of {self.configuration_name} in {project_and_group}")
+                matching.delete()
+                existing_keys.remove(matching)
```
amimas: I don't exactly remember but we normally add a log/warning if the config refers to an item to be deleted but it's not in gitlab?

gitlabform/processors/project/deploy_keys_processor.py
```diff
@@ -1,18 +1,99 @@
...
+            # GitLab sometimes returns HTTP 400 with "has already been taken" when adding an SSH key
+            # that already exists on another project, despite the docs saying it should just associate.
+            # As a workaround, look up the existing key on the instance and enable it for this project.
+            if e.response_code != 400 or "has already been taken" not in str(e):
+                raise
+
+            existing_id = self._find_existing_deploy_key_id(key_config["key"])
+            if existing_id is None:
+                raise
+            project.keys.enable(existing_id)
```
amimas: Can we flip the logic so that the actual condition is easier to understand?


```suggestion
            if e.response_code == 400 and "has already been taken" in str(e):
                existing_id = self._find_existing_deploy_key_id(key_config["key"])
                if existing_id is None:
                    raise
                project.keys.enable(existing_id)
            raise
```

amimas (commented):
Thanks for taking care of this processor's migration to python-gitlab. Overall looks good to me. I left a few comments to understand the code a little better.

Since we have acceptance tests running against a live gitlab instance, do the unit test provide any additional benefit? Are there any test case that should be in the acceptance test instead?

gitlabform/processors/project/deploy_keys_processor.py
rickbrouwer: Ah, yes, I will remove that.

gitlabform/processors/project/deploy_keys_processor.py
rickbrouwer: https://docs.gitlab.com/api/deploy_keys/#add-a-deploy-key-for-a-project

Not everything is quite right. I’m going to go through it all (the docs), including the links, to see if they’re correct. They might be outdated.

gitlabform/processors/project/deploy_keys_processor.py
rickbrouwer: I think we're not really consistent about that.
Here it is info:
https://github.com/gitlabform/gitlabform/blob/630a02e411c27238b1ef84dab728f42f74c38116/gitlabform/processors/project/remote_mirrors_processor.py#L93

and here debug:

https://github.com/gitlabform/gitlabform/blob/630a02e411c27238b1ef84dab728f42f74c38116/gitlabform/processors/project/schedules_processor.py#L40-L41

rickbrouwer: ah, i see, I forgot that..ha...will add!

gitlabform/processors/project/deploy_keys_processor.py
rickbrouwer: Because the loop removes items from existing_keys while iterating it, which would make the iterator skip elements.

gitlabform/processors/project/deploy_keys_processor.py
rickbrouwer: yes, that's ok. I will make the change because your `raise` on line 81 is not on the correct position I think.
I'm going to work on it this afternoon (it's morning here right now :))

gitlabform/processors/project/deploy_keys_processor.py
rickbrouwer: The removals turned out to be redundant though, so both are gone now.

gitlabform/processors/project/deploy_keys_processor.py
amimas: Thanks for adding the log.

I agree with you. It is inconsistent right now. I see you went with `info` for this PR. Also noticed that some of the log messages start with (` * `) prefix. To be honest, I'm not sure it's clear when to add those prefix or not or what log level to use. The old base class didn't do much logging either. So, at least this PR is an improvement over that.

I'll open a separate issue to look into the log messages later so that we have a consistent structure and perhaps have a way of making it easier/less-confusing from development perspective.

amimas: Opened issue #1425 . Closing this thread for now.

amimas (approved):

Issue #1427, opened 2026-09-14
ai_settings.ai_catalog_restricted_to_group: can only be set for top-level groups


itewk:
**Describe the bug**

Without changing any configuration on my side i recently started getting the following failure. though i think this started after a gitlab upgrade.

```
WARNING  Error occurred while processing group        __init__.py:464
Clients/exampels/dont-panic-inc, exception:                 
                                                            
400:                                                        
{'ai_settings.ai_catalog_restricted_to_group                
_hierarchy': ['can only be set for top-level                
groups']} 
```

The problem is that I don't know of any way to set a setting only on a parent group and not have it propagate to child groups, which would then make it impossible to set this setting. so maybe this is really just a duplicate of, https://github.com/gitlabform/gitlabform/issues/842 but makes it in stark relief that problem has to be solved.

**GitLabForm version**

Output of `gitlabform -V`

**GitLab version**

`[v19.2.6-ee]`

**GitLabForm configuration**

```
projects_and_groups:
  "*":
    group_settings:
      ai_settings_attributes:
        ai_catalog_restricted_to_group_hierarchy: true
```

**Would you be open to contributing a fix?**

- [ ] Yes
- [ ] No
- [x] Maybe

itewk:
one possible fix for this is: https://github.com/gitlabform/gitlabform/pull/1429

PR #1429, opened 2026-09-16
feat: add non-recursive group config syntax (`group/`)

### LINKED ISSUE #541: How to add users ONLY to top-level group?

itewk:
- Add `group/` trailing-slash key syntax for non-recursive group config, distinct from `group/*` which cascades to all subgroups
- Add `group_exact` key type in `_get_key_type()` to classify `/`-ending keys
- Update `get_groups()` to collect and deduplicate both `/*` and `/` keys
- Update `get_effective_config_for_group()` to merge exact config on top of recursive config before the common merge
- Add `exact` parameter to `_get_group_config()` to look up `group/` keys
- Add 14 unit tests covering config resolution, `inherit: false` interaction, key type classification, and project exclusion
- Update docs/reference/index.md with the new configuration level

Closes #541. Based on the approach from PR #986.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


itewk:
this would also provide a resolution to: https://github.com/gitlabform/gitlabform/issues/1427

PR #1432, opened 2026-09-19
fix(project): wait for asynchronous project transfer to complete

### LINKED ISSUE #1431: Support GitLab 19.4 behaviour changes
Since GitLab 19.4 project transfers are processed asynchronously:
  
  > Asynchronous transfers for projects introduced in GitLab 18.11 with a feature flag named `groups_and_projects_async_transfer`. Disabled by default.
  > Generally available in GitLab 19.4. Feature flag `groups_and_projects_async_transfer` removed.
  
([Manage projects](https://docs.gitlab.com/user/project/working_with_projects/)) (see "Transfer a project")
  
`PUT /projects/:id/transfer` now returns 200 while the project is still in its old namespace; it arrives in the destination a few seconds later. GitLabForm continues immediately, so a `transfer_from` run reports success before the transfer has actually happened, and the sections processed after it (`archive`, `project_settings`, `branches`, ...) fail with 404.

GitLabForm should wait for the transfer to complete before continuing.

rickbrouwer:
Since GitLab 19.4 project transfers are [asynchronous](https://docs.gitlab.com/user/project/working_with_projects/): `PUT /projects/:id/transfer` returns while the project is still in its old namespace. GitLabForm continued immediately, so a run combining `transfer_from` with other sections failed with `404 Project Not Found`, and a transfer-only run reported success before the transfer had happened.

Poll until the project shows up in the destination group.

Fixes #1431


amimas (commented):
Thanks @rickbrouwer for looking into it quickly.

Looks like the change makes gitlabform wait for the transfer to complete. Should it? What if maybe gitlabform follow gitlab and simply request for the transfer? In that case the acceptance tests needs to be adjusted. We could have a test with the wait logic to make sure transfer is completed and there aren't anything unexpected. Also maybe make sure the processor is idempotent when gitlabform is rerun and transfer is still in progress.

What do you think?

Not a big deal, but would be nice if the ci job token is in a separate PR.

rickbrouwer:
> Looks like the change makes gitlabform wait for the transfer to complete. Should it? What if maybe gitlabform follow gitlab and simply request for the transfer? In that case the acceptance tests needs to be adjusted. We could have a test with the wait logic to make sure transfer is completed and there aren't anything unexpected. Also maybe make sure the processor is idempotent when gitlabform is rerun and transfer is still in progress.
> 
> What do you think?

Without the wait, a config that combines `transfer_from` with any other section breaks on 19.4.
I can skip that, but then the remaining sections needs a re-run.

I’m personally not a fan of waits at all, so I understand your question.
The first thing I did was check whether this occurs elsewhere in our codebase, and I came across this:

https://github.com/gitlabform/gitlabform/blob/3095c63469150997a667183cae22388d98242368/gitlabform/gitlab/projects.py#L45-L57

That’s why I went ahead and did this.

> Not a big deal, but would be nice if the ci job token is in a separate PR.

Will do.

rickbrouwer:
#1433 needs to be merged first. Then this PR must be rebased.

amimas:
> Without the wait, a config that combines transfer_from with any other section breaks on 19.4.
I can skip that, but then the remaining sections needs a re-run.

Not sure what you mean. Could you share a config sample or explain again? I was going to try a simple test using gitlab API to see what happens to an attempt of project setting update after a transfer request is sent. Unfortunately having trouble getting gitlab running locally in container and ran out of time.

Don't want to rush this. Hoping we can avoid adding the wait logic. Although, I can see this not adding too performance impact overall for normal execution/workflow? Right?

Btw, that code snippet you referred to is no in use, I believe.

rickbrouwer:
Config sample that breaks without the wait:

  ```
  projects_and_groups:
    new_group/my_project:
      project:
        transfer_from: old_group/my_project
      project_settings:
        visibility: internal
        description: Some description
```

GitLabForm looks the project up by the key of the config entry, so the section after the transfer resolves `new_group/my_project`, which does not exist yet:

```
ERROR root:__init__.py:533 Error occurred while processing project new_group/my_project, exception:
404: 404 Project Not Found
gitlabform/__init__.py:535: SystemExit: 2
```

About potential performance: the wait is only for projects that have `transfer_from` in their config.

I don't know of a better solution for now. One option would be to handle the 400 error by indicating that the process is still underway and that it should succeed on the next run. But I don't consider that a good solution myself.

If you’d like to take a different approach or have a better solution, feel free to submit a code proposal or another PR so we can compare and test it.

amimas:
I agree with you. I wanted to avoid the waiting but I think what you have proposed works better.

amimas (approved):
Looks good.

